PR #83, opened 2021-11-10
Allow for 'main' as branch name value

guineveresaenger:
Part of https://github.com/pulumi/home/issues/933

This PR should allow us to generate a Workflow for building on a default branch named `main`.

I changed the class name as well to better illustrate the purpose of addressing a repo's default branch.

### DIFF
--- a/provider-ci/lib/workflows.js
+++ b/provider-ci/lib/workflows.js
@@ -33,7 +33,7 @@ export class DefaultBranchWorkflow extends g.GithubWorkflow {
     constructor(name, jobs) {
         super(name, jobs, {
             push: {
-                branches: ["master", "main"],
+                branches: [name],
                 'tags-ignore': ['v*', 'sdk/*', '**'],
                 'paths-ignore': [
                     "CHANGELOG.md"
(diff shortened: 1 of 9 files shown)


provider-ci/lib/workflows.js
```diff
@@ -29,11 +29,11 @@ const env = Object.assign({
     SLACK_WEBHOOK_URL: '${{ secrets.SLACK_WEBHOOK_URL }}',
     PULUMI_GO_DEP_ROOT: '${{ github.workspace }}/..',
 }, extraEnv);
-export class MasterWorkflow extends g.GithubWorkflow {
+export class DefaultBranchWorkflow extends g.GithubWorkflow {
```
stack72, line 32: Do we need to have 2 different workflow files? Or can we have a single workflow that has both values in the branch list? The bran list is only set to a single value using "name" below but that can easily be "master" and "main" right?

stack72 (commented):
Left a comment about design approach but not suggesting we need to change it

guineveresaenger:
Tried this out on Tailscale and it generates both a `main.yml` as well as a `master.yaml`. 

Instead of hard coding "master" I named the job by its inherited job name, since they were the same anyway.

Either way, the danger of doing this is that when migrating from `master` to `main`, folks may accidentally trigger both runs as they would exist in parallel. I would love to add actual logic so that we don't have equally valid runs on two "default" branches ever.

stack72:
Ok I guess this puts us in better shape going forward since we actually have a workflow file we can delete later :)

stack72 (approved):

provider-ci/lib/workflows.js
guineveresaenger: you caught me typing out some thoughts on the PR :)

I was going with two separate files so that we can pick and choose which to apply when copying over the generated workflows and so it's really obvious which is which

PR #92, opened 2021-11-23
Remove Microplane.

jkodroff:
We've replaced Microplane (mp) with a GitHub Action that deploys changes
to all providers.

Also corrects spacing in the README that broke numbering on ordered
lists.


stack72 (approved):
I love this PR so much!! thank you :)

PR #95, opened 2021-12-06
Go version + actions updates + gotestfmt addition

stack72:
- Upgrade the versions of the Actions used in ci-mgmt
- Upgrade to Go 1.17.x
- Adding Gotestfmt to CI scripts for ensuring simplicity of test runs
- Regenerating source based on updated workflows


jkodroff (approved):
Looks fine, but would be able to give a more in-depth review with the added generation/diff of a single provider's files.

stack72:
> Looks fine, but would be able to give a more in-depth review with the added generation/diff of a single provider's files.

I included them as you asked

PR #134, opened 2022-02-14
Allow for non-master default branch names

jkodroff:
Removed the unnecessary "base" option when creating PRs for updating the
bridge or the upstream provider. The create-pull-request action defaults
to the branch that is checked out in actions/checkout, which is the
default branch anyway.

Tested both the bridge and upstream provider workflows with pulumi-artifactory, which was has a default branch named "main".


jkodroff:
Duplicate of #129

PR #135, opened 2022-02-15
When an upstream provider upgrade fails, comment on the linked issue …

### LINKED ISSUE #120: Add notification if we fail to create a PR for a provider upgrade
A failed update-upstream-provider workflow attempt (e.g. because there are new resources) currently alerts to #builds in Slack. The failed attempt is not actually an alert-type occurrence because while it requires human intervention to add the mappings, it's not urgent.

The failed attempt should instead add a comment to the linked issue (if provided) indicating the automation attempt failed.

~In addition, will need to change the code [here](https://github.com/pulumi/home/blob/master/infrastructure/build-metrics/compute.ts#L158) to ignore the workflow in the #builds channel.~ <-- @stack72 took care of this

jkodroff:
…with a link to the failed run.

Fixes #120

### DIFF
--- a/provider-ci/lib/workflows.js
+++ b/provider-ci/lib/workflows.js
@@ -297,6 +297,11 @@ export class UpdateUpstreamProviderWorkflow extends g.GithubWorkflow {
                     nodeversion: [nodeVersion],
                 },
             })
+                .addStep({
+                id: "run-url",
+                name: "Create URL to the run output",
+                run: "echo ::set-output name=run-url::https://github.com/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID"
+            })
                 .addStep(new steps.CheckoutRepoStep())
                 .addStep(new steps.CheckoutTagsStep())
                 .addStep(new steps.InstallGo())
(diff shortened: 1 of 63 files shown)


provider-ci/lib/workflows.js
```diff
@@ -334,6 +339,17 @@ export class UpdateUpstreamProviderWorkflow extends g.GithubWorkflow {
                 if: "${{ github.event.inputs.linked_issue_number }}",
                 with: Object.assign(Object.assign({}, prStepOptions), { body: "Fixes #${{ github.event.inputs.linked_issue_number }}\n\nThis pull request was generated automatically by the update-upstream-provider workflow in this repository." })
             })
+                .addStep({
+                name: "Comment on failed attempt",
+                if: "${{ failure() && github.event.inputs.linked_issue_number }}",
```
stack72, line 344: Any reason you use this comment action rather than 1 we already use in this repo?

jkodroff: Yes - the action we use elsewhere appears to only work on PRs. This step has to comment on an issue.

PR #140, opened 2022-02-21
change auth0 provider from 'alexkappa' to 'auth0'

kevinschoonover:
Not sure if this is the correct place for this PR, but [it seems like auth0 is taking official ownership of their terraform plugin](https://github.com/auth0/terraform-provider-auth0/).

This should update the upstream provider to reflect that and pickup the latest release.

### DIFF
--- a/provider-ci/package-lock.json
+++ b/provider-ci/package-lock.json
@@ -1,8 +1,2387 @@
 {
   "name": "@pulumi/ci-mgmt",
   "version": "0.1.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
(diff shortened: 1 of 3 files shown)


jkodroff:
@kevinschoonover Thanks for the heads up and PR submission!

FYI, this is one of several places we'd have to make this change. Most notably, we'd also have to change https://github.com/pulumi/pulumi-auth0/blob/master/provider/go.mod first. This work can be tracked in https://github.com/pulumi/pulumi-auth0/issues/122. Once that's done, we can merge here.

kevinschoonover:
@jkodroff looks like PR was updated, do we need to do anything else here?

guineveresaenger:
pinging @josh to see if we can merge this now <3

PR #142, opened 2022-02-24
Ensure that ONLY release workflows tag the Go SDK

stack72:
This is not needed for anything other than the release workflow
and we should not use the ref name - we should calculate it like
we do during build time

Without this change ALL master builds would create a tag and push that tag to the repo - even more - master builds would be being tagged as sdk/master and pushing those tags

### DIFF
--- a/provider-ci/lib/steps.js
+++ b/provider-ci/lib/steps.js
@@ -483,9 +483,8 @@ export class TagSDKTag extends step.Step {
     constructor() {
         super();
         return {
-            if: 'success() && github.event_name == \'push\'',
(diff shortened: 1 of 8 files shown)


danielrbradley (approved):
So, adding a step to the `PublishSDKJob` was incorrect because that's also reused in other workflows - not only the release workflow.

Therefore, you're creating a new `TagSDKJob` which is only added to the release workflow.

Would be nice to have the slack alerting after the tagging in case the tag push fails, but perhaps we get this in as-is and iterate more after?

PR #150, opened 2022-03-07
Change how Gorelease strips internal changelog messages

stack72:


provider-ci/lib/goreleaser.js
```diff
@@ -87,9 +87,9 @@ export class PulumiGoreleaserConfig extends PulumiGoreleaserPreConfig {
                 exclude: [
                     "Merge branch",
                     "Merge pull request",
-                    "[internal]",
-                    "[ci]",
-                    "[chore]",
+                    "internal",
```
guineveresaenger, line 90: should we keep the square brackets as part of the pattern though? we don't want to accidentally exclude any commits that just have this word in them?

provider-ci/providers/aiven/repo/.goreleaser.yml
```diff
@@ -33,9 +33,9 @@ changelog:
     exclude:
     - Merge branch
     - Merge pull request
-    - '[internal]'
-    - '[ci]'
-    - '[chore]'
+    - internal
```
guineveresaenger, line 36: I think we should come up with a way to escape the square brackets so they're part of the regex.

guineveresaenger, line 36: I think we should come up with a way to generate this to say `[internal]`.

guineveresaenger (changes requested):
I think this is going to improve things but I have a concern that it's going to catch real commit messages accidentally.

provider-ci/providers/aiven/repo/.goreleaser.yml
```diff
@@ -33,9 +33,9 @@ changelog:
     exclude:
     - Merge branch
     - Merge pull request
-    - '[internal]'
-    - '[ci]'
-    - '[chore]'
+    - WinternalW
```
guineveresaenger, line 36: This doesn't look right either.. we want it to still say `\Wchore\W` no?

stack72: The \ isn’t being kept in the generate… that’s weird

jkodroff (approved):

guineveresaenger (approved):
LGTM! Thanks!

Issue #153, opened 2022-03-09
Use native automerge


jkisk:
Using the built in automerge functionality will reduce the number of GHA jobs we are running and reduce our likelihood of getting throttled.

- [ ] toggle ```allow automerge``` in all provider repos
- [ ] add a step to deploy-all-providers.yml to request automerge when prs are created
- [ ] add documentation to the tf-bridge-boilerplate repo to inform creators of new providers to toggle on automerge in the repo 
settings.

PR #154, opened 2022-03-09
Use built in GH automerge

### LINKED ISSUE #144: Automerge job times out before all checks are green
## Issue details

The automerge feature is timing out before the other checks are green.
Code is here: https://github.com/pulumi/ci-mgmt/blob/master/provider-ci/src/shared-workflows.ts#L27

example PR that should have automerged https://github.com/pulumi/pulumi-aiven/pull/183

### Steps to reproduce

This occurs most often on automated provider action updates, and the time it takes for the other checks to run and succeed is variable, depending on action runner availability.

Expected: Most smaller providers should automerge
Actual: The automerge action times out before the entire check suite is green

### LINKED ISSUE #155: add a step to deploy-all-providers.yml to request automerge when prs are created

jkisk:
Add a step to deploy-all-providers.yml that will enable automerge via squash on all created prs.  Automerge will only happen if all required checks pass, but it will not time out and will remove our need to run our own automerge job.

related #153 
fixes #144
resolves #155

### DIFF
--- a/.github/workflows/deploy-all-providers.yml
+++ b/.github/workflows/deploy-all-providers.yml
@@ -55,9 +55,6 @@ jobs:
           title: "Update GitHub Actions workflows."
           path: pulumi-${{ matrix.provider }}
           token: ${{ secrets.PULUMI_BOT_TOKEN }}
-      # See: https://docs.github.com/en/rest/guides/best-practices-for-integrators#dealing-with-secondary-rate-limits
-      - name: Sleep to prevent hitting secondary rate limits
-        run: sleep 1
       - name: Set PR to auto-merge
         if: steps.create-pr.outputs.pull-request-operation == 'created'
         uses: peter-evans/enable-pull-request-automerge@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-all-providers.yml
```diff
@@ -43,17 +43,27 @@ jobs:
...
       - name: Sleep to prevent hitting secondary rate limits
         run: sleep 1
+      - name: Set PR to auto-merge
+        if: steps.create-pr.outputs.pull-request-operation == 'created'
+        uses: peter-evans/enable-pull-request-automerge@v1
+        with:
+          token: ${{ secrets.PULUMI_BOT_TOKEN }}
+          pull-request-number: ${{ steps.create-pr.outputs.pull-request-number }}
+          merge-method: squash
+      - name: Sleep again to prevent hitting secondary rate limits
```
jkodroff, line 68: Nit: I think we just need to sleep once to ensure that the entire workflow takes at least 1 second.

jkodroff (approved):

PR #164, opened 2022-03-18
Try updating bridge with single provider

jkisk:
First commit for the purpose of testing flow in single provider before updating more widely.


stack72 (changes requested):
You have some yaml formatting errors here with the if layer

jkodroff (approved):

PR #166, opened 2022-03-22
Add update-bridge-single-provider.yml

jkisk:
Adding this because I wanted to use it for [#1896](https://github.com/pulumi/home/issues/1896) and thought it might be useful from time to time.

### DIFF
--- a/.github/workflows/update-bridge-single-provider.yml
+++ b/.github/workflows/update-bridge-single-provider.yml
@@ -1,0 +1,31 @@
+# Performs a workflow_dispatch to run the update-bridge workflow in a specified
+# provider.
+name: Update bridge, single provider
+on:
+  workflow_dispatch:
+    inputs:
+      provider_name:
+        description: The name of the provider to update - do not include the pulumi prefix in the name.
+        required: true
+        type: string
+      bridge_version:
+        description:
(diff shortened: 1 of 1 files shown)


.github/workflows/update-bridge-single-provider.yml
```diff
@@ -0,0 +1,31 @@
...
+        type: string
+      bridge_version:
+        description:
+          The version of pulumi/pulumi-terraform-bridge to update to. Do
+          not include the 'v' prefix. Must be major version 3.
+        required: true
+        type: string
+      sdk_version:
+        description:
+          The version of pulumi/pulumi/sdk to update to. Do not include
```
jkodroff, line 19: Nit: Should be `pulumi/sdk`, I think?

jkodroff (approved):
Approved with an FYI and a nit.

.github/workflows/update-bridge-single-provider.yml
jkisk: I think either could be considered correct, and as a Pulumian this helps orient me a bit, but I can see why not as useful for community. This is consistent with the other file, so if we change we should change both, I'll leave as is for now.

PR #168, opened 2022-03-24
Replace automerge action with native automerge

jkisk:
In this PR I remove the older automerge job, but the existing pr-automation.yml files will need to be removed from the provider repos, perhaps with a one off run of the deploy all providers action?  Added is a step in the update terraform bridge workflow that will set the created PR to automerge, via the squash method.

Resolves [#1901](https://github.com/pulumi/home/issues/1901)


provider-ci/src/workflows.ts
```diff
@@ -268,6 +268,7 @@ export class UpdatePulumiTerraformBridgeWorkflow extends g.GithubWorkflow {
                 .addStep(new steps.RunCommand('make build_sdks'))
                 .addStep({
                     name: "Create PR",
+                    id: "create-pr",
```
jkodroff, line 271: I do not typically set `id` for GHA steps. Are there benefits to doing so (aside from where you need a step's output, which requires the ID)? (Just asking out of curiosity - it has no effect on my review.)

jkodroff (commented):
Need some additional info:

1. If we haven't already done so, could we generate files for a single provider, deploy, and verify that it works?
2. Can we ensure the commit message details _why_ we're doing this (to save GHA runners)?

provider-ci/src/workflows.ts
jkisk: I'm not aware of any other benefit unless you need it, and you can avoid needing it by not using spaces in the name and using that instead to access inputs and outputs.

jkisk:
> Need some additional info:
> 
> 1. If we haven't already done so, could we generate files for a single provider, deploy, and verify that it works?
> 2. Can we ensure the commit message (or the linked GH issue) details _why_ we're doing this (to save GHA runners)?

1. Yes, I verified using the pulumi-artifactory provider, you can see update-bridge.yml there currently and it worked as expected.

2. I can update the linked GitHub issue, but aside from saving GHA runners, the native auto merge is preferable for several reasons, including never timing out, and working instantly when a green state is achieved	.

jkodroff (approved):

PR #170, opened 2022-03-25
Hack better types

danielrbradley:
- Remove use of `jk` tool - use the `yaml` npm package instead
- Make github workflow schema types reproducable
- Use types better to verify correctness
- Use vanilla Typescript approach - generate all providers in single program run
- Use zod for `config.yaml` parsing rather than `jkcfg`
- Fix ESLint & add auto-formatting
- Avoid having to generate lib JS files - just run via ts-node


jkodroff (approved):
Went over on Zoom. All looks good.

stack72:
Let's talk more about this before merging so that we are all on the same page

danielrbradley:
> Let's talk more about this before merging so that we are all on the same page

Maybe Monday would be a good time to quickly run through this. Also, I'd like to wait until Guin's `moderator` branch is merged so there's no outstanding work that will need re-doing.

stack72 (approved):

PR #171, opened 2022-03-25
Remove old automerge action

jkisk:
This change will add removing all existing workflows as a step before we copy in new ones, to ensure cruft is not left behind as we go.

### DIFF
--- a/.github/workflows/update-workflows-all-providers.yml
+++ b/.github/workflows/update-workflows-all-providers.yml
@@ -39,6 +39,8 @@ jobs:
         with:
           repository: pulumi/pulumi-${{ matrix.provider }}
           path: pulumi-${{ matrix.provider }}
+      - name: Delete old automerge
+        run: rm pulumi-${{ matrix.provider }}/.github/workflows/pr-automation.yml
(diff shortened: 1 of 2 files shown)


jkisk:
validated with single provider here: https://github.com/pulumi/pulumi-slack/pull/10/files

stack72 (changes requested):
I think we should delete all of the yml files in the `pulumi-${{ matrix.provider }}/.github/workflows` folder - that way we can always be sure we are deleting old files

.github/workflows/update-workflows-all-providers.yml
```diff
@@ -39,6 +39,8 @@ jobs:
         with:
           repository: pulumi/pulumi-${{ matrix.provider }}
           path: pulumi-${{ matrix.provider }}
+      - name: Delete old automerge
+        run: rm pulumi-${{ matrix.provider }}/.github/workflows/pr-automation.yml
```
stack72, line 43:
```
run: rm pulumi-${{ matrix.provider }}/.github/workflows/*.yml
```

stack72 (approved):

PR #192, opened 2022-05-24
Add Central Management for Native Providers

jkisk:
Resolves #https://github.com/pulumi/home/issues/2066


stack72 (approved):
@jkisk this is awesome! One thing we will want to ensure is set in the individual config files for the providers is the major-version. It defaults to 2 and none of the native providers are v2 :D But that's not about this code, that's just an optimization

jkisk:
> @jkisk this is awesome! One thing we will want to ensure is set in the individual config files for the providers is the major-version. It defaults to 2 and none of the native providers are v2 :D But that's not about this code, that's just an optimization

Will do!  Thanks for drawing that to my attention :)

PR #195, opened 2022-05-30
Generate makefiles

danielrbradley:
- Comments are removed, some line breaks and spacing changed, otherwise identical.
- Make generation opt-in via setting `makeTemplate: bridged` in the `config.yaml`.
- Adapt generation to handle non-yaml output.

### DIFF
--- a/provider-ci/src/make-bridged-provider.ts
+++ b/provider-ci/src/make-bridged-provider.ts
@@ -1,5 +1,4 @@
 import { Makefile, Target } from "./make";
-import { goBuild, genSchema, yarnInstall, cwd } from "./make-targets";
 
 type BridgedProviderConfig = {
   provider: string;
@@ -15,7 +14,7 @@ export function bridgedProvider(config: BridgedProviderConfig): Makefile {
   const PROVIDER_PATH = `provider/v5`;
   const SDK_PATH = `sdk/v5`;
   const VERSION_PATH = `${PROVIDER_PATH}/pkg/version.Version`;
-  const TF_GEN = `pulumi-tfgen-${PACK}`;
+  const TFGEN = `pulumi-tfgen-${PACK}`;
   const PROVIDER = `pulumi-resource-${PACK}`;
   const VERSION = "$(shell pulumictl get version)";
   const TESTPARALLELISM = "10";
@@ -30,7 +29,7 @@ export function bridgedProvider(config: BridgedProviderConfig): Makefile {
     PROVIDER_PATH,
     SDK_PATH,
     VERSION_PATH,
-    TF_GEN,
+    TFGEN,
     PROVIDER,
     VERSION,
     TESTPARALLELISM,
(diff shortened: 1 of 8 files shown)


provider-ci/src/make-bridged-provider.ts
```diff
@@ -0,0 +1,228 @@
...
+    VERSION,
+    TESTPARALLELISM,
+    WORKING_DIR,
+  } as const;
+
+  const install_plugins: Target = {
+    name: "install_plugins",
+    phony: true,
+    commands: [
+      "[ -x $(shell which pulumi) ] || curl -fsSL https://get.pulumi.com | sh",
```
stack72, line 43: We may be able to do this in a loop right? Make it less verbose - it would iterate a map of provider / version

provider-ci/src/makefiles.ts
```diff
@@ -0,0 +1,14 @@
+import { render } from "./make";
+import { bridgedProvider } from "./make-bridged-provider";
+
+export function buildMakefile(config: {
+  provider: string;
+  makeTemplate: string;
+}): string {
+  switch (config.makeTemplate) {
+    case "bridged":
```
stack72, line 9: I like this extensibility - gives us the ability to add the native providers :)

provider-ci/providers/aiven/repo/Makefile
```diff
@@ -0,0 +1,98 @@
...
+		dotnet build /p:Version=$(DOTNET_VERSION)
+
+lint_provider:: provider
+	cd provider && golangci-lint run -c ../.golangci.yml
+
+cleanup:: 
+	rm -r $(WORKING_DIR)/bin
+	rm -f provider/cmd/$(PROVIDER)/schema.go
+
+help::
```
jkodroff, line 71: I'd like it if we did have the help text for the various targets. Right now `make help` does nothing.

provider-ci/providers/aiven/repo/Makefile
```diff
@@ -0,0 +1,98 @@
...
+TFGEN := pulumi-tfgen-$(PACK)
+PROVIDER := pulumi-resource-$(PACK)
+VERSION := $(shell pulumictl get version)
+TESTPARALLELISM := 10
+WORKING_DIR := $(shell pwd)
+
+development:: install_plugins provider build_sdks install_sdks
+
+build:: install_plugins provider build_sdks install_sdks
+
```
jkodroff, line 17: Would like to see targets for `provider/go.sum`, etc. This would also probably necessitate taking the option into account whether the provider is shimmed.

provider-ci/providers/aiven/repo/Makefile
stack72: Let's not worry about that yet - it doesn't work already so won't work now either :)

provider-ci/providers/aiven/repo/Makefile
stack72: We are generating like for like right now - let's not couple enhancements into the same PR

stack72 (approved):
This is one of the biggest drives for engineering systems we have made this year :D Thank you!!!

Issue #205, opened 2022-06-03
GitHub Bot should not reply to pulumi-bot RE community contributions


jkodroff:
In <https://github.com/pulumi/pulumi-azure/pull/1093>, the bot is replying to pulumi-bot, which is silly. We should consider pulumi-bot a member of the pulumi org.

PR #207, opened 2022-06-06
Only add codegen warning if the user is not a pulumi member

stack72:
Fixes: #205

### DIFF
--- a/provider-ci/src/workflows.ts
+++ b/provider-ci/src/workflows.ts
@@ -974,7 +974,7 @@ export function ModerationWorkflow(
     },
 
     jobs: {
-      warn_codegen: new WarnCodegenJob("warn_codegen"),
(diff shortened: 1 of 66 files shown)


provider-ci/src/workflows.ts
```diff
@@ -974,7 +974,7 @@ export function ModerationWorkflow(
     },
 
     jobs: {
-      warn_codegen: new WarnCodegenJob("warn_codegen"),
+      warn_codegen: new WarnCodegenJob("warn_codegen")
```
guineveresaenger, line 977: I'm pretty sure we don't strictly need a comma here because TS magic, but everywhere else has it

guineveresaenger (approved):
I would love to change the PR description to

"if the PR originated from a fork" because that is more precise.

I love this solution though. Thank you.

PR #208, opened 2022-06-08
Dev experience improvements

danielrbradley:
Major changes:
- Use yarn instead of npm (as we do in most other repos)
- Make a smarter makefile to allow up-to-date targets (speed up dev loop)
- Unify config between examples and bridged repos
- Generate a schema for the config yaml for autocomplete

More ambitious:
- Create single generate entry point for all repos
- Start planning how to migrate config files and output directories for multiple types of repos

Next up:
- [ ] Generate our own ci-mgmt actions
- [ ] Flatten whole directory structure down a bit


jkisk (approved):
This all works for me, thanks Daniel!

PR #210, opened 2022-06-22
Weekly Pulumi update for native providers

jkisk:
resolves https://github.com/pulumi/home/issues/2084


native-provider-ci/providers/aws-native/repo/.github/workflows/weekly-pulumi-update.yml
```diff
@@ -0,0 +1,141 @@
...
+
+        if ! git diff-files --quiet; then 
+        	echo ::set-output name=changes::1 
+        fi
+    - name: Provider with Pulumi Upgrade
+      if: steps.gomod.outputs.changes != 0
+      run: >-
+        make codegen && make local_generate
+
+        git add sdk/nodejs
```
jkodroff, line 93: It's not clear to me why we're doing multiple commits. Could we either comment why we're doing this or use a single commit?

native-provider-ci/providers/aws-native/repo/.github/workflows/weekly-pulumi-update.yml
```diff
@@ -0,0 +1,141 @@
+name: weekly-pulumi-update
+on:
+  schedule:
+  - cron: 35 12 * * 4
```
jkodroff, line 4: Suggest adding a `workflow_dispatch` and maybe consider adding the ability to supply a ref or version number. If we only have `schedule`, we won't catch bug fix releases which occur async to the schedule and we'll have no way to trigger this workflow (because there's no `workflow_dispatch`).

We might also want to consider changing pulumi/pulumi to issue a `workflow_dispatch` when a new version of p/p is released to ci-mgmt which would in turn trigger update all the native providers, thereby making the entire process event-driven.

jkodroff (changes requested):
Thanks for doing this - it's clearly going to improve native provider maintenance!

My general thoughts are that we should make this process more in line with [the update-bridge workflow](https://github.com/pulumi/ci-mgmt/blob/master/provider-ci/providers/aiven/repo/.github/workflows/update-bridge.yml) or document why different steps are necessary.

native-provider-ci/providers/aws-native/repo/.github/workflows/weekly-pulumi-update.yml
jkisk: There is a workflow_dispatch on line 5

native-provider-ci/providers/aws-native/repo/.github/workflows/weekly-pulumi-update.yml
jkisk: I had fairly strict instructions to recreate the workflows that were already in place and not change them up a lot at this stage.

native-provider-ci/providers/aws-native/repo/.github/workflows/weekly-pulumi-update.yml
jkisk: I'm open to doing event driven p/p updates, but I think that is a decision that should be taken separately and for bridged providers as well.  The tradeoff is that immediate upgrades don't give a lot of time for issues to be discovered before being disseminated through provider ecosystem.

jkodroff (approved):

PR #217, opened 2022-07-13
Ensure we don't set v on the java version tag

stack72:


t0yv0 (approved):
Thank you!

provider-ci/providers/aiven/repo/.github/workflows/main.yml
```diff
@@ -79,7 +79,7 @@ jobs:
     - name: Update path
       run: echo "${{ github.workspace }}/bin" >> $GITHUB_PATH
     - name: Set PACKAGE_VERSION to Env
-      run: echo "PACKAGE_VERSION=v$(pulumictl get version --language generic)" >>
+      run: echo "PACKAGE_VERSION=$(pulumictl get version --language generic)" >>
```
danielrbradley, line 82: Why do we set the variable outside the makefile? Does this variable override the one in the makefile?

stack72: The makefile is used for the build of the SDK - we need to ensure the same version is set during the publish - that doesn't happen in the makefile

danielrbradley (approved):

provider-ci/providers/aiven/repo/.github/workflows/main.yml
t0yv0: Yes precisely! We thought through this and decided it would be better for code generation to emit version-agnostic code and then the build files decide on the version at build/publish time. Though still buggy (https://github.com/pulumi/pulumi-java/issues/735) I still believe that's the better approach.

PR #227, opened 2022-08-09
Jkisk/unique native workflows

jkisk:
Resolves #https://github.com/pulumi/home/issues/2075

I will open PRs in azure-native and aws-native that will be easier to review alongside current actions.


jkisk:
merging this as both resulting PRs were approved in azure-native and aws-native

PR #234, opened 2022-08-17
adding java to native providers

stack72:
- Adding Java to Native Provider builds
- Adding the sentinel workflow to the native providers


jkisk (approved):
Looks good, glad you are adding the sentinel step so we can simplify the branch protection changes!

PR #241, opened 2022-08-29
Install gke-cloud-auth-plugin in gnative CI env

viveklak:
Copied from https://github.com/pulumi/pulumi-google-native/pull/655


jkisk:
We will need to update the Google Cloud login step here: https://github.com/pulumi/ci-mgmt/blob/18340ac08d5d70fc26b6924c825a49aebf9eda4e/native-provider-ci/src/steps.ts#L155 and then run `make all` from the native-providers directory to regenerate the actions.

jkisk:
@viveklak does the commit I added look all right to you? If so :shipit:

jkisk (approved):

lblackstone (approved):

PR #254, opened 2022-09-16
Update examples repo with GCP ServiceAccount changes

lblackstone:


lblackstone:
Note: several of the examples tests are failing, but I'm not sure if that's related to mismatches in the GHA that were active and the updates here or not. All of the GCP authentication appears to be working as expected.

stack72 (approved):
I'm pretty sure the examples work here is very out of date so nice to see a specific Pr updated as well :)

PR #257, opened 2022-09-20
Makefile improvements

### LINKED ISSUE #202: [Epic] Makefile generation improvements
- [ ] Use file targets (see [awsx Makefile](https://github.com/pulumi/pulumi-awsx/blob/master/Makefile) for reference). This should allow local development loops to be faster.
- [ ] Generalise the makefile options to not assume providers are owned by Pulumi.
- [ ] Add provider shim mods target (i.e. run `go mod tidy` in a target)
- [ ] Make `provider-ci` consumable externally – something like `npx pulumi-gen-build`

danielrbradley:
Fixes #202 

Create a new version of makefile template: "bridged-v2" which must be opted into. We can enable this on a per-repo basis as we gain confidence in testing.

This includes a few general makefile formatting improvements:
- Explicitly set the "default" target so it's always first.
- Sort targets by name, grouped by type (alias/phony/file).

New makefile changes:

- Use file targets so we can have up-to-date builds.
- Model dependencies so dependencies automatically get built first.
- Use local pulumictl version.
- Use sentinel files to atomically execute targets.

When being rolled out - this will require some minor changes to the build process.


thomas11 (approved):
Tricky to review but looks good at first glance. Thank you, this will be a nice improvement!

provider-ci/src/make.ts
```diff
@@ -96,17 +97,57 @@ function phonyTarget(targets: Target[]): Target | undefined {
...
+}
+
+function descendentTargets(target: Target): Target[] {
+  if (target.dependencies == undefined) {
+    return [target];
+  }
+  return deduplicateTargets([
+    target,
+    ...target.dependencies.flatMap((t) =>
+      typeof t != "string" ? descendentTargets(t) : []
```
lblackstone: Should we use `!==` here instead? The `typeof` check on 119 uses `===`.

provider-ci/src/make.ts
```diff
@@ -96,17 +97,57 @@ function phonyTarget(targets: Target[]): Target | undefined {
...
+    ),
+  ]);
+}
+
+function sortTargets(targets: Target[], defaultTarget?: Target | string) {
+  const defaultName =
+    typeof defaultTarget === "string" ? defaultTarget : defaultTarget?.name;
+  const sorted = [...targets].sort((a, b) => {
```
lblackstone: Why is the `[...targets]` needed? Isn't `targets` already an array?

provider-ci/src/make.ts
```diff
@@ -96,17 +97,57 @@ function phonyTarget(targets: Target[]): Target | undefined {
...
+    if (a.name == defaultName) {
```
lblackstone: Should we use `===` here?

lblackstone (approved):

provider-ci/src/make.ts
danielrbradley: Calling `targets.sort` would modify the array passed into the function. I always avoid mutating input arguments unless it's performance critical, so I create a copy of the array then, sort the copy and return the copy.

ringods:
@danielrbradley can we also test drive this at Pulumiverse side for one of the bridged providers?

danielrbradley:
@ringods maybe - we've not had any time to do the work to make this easy to consume outside our core provider base, but feel free to fork for now or find another way to utilize it.

Issue #256, opened 2022-09-19
Do not block GitHub SDK tag on Java SDK publishing


guineveresaenger:
1. Java SDK isn't GA yet
2. We're dealing with [intermittent publish failures for Java SDK](https://github.com/pulumi/pulumi-java/issues/830)
3. This results in [failure to publish the SDK tag in the release job](https://github.com/pulumi/pulumi-mongodbatlas/actions/runs/3085502151/jobs/4989190896#step:23:402), which seems overly strict, since we only need these for Go, and failure to publish the Java SDK does not mean the GO SDK is broken
4. We should change the conditions in the [release workflow(example)](https://github.com/pulumi/pulumi-aws/blob/master/.github/workflows/release.yml#L240), either by:
  - change [this step](https://github.com/pulumi/pulumi-aws/blob/master/.github/workflows/release.yml#L372) to rely on `needs: publish`
  - separate out the Java SDK into its own step for the time being

PR #258, opened 2022-09-20
Splitting the publishing of the Java SDK

stack72:
Fixes: #256

While the Java project is in preview, we should not fail
the release jobs if the java sdk is not released. By failing the job,
we may have already released the node|dotnet|python packages as
well as the binaries. Failing the job means we don't cut a
Go SDK or create a docs release so the release job isn't correct

We are going to silently continue with the java sdk job. When the
Java SDK is in GA, we can combine these jobs again


viveklak (approved):
Thanks for getting this in @stack72!

PR #262, opened 2022-09-22
Fix weekly pulumi update for k8s provider

lblackstone:


jkisk (approved):
Thank you! @danielrbradley so you can see updated use of makefile

native-provider-ci/providers/kubernetes/repo/.github/workflows/weekly-pulumi-update.yml
```diff
@@ -83,25 +83,7 @@ jobs:
 
         git checkout -b update-pulumi/${{ github.run_id }}-${{ github.run_number }}
 
-        cd provider
-
-        go get github.com/pulumi/pulumi/pkg/v3
-
-        go get github.com/pulumi/pulumi/sdk/v3
```
lblackstone, line 90: Whoops -- it would help if I actually updated the modules...

PR #268, opened 2022-09-28
Automate the creation of the gitignore workflows

stack72:


native-provider-ci/providers/kubernetes/repo/.gitignore.yml
```diff
@@ -0,0 +1,40 @@
+# WARNING: This file is autogenerated - changes will be overwritten if not made via https://github.com/pulumi/ci-mgmt
+
+|
```
danielrbradley, line 3: Where's this `|` coming from?

native-provider-ci/src/gitignore.ts
```diff
@@ -0,0 +1,38 @@
...
+.version
+
+sdk/java/build
+sdk/java/src
+sdk/java/.gradle
+sdk/java/gradle
+sdk/java/gradlew
+sdk/java/gradlew.bat
+pulumi-azure-native-sdk
+.DS_Store
```
danielrbradley, line 36: In addition to our provider-specific ignores, we could base the more generic ignores from this list: https://www.toptal.com/developers/gitignore/api/node,java,dotnetcore,macos,windows,linux,python,go

stack72:
@danielrbradley I found a bug in the implementation of the native providers generation that always treated as yaml therefore the `|`. That's now fixed :) I will let you fine peeps discuss what needs to go into the gitignore file now :)

PR #269, opened 2022-09-28
Switch polarity for failing on mapping errors

viveklak:


provider-ci/src/workflows.ts
```diff
@@ -264,8 +264,8 @@ export function UpdatePulumiTerraformBridgeWorkflow(
       // If there are missing or extra mappings, they can not have been
       // introduced by updating the bridge, so for this workflow we'll
       // ignore mapping errors.
-      PULUMI_EXTRA_MAPPING_ERROR: false,
-      PULUMI_MISSING_MAPPING_ERROR: false,
+      PULUMI_SKIP_EXTRA_MAPPING_ERROR: true,
```
viveklak, line 267: Not convinced this is what we want to do here but sticking with the existing semantics for now in the case of bridge upgrades.

jkodroff:
I believe AliCloud will need an exception to missing mappings.

viveklak:
> I believe AliCloud will need an exception to missing mappings.

That is done here: https://github.com/pulumi/ci-mgmt/pull/269/files#diff-ebf701c14f061a2f57de60aa77b9b6e45cc4c1cbc3f22dacd756f58d728b9ad1R15

lblackstone (approved):

viveklak:
I won't merge this till we fix the builds for TF provider.

PR #274, opened 2022-10-27
Add option to switch schema tool

viveklak:


lblackstone (approved):

provider-ci/src/steps.ts
```diff
@@ -509,13 +509,18 @@ export function PullRequest(
   };
 }
 
-export function CheckSchemaChanges(): Step {
+export function CheckSchemaChanges(preferredSchemaChecker: string): Step {
+  const schemaCheckerCommand =
+    preferredSchemaChecker == "mikhailshilkov/schema-tools"
+      ? "schema-tools compare ${{ env.PROVIDER }} master --local-path=provider/cmd/pulumi-resource-${{ env.PROVIDER }}/schema.json >> $GITHUB_ENV\n"
```
guineveresaenger, line 515: For pulumi/schema-tools, we need to be [explicit with the `-o` flag here](https://github.com/pulumi/schema-tools/blob/master/cmd/compare.go#L29) - I notice pulumi/schema-tools defaults to `master` and that is not the default branch for every provider (pulumi-artifactory being an example but there are others).
Something like `-o github.event.repository.default_branch` should work for all cases. 

Additionally, in Mikhail's schema tools, there is no default nor an `-o` flag, but we still should not hard code master here.

provider-ci/src/config.ts
```diff
@@ -13,6 +13,7 @@ const BridgedConfig = z
   .object({
     template: z.literal("bridged").default("bridged"),
     provider: z.string(),
+    preferredSchemaChecker: z.string().default("mikhailshilkov/schema-tools"),
```
guineveresaenger, line 16: I'm curious why we're making this configurable.
Also wondering why we're defaulting to Mikhail's tool.

guineveresaenger (changes requested):
Unfortunately, it appears there's differences in incantation between the two schema tools.

I want to say that adding an option to choose isn't strictly necessary here either - I think a plain cutover is reasonably safe - but either way we should discuss more.

provider-ci/src/config.ts
viveklak: Because this is the current default. My hope is we roll this out in a few providers as we upgrade them and validate things work as expected. Once the tier 1 providers are updated we can switch the default.

provider-ci/providers/auth0/repo/.github/workflows/main.yml
```diff
@@ -157,7 +157,7 @@ jobs:
       name: Install Schema Tools
       uses: jaxxstorm/action-install-gh-release@v1.5.0
       with:
-        repo: mikhailshilkov/schema-tools
+        repo: pulumi/schema-tools
```
thomas11, line 160: You made the schema-tools version configurable, but there are also several places such as this one where the switch is hard-coded. What's the reasoning behind that?

viveklak (commented):
As discussed in-person, the plan here is to instead add `pulumi/schema-tools` comment in addition to existing `schema-tools` for non-tier1 providers. Once we feel confident the coverage is accurate we can cut over to `pulumi/schema-tools`

provider-ci/providers/auth0/repo/.github/workflows/main.yml
viveklak: This (and all files in the auth0/repo directory) are generated by the tool - in this case the pulumi one was explicitly opted-in here: https://github.com/pulumi/ci-mgmt/pull/274/files#diff-a4be1b53045f8817dbfd12f8cd71c655048b06d0cc2ef084bf4717b677e05a77R2.

viveklak: As mentioned in https://github.com/pulumi/ci-mgmt/pull/274#pullrequestreview-1160555851 I plan on instead adding a parallel evaluation for pulumi/schema-tools on some non-tier1 providers. Once we have a little bit of confidence, we will switch entirely to the pulumi version and make that the preferred tool.

provider-ci/src/config.ts
danielrbradley, line 16: I'd suggest using an explicit union to restrict the valid input types. I.e. something like:

```typescript
preferredSchemaChecker: z.enum(["mikhailshilkov/schema-tools", "pulumi/schema-tools"]).default("mikhailshilkov/schema-tools")
```

provider-ci/src/steps.ts
```diff
@@ -197,13 +197,13 @@ export function InstallPulumiCtl(): Step {
   };
 }
 
-export function InstallSchemaChecker(): Step {
+export function InstallSchemaChecker(preferredSchemaChecker: string): Step {
```
danielrbradley, line 200: It might be nice to make the valid set of inputs explicit here:

```suggestion
type PreferredSchemaChecker = "mikhailshilkov/schema-tools" | "pulumi/schema-tools"
export function InstallSchemaChecker(preferredSchemaChecker: PreferredSchemaChecker): Step {
```

This will mean that the compiler will tell you if we update the allowed config values but forget to handle the addtional cases here.

danielrbradley (approved):
I think the approach of not yet changing the default is good - do a gradual rollout and check there's not unintended impact. Only changes I'd suggest is avoiding string types - be more specific on allowed values, though given this might only be short-lived it might not matter much.

iwahbe:
Is this PR still worth carrying forward? It looks like we are still relying on @mikhailshilkov's repo:

https://github.com/pulumi/ci-mgmt/blob/bb2bd5cb4dbf7ce3a8fbc2285ab3695280e7e61b/provider-ci/providers/auth0/repo/.github/workflows/main.yml#L154-L158

CC @thomas11 @guineveresaenger @lblackstone @danielrbradley

danielrbradley:
@iwahbe we should definitely get off Mikhail's repo and onto our own. We've already switched to this version for azure-native and it's fine so maybe could just roll this out in one go now. Looks like a bunch of conflicts too so might be worth just opening a new PR.

iwahbe:
I opened https://github.com/pulumi/ci-mgmt/issues/404 to track.

iwahbe:
Closing in favor of https://github.com/pulumi/ci-mgmt/pull/405.

PR #276, opened 2022-11-01
add env var to append user agent

jaxxstorm:
If a provider uses the plugin SDK mechanism for setting the user agent,
we can append pulumi to it automatically this way


guineveresaenger:
this looks good to me - can you [generate the code](https://github.com/pulumi/ci-mgmt#updating-all-providers) for all TF providers?

t0yv0:
Generated code for providers @guineveresaenger - should we merge?

guineveresaenger (approved):

PR #278, opened 2022-11-08
Add support for building Go SDK

### LINKED ISSUE #275: All providers should build the go sdk as part of publishing as min-bar
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details
We discovered that we missed a regression in go sdk for gcp - see https://github.com/pulumi/pulumi/issues/11207#issuecomment-1297642228. In most of our native providers we actually build our go sdk, e.g.: https://github.com/pulumi/pulumi-google-native/blob/master/Makefile#L98 (makefiles not managed in ci-mgmt as yet). We need to extend the same to bridged providers as well so we can catch build issues quickly.

viveklak:
Fixes #275 

Change tested in https://github.com/pulumi/pulumi-gcp/pull/943

### DIFF
--- a/provider-ci/src/make-bridged-provider.ts
+++ b/provider-ci/src/make-bridged-provider.ts
@@ -109,6 +109,7 @@ export function bridgedProvider(config: BridgedConfig): Makefile {
     phony: true,
     commands: [
       "$(WORKING_DIR)/bin/$(TFGEN) go --overlays provider/overlays/go --out sdk/go/",
+      `cd sdk && go list \`grep -e "^module" go.mod | cut -d ' ' -f 2\`/go/... | xargs go build`,
     ],
   };
   const build_dotnet: Target = {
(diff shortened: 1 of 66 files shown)


provider-ci/src/make-bridged-provider.ts
```diff
@@ -109,6 +109,7 @@ export function bridgedProvider(config: BridgedConfig): Makefile {
     phony: true,
     commands: [
       "$(WORKING_DIR)/bin/$(TFGEN) go --overlays provider/overlays/go --out sdk/go/",
+      `cd sdk && go list \`grep -e "^module" go.mod | cut -d ' ' -f 2\`/go/... | xargs go build`,
```
danielrbradley, line 112: Would be good to have a comment in the source ... or in the generated code explaining this ... I'm unfamiliar with the `grep` + `cut` +`xargs` magic or what we're trying to achive 🙈

viveklak: Just updated with a comment explaining what this is doing :).

PR #281, opened 2022-11-14
Support for kicking off a resync build workflow across providers

viveklak:
This change primarily adds a general-purpose `resync-build` workflow. This will eventually become a more pluggable workflow for operational or house-keeping changes. The current incarnation is specifically focussed on fixing up broken java SDK roll out.

This also switches to using the `gh` to trigger auto merge when safe instead of using a bespoke github action. Moreover, automerge is still turned off by default unless the dispatch of the relevant workflows explicitly set the workflow parameter to trigger automerge.

I intend to roll it out gradually starting with providers with broken builds.


provider-ci/src/steps.ts
```diff
@@ -1,9 +1,29 @@
...
+    }
+    if (args.path) {
+      checkOutWith["path"] = args.path;
+    }
+    return {
+      name: "Checkout repo",
+      uses: action.checkout,
+      with: checkOutWith,
+    };
+  }
```
guineveresaenger, line 26: nice!

provider-ci/src/workflows.ts
```diff
@@ -336,6 +346,108 @@ export function UpdatePulumiTerraformBridgeWorkflow(
   };
```
guineveresaenger, line 346: do we want to remove this?

provider-ci/src/workflows.ts
```diff
@@ -319,6 +325,10 @@ export function UpdatePulumiTerraformBridgeWorkflow(
             "team-reviewers": "platform-integrations",
             token: "${{ secrets.PULUMI_BOT_TOKEN }}",
           },
+        })
+        .addStep({
+          if: "steps.create-pr.outputs.pull-request-operation == 'created' && github.event.inputs.automerge == 'true'",
+          run: "gh pr merge --auto --squash ${{ steps.create-pr.outputs.pull-request-number }}",
```
guineveresaenger, line 331: Do we know whether this is going to _not_ get prematurely triggered by race conditions  between CI jobs?

guineveresaenger (approved):
Left a nit and a question but LGTM!

provider-ci/src/workflows.ts
viveklak: Not entirely sure. However, I am definitely being defensive here: `&& github.event.inputs.automerge == 'true'`

provider-ci/src/workflows.ts
viveklak: Yes. Just did.

PR #283, opened 2022-11-15
Use pulumi package sdk-gen to generate SDKs

Frassle:
Instead of using the built tfgen, eventually tfgen will not have code generation capabilities included in it.

Some providers might need schemas updating to specifiy `--build gradle-nexus` for Java, as `gen-sdk` doesn't pass options down to SDK generation yet. See
https://github.com/pulumi/pulumi-command/pull/158/files.

### DIFF
--- a/provider-ci/src/make-bridged-provider-2.ts
+++ b/provider-ci/src/make-bridged-provider-2.ts
@@ -163,8 +163,8 @@ export function bridgedProviderV2(config: BridgedConfig): Makefile {
   const sdk_nodejs_gen: Target = {
     name: "sdk/nodejs/.gen.sentinel",
     autoTouch: true,
-    dependencies: [bin_tfgen, "$(OVERLAYS_NODEJS)"],
-    commands: ["bin/$(TFGEN) nodejs --out sdk/nodejs/"],
+    dependencies: [provider_schema, "$(OVERLAYS_NODEJS)"],
+    commands: ["pulumi package gen-sdk provider/cmd/$(PROVIDER)/schema.json --language nodejs --out sdk/nodejs/"],
   };
   const sdk_nodejs_build: Target = {
     name: "sdk/nodejs/.build.sentinel",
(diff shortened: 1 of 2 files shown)


provider-ci/src/make-bridged-provider-2.ts
```diff
@@ -163,8 +163,8 @@ export function bridgedProviderV2(config: BridgedConfig): Makefile {
   const sdk_nodejs_gen: Target = {
     name: "sdk/nodejs/.gen.sentinel",
     autoTouch: true,
-    dependencies: [bin_tfgen, "$(OVERLAYS_NODEJS)"],
-    commands: ["bin/$(TFGEN) nodejs --out sdk/nodejs/"],
+    dependencies: [provider_schema, "$(OVERLAYS_NODEJS)"],
+    commands: ["pulumi package gen-sdk provider/cmd/$(PROVIDER)/schema.json --language nodejs --out sdk/nodejs/"],
```
danielrbradley, line 167: How do we specify which version of codegen we want to run?
1. We need reproducable builds
2. We need to be able to revert if the latest version has regressions

Frassle: It currently uses the version of codegen baked into the CLI, given your setting up your workflows with the `pulumi/setup-pulumi@v2` action you can just fix the version of that with:
```
  with:
    pulumi-version: 3.46.1
```

I accept this is a pretty lame way to manage this right now. I'm looking at moving codegen logic so it lives in the runtime plugins instead of in the CLI (yes it really is linked directly into the CLI right now), and making it so we can download different versions of the runtime plugins, that would then let you do something like:
`pulumi package gen-sdk provider/cmd/$(PROVIDER)/schema.json --language nodejs@3.46.1 --out sdk/nodejs/"],`

If you _need_ that feature before this could merge let us know, we can swap some priorities around.

danielrbradley: Yeah, we're going to need this before we can switch. Merging this would result in non-reproducable builds locally as the output would depend on which version of the pulumi CLI the user has installed locally. The only workaround would be to add extra machinery to download a repo-local version of the pulumi CLI, then use that rather than the version on the path, but I'd much rather avoid the hacks and add this as a feature of the codegen plugin.

Another aspect to consider here is how this version pinning could fit into our automation story too ... we need to be able to automatically bump dependency versions when they become available. This would probably stick to using a GH workflow rather than dependabot, but we need an easy way to search for the latest version and update the "lock" file format that we come up with for the plugin version. I.e hard coding the version in the command isn't going to work because it would be painful to update that command. 

These problems were why I also raised the idea a while back of just wrapping and publishing these binaries via NPM so we can just leverage all the existing ecosystem of lock files, update workflows and dependabot.

PR #287, opened 2022-11-18
Give the ci-mgmt commit in rollout PRs

squaremo:
There's no (obvious) link established between PRs created in provider repos, and the state of this repo at the time. This makes it difficult, for example, to pin failures downstream to any particular change here.

To be more explicit, include $GITHUB_SHA in the PR message.

Testing CI is always tricky, but I _think_ a demonstration that it works may be possible, by triggering a workflow from this branch (then rejecting the PR it creates downstream).


AaronFriel (approved):

squaremo:
@AaronFriel I'm not sure I had this working -- sorry for not keeping it as a draft 😳 The trials I did (by invoking the action directly) resulted in empty values for the SHA. I think it can be made to work, I just hadn't hit on the correct expression for getting the git SHA.

PR #292, opened 2022-11-22
Fix update upstream provider workflow to be resilient to redirects

viveklak:
The curl command used was not resilient to redirects. For instance `curl https://api.github.com/repos/databrickslabs/terraform-provider-databricks/git/ref/tags/v1.6.5` redirects:

```
{
  "message": "Moved Permanently",
  "url": "https://api.github.com/repositories/240289534/git/ref/tags/v1.6.5",
  "documentation_url": "https://docs.github.com/v3/#http-redirects"
}
```

Which would cause the subsequent jq to command to get a `null` value.

The `-L` flag makes sure we follow redirects.

### DIFF
--- a/provider-ci/providers/databricks/config.yaml
+++ b/provider-ci/providers/databricks/config.yaml
@@ -10,6 +10,6 @@ provider-default-branch: main
 makeTemplate: bridged
 plugins:
   - name: random 
-    version: "4.3.1"
+    version: "4.8.2"
   - name: aws 
-    version: "5.1.0"
+    version: "5.18.0"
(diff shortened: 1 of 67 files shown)


provider-ci/providers/databricks/config.yaml
```diff
@@ -10,6 +10,6 @@ provider-default-branch: main
 makeTemplate: bridged
 plugins:
   - name: random 
-    version: "4.3.1"
+    version: "4.8.2"
   - name: aws 
-    version: "5.1.0"
+    version: "5.18.0"
```
viveklak, line 15: These versions match what is in the repository.

guineveresaenger (approved):

PR #300, opened 2022-12-06
Fix to allow safe automerge

### LINKED ISSUE #272: Re-enable automerge of ci-mgmt workflows if safe to do so
### What happened?

We had to pause auto-merge of ci-mgmt PRs in https://github.com/pulumi/ci-mgmt/pull/265 and https://github.com/pulumi/ci-mgmt/pull/267 because these were being merged despite build failures. However, we need ci-mgmt to be synced to the relevant providers automatically to avoid issues due to workflows being stale (e.g. pulumi-azure has not yet synced to remove java publish as a blocking step: https://github.com/pulumi/pulumi-azure/blob/master/.github/workflows/release.yml#L351

We need to investigate and ensure automerge can be re-enabled safely at the earliest for the above drift to no longer occur.

### Steps to reproduce

See above.

### Expected Behavior

Automerge occurs only when all of CI passes.

### Actual Behavior

Automerge is currently paused because it was steamrolling despite CI failures.

### Output of `pulumi about`

_No response_

### Additional context

_No response_

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

viveklak:
Fixes #272

Previously, automergable PRs had automerge enabled externally immediately after the PR was created assuming that only if the sentinel check passed, the PR would work. However, in case of failure in tests the sentinel check is skipped entirely causing the PR to be immediately merged when we don't want.

This change tries to enable automerge from within the sentinel step which will execute when the steps it depends on are successful.

### DIFF
--- a/.github/workflows/update-workflows-bridged-providers.yml
+++ b/.github/workflows/update-workflows-bridged-providers.yml
@@ -63,9 +63,9 @@ jobs:
           title: "Update GitHub Actions workflows."
           path: pulumi-${{ matrix.provider }}
           token: ${{ secrets.PULUMI_BOT_TOKEN }}
-      # - name: Set PR to auto-merge
-      #   if: steps.create-pr.outputs.pull-request-operation == 'created' && github.event.inputs.automerge == 'true'
-      #   run: "gh pr merge --auto --squash ${{ steps.create-pr.outputs.pull-request-url }}"
+      - name: Add auto-merge label to PR
+        if: steps.create-pr.outputs.pull-request-operation == 'created' && github.event.inputs.automerge == 'true'
+        run: "gh pr edit ${{ steps.create-pr.outputs.pull-request-url }} --add-label 'auto-merge'"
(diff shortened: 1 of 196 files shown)


viveklak:
> One side-effect of checking the user who opened the PR is that if anyone adds extra commits on the end, it'll then auto-merge.
> 
> We also no-longer disable auto-merge if there's breaking changes.

This change is a bit more myopically done for the long-tail TF providers. AFAICR there was no such guardrail in place for these providers around breaking change detection.  

> 
> I think the safer approach would be to also introduce an `auto-merge` tag which can be added immediately and checked in the enable-automerge step. We could also then later adapt the schema check to remove the tag if there's breaking changes found.
> 

Yup - I wanted to test things out with this approach which seemed to work based on a couple of toy things I ran. I will add the auto-merge model to this PR.

t0yv0:
Closing in favor of https://github.com/pulumi/ci-mgmt/issues/311

PR #306, opened 2023-01-06
Install .NET 6 on runners

t0yv0:
Anticipating the upgrade to .NET6 , ensure that it is available on runners.

### DIFF
--- a/provider-ci/platform/examples/repo/.github/workflows/cron.yml
+++ b/provider-ci/platform/examples/repo/.github/workflows/cron.yml
@@ -44,7 +44,9 @@ jobs:
       fail-fast: false
       matrix:
         dotnetversion:
-        - 3.1.301
+        - |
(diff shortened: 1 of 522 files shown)


t0yv0:
I've tested this change on pulumi-tls. Full upgrade instructions in https://github.com/pulumi/home/issues/2531

lblackstone (approved):
Just to make sure I understand the change, this adds .NET 6.0 to the test matrix, but doesn't change the provider dependencies?

t0yv0:
This is a preparatory change it should be no-op for existing code. It makes sure .NET 6.0 is available in the GHA environment. Once this is applied, providers may continue upgrading to newer pulumi/pkg that will actually update provider .NET SDK dependency to .NET 6.0.

t0yv0:
Thanks for the quick review @lblackstone

PR #307, opened 2023-01-11
Add scripts to automatically edit provider source code

t0yv0:
Automates (crudely) common tasks such as upgrading .NET to 6.0 and Pulumi reference to 3.x.x across providers.

Testing:

- https://github.com/pulumi/pulumi-mongodbatlas/pull/168
- https://github.com/pulumi/pulumi-postgresql/pull/175
- https://github.com/pulumi/pulumi-opsgenie/pull/83
 
This has been partially rolled out in https://github.com/pulumi/home/issues/2531 and helped auto-convert a lot of repos.


t0yv0:
Friendly ping - can I get this reviewed please.

tools/sourcemigrator/index.ts
```diff
@@ -0,0 +1,101 @@
...
+interface MigrateContext {
+    dir: string;
+}
+
+interface MigrateResult {
+    filesEdited: Number;
+}
+
+function updateExamplesToDotNet6(): SourceMigration {
+    let pattern = new RegExp("[<]TargetFramework[>]netcoreapp3.1[<][/]TargetFramework[>]");
```
guineveresaenger, line 30: Do we know that each and every example we want to update is specifically `netcoreapp3.1`?

tools/sourcemigrator/index.ts
```diff
@@ -0,0 +1,101 @@
...
+    let replacement = "<TargetFramework>net6.0</TargetFramework>"
+    let sm: SourceMigration = {
+        name: "updateExamplesToDotNet6",
+        execute: (ctx: MigrateContext) => {
+            let stdout = child.execSync("git ls-files examples", {cwd: ctx.dir});
```
guineveresaenger, line 35: curious - is there a specific reason to use git here?

tools/sourcemigrator/index.ts
```diff
@@ -0,0 +1,101 @@
...
+            let filesEdited = String(stdout).split("\n")
+                .filter(x => x.startsWith("examples"))
```
guineveresaenger, line 37: Do we need this filter? We already have all the files from passing the `examples` folder above.

guineveresaenger (changes requested):
These source migrations are by nature very context specific, and in the advent of a pulumi v4.x.x or a new node version, etc, we would need to expand on this.
Therefore I would like there to be an extra section in the README that denotes the process - it doesn't have to be huge, just a lightweight "here's how you use this repo to migrate examples to a new version" - the README is certainly where I'd look for it.
 
Do you have a sense of how many providers would fail in CI as a result of these upgrades?
In general, I'm very much a fan! Loud and clear CI failures are great.

t0yv0:
> Do you have a sense of how many providers would fail in CI as a result of these upgrades?

I've rolled this out and it fixed ~10 builds. People have been fixing it by hand in a few places.

tools/sourcemigrator/index.ts
t0yv0: Yes - the migration is very targeted - it only upgrades netcoreapp3.1 to dotnet6 as we know this kind of upgrade requires no further action. I can perhaps make it more prominent in the name of the migration.

tools/sourcemigrator/index.ts
t0yv0: This avoids looking at non-source files such as node_modules etc.

tools/sourcemigrator/index.ts
t0yv0: Ah fair point, leftover from refactoring.

guineveresaenger (approved):
Thank you for adding the instructions!

PR #308, opened 2023-01-11
Put VERSION_PREFIX=4.0.0 in akamai

t0yv0:


t0yv0:
It turned out this is redundant.

PR #309, opened 2023-01-12
Install .NET 6 for native provider CI

t0yv0:
Native providers CI support for .NET 6 in the runner. 

This has been rolled out except for azure-native, see:

https://github.com/pulumi/home/issues/2531


t0yv0:
Friendly nudge @danielrbradley  - really appreciate if we could get this in sooner than later; I need this as part of rolling out #301 in more places.

t0yv0:
Oops, reopening after an accidental close.

danielrbradley (approved):

PR #310, opened 2023-01-12
Update required steps for automerge workflow

t0yv0:
Following https://github.com/pulumi/ci-mgmt/pull/301 , this locks GitHub branch protection to not auto-merge until all the important jobs have passed.

Pondering rollout order here. Perhaps before merging, we need to make sure that all the providers have been upgraded with the results of 301? If we merge this does send some providers into a state where no PRs can't merge until they catch up with 301?

### DIFF
--- a/infra/providers/index.ts
+++ b/infra/providers/index.ts
@@ -12,6 +12,19 @@ const providers = [...tfProviders, ...nativeProviders];
 for (let provider of providers) {
     const contexts: string[] = [
         "Update Changelog",
+        "prerequisites",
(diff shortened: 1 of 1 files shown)


danielrbradley (approved):
Historically, we've not checked for lint, but it might not be a bad inclusion.

t0yv0:
Opting to roll this out - I've notified the teams in Slack of potential undesirable intermediate states of getting unmerge-able PRs. Hoping this is relatively safe, we can roll back if needed.

PR #316, opened 2023-01-26
Source migrator to upgrade go to 1.19

t0yv0:
Some of our repositories are on Go 1.14 or some such. This contributes to broken build/test issues. This source migration detects go.mod and makes sure it's 1.19. It does not touch go.mod generated by pulumi/pkg such as sdk/python/go.mod that are installed as placeholders (those are not real go modules). For every go module it updates, it runs `go mod tidy` to re-sync `go.sum`.

To make this work, trigger Actions as needed, they will stand up PRs that apply the migrations as part of the PR.

### DIFF
--- a/tools/sourcemigrator/index.ts
+++ b/tools/sourcemigrator/index.ts
@@ -58,6 +58,29 @@ function updatePulumiCoreRefTo3x(): SourceMigration {
     return sm;
 }
 
+function updateGoTo_1_19(): SourceMigration {
+    let pattern = new RegExp('^go \\d+[.]\\d+$', 'm');
+    let replacement = "go 1.19";
+    let sm: SourceMigration = {
+        name: "updateGoTo_1_19",
+        execute: (ctx: MigrateContext) => {
+            let stdout = child.execSync("git ls-files -- '**/go.mod'", {cwd: ctx.dir});
+            let filesEdited = String(stdout).split("\n")
+                .filter(x => x.endsWith("go.mod"))
+                .filter(x => {
+                    let f = path.join(ctx.dir, x);
+                    let replaced = replaceInFile(f, pattern, replacement);
+                    if (replaced) {
+                        child.execSync("go mod tidy", {cwd: path.dirname(f)});
+                    }
+                    return replaced;
+                }).length;
+            return {filesEdited: filesEdited};
+        },
+    };
+    return sm;
+}
+
 function replaceInFile(f: string, pattern: RegExp, replacement: string): boolean {
     let contents = String(fs.readFileSync(f));
     if (pattern.test(contents)) {
@@ -83,7 +106,8 @@ function runMigrations(context: MigrateContext, migrations: SourceMigration[]) {
 function allMigrations(): SourceMigration[] {
     return [
         updateExamplesFromCore31DotNet6(),
-        updatePulumiCoreRefTo3x()
+        updatePulumiCoreRefTo3x(),
+        updateGoTo_1_19(),
     ];
 }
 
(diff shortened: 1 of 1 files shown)


tools/sourcemigrator/index.ts
```diff
@@ -58,12 +58,45 @@ function updatePulumiCoreRefTo3x(): SourceMigration {
...
+}
+
 function replaceInFile(f: string, pattern: RegExp, replacement: string): boolean {
     let contents = String(fs.readFileSync(f));
     if (pattern.test(contents)) {
         let updatedContents = contents.replace(pattern, replacement)
-        fs.writeFileSync(f, updatedContents);
-        return true;
+        if (updatedContents != contents) {
+            fs.writeFileSync(f, updatedContents);
```
t0yv0, line 95: Fix here for stats. Only count file as edited if updatedContents != contents.

tools/sourcemigrator/index.ts
```diff
@@ -58,12 +58,45 @@ function updatePulumiCoreRefTo3x(): SourceMigration {
     return sm;
 }
 
+function updateGoTo_1_19(): SourceMigration {
```
jazzyfresh, line 61:
```suggestion
function updateGo_1_19(): SourceMigration {
```

tools/sourcemigrator/index.ts
```diff
@@ -58,12 +58,45 @@ function updatePulumiCoreRefTo3x(): SourceMigration {
...
+    let pattern = new RegExp('^go \\d+[.]\\d+$', 'm');
+    let replacement = "go 1.19";
+    let sm: SourceMigration = {
+        name: "updateGoTo_1_19",
```
jazzyfresh, line 65:
```suggestion
        name: "updateGo_1_19",
```

tools/sourcemigrator/index.ts
```diff
@@ -83,7 +116,8 @@ function runMigrations(context: MigrateContext, migrations: SourceMigration[]) {
 function allMigrations(): SourceMigration[] {
     return [
         updateExamplesFromCore31DotNet6(),
-        updatePulumiCoreRefTo3x()
+        updatePulumiCoreRefTo3x(),
+        updateGoTo_1_19(),
```
jazzyfresh, line 120:
```suggestion
        updateGo_1_19(),
```

jazzyfresh (approved):
Made a suggestion on the name bc the "goto" confused me for a moment, but it's not a necessary edit.

LGTM!

squaremo:
> Some of our repositories are on Go 1.14 or some such.  This contributes to broken build/test issues.

What's the problem with this, can you give an example? It seems unlikely that all the providers suddenly require Go 1.19, e.g., they use go generics.

t0yv0:
Good question! 

I haven't added a reference here so the insinuation that it causes broken tests may not be factually accurate.

If there's an issue this code is causing in ci-mgmt we can very well disable it.

The context around the source-migration scripts was that the ecosystem team has a 50-60 repos with various degrees of drift, and the idea was to use some possibly ad-hoc tooling to reduce the drift and normalize these repos to a modern and uniform state via auto-merge PRs that run this tooling.

@aq17 you were asking yesterday about migrating to Go 1.19 do you have a motivating use case?

The more pernicious kind of problem in this same category is examples that depend on outdated Pulumi SDK, say 2.x.x instead of 3.x.x.

squaremo:
> If there's an issue this code is causing in ci-mgmt we can very well disable it.

I should have been clearer about why i was here :-) I was investigating downstream changes (https://github.com/pulumi/pulumi-azure/pull/1225#pullrequestreview-1319807715), reached this PR, and found the explanation was a dead end -- "citation needed" :-) It's pretty low stakes.
On the other hand, sometimes there are high stakes -- "we need to get the release out, and such-and-such change seems to be blocking it, and we don't know why it was made!!" So, we could look at this as a gentle signal that spending a bit of time giving a full explanation is worthwhile.

PR #319, opened 2023-01-27
Simplify major provider upgrades

### LINKED ISSUE #318: Allow providers to perform major version updates without ci-mgmt PRs
Updating a major version in a provider currently requires four or more pull requests, as shown with recent updates to Keycloak.

See: 
- https://github.com/pulumi/pulumi-keycloak/pull/162
- https://github.com/pulumi/pulumi-keycloak/pull/163
- https://github.com/pulumi/ci-mgmt/pull/315
- https://github.com/pulumi/pulumi-keycloak/pull/164

Even major version updates should follow a normal release process for a software project: a code change followed by a release ceremony. For our providers, this would be a pull request and a `@release-bot` automation.

A pull request was made to Keycloak to demonstrate changes to ci-mgmt outputs that we'd have to make to support this:
- https://github.com/pulumi/pulumi-keycloak/pull/165

If these changes were in place, three of the four pull requests would not have been needed to update Keycloak to v5.

iwahbe:
Fixes #318 

Reviewing this PR should be done by commit. The first commit contains code changes. The second commit contains a single regenerated provider (keycloak, to match https://github.com/pulumi/pulumi-keycloak/pull/165/files). The third commit contains a full regeneration of all providers.


AaronFriel:
Left a note on the keycloak PR used as a reference, I think we'll need ci-mgmt to create a go.mod and go.sum for `provider/pkg/version`, and fortunately both will be static as the module has no dependencies.

AaronFriel (changes requested):
Making a request changes block based on my previous comment.

guineveresaenger (approved):
For each provider, we'll still need to make sure all the example dependencies are replaced with the local SDK, correct?

AaronFriel:
Perhaps the code modification tool @t0yv0 has used can help us avoid busy-work there.

provider-ci/providers/aws/repo/Makefile
```diff
@@ -3,8 +3,7 @@
 PACK := aws
 ORG := pulumi
 PROJECT := github.com/$(ORG)/pulumi-$(PACK)
-PROVIDER_PATH := provider/v5
-VERSION_PATH := $(PROVIDER_PATH)/pkg/version.Version
+VERSION_PATH := $(PROJECT)/provider/pkg/version.Version
```
AaronFriel, line 6: Let's run this by @danielrbradley & @kpitzen before merging.

provider-ci/providers/keycloak/repo/provider/pkg/version/go.mod
```diff
@@ -0,0 +1,5 @@
+# WARNING: This file is autogenerated - changes will be overwritten if not made via https://github.com/pulumi/ci-mgmt
+
+module github.com/pulumi-keycloak/provider/pkg/version
```
AaronFriel, line 3: This is tricky - this path was previously imported with `provider/vX/pkg/version`.

I think we need to apply this with a code modification change at the same time, to find-replace `provider/vX/pkg/version` with `provider/pkg/version`

AaronFriel (changes requested):
Apologies for keeping this on block.

Let's make sure Providers team reviews changes to their makefiles. 

Then let's figure out how we can roll this change out with one PR, using the codemod work @t0yv0 did. These changes include:

```patch
# provider/go.mod
+ replace github.com/pulumi/pulumi-keycloak/provider/pkg/version => ./pkg/version
+ 
+ require github.com/pulumi/pulumi-keycloak/provider/pkg/version v0.0.0
```

```patch
# provider/resources.go
-	"github.com/pulumi/pulumi-keycloak/provider/v5/pkg/version"
+	"github.com/pulumi/pulumi-keycloak/provider/pkg/version"
```

```patch
# provider/cmd/pulumi-tfgen-keycloak/main.go
-	"github.com/pulumi/pulumi-keycloak/provider/v5/pkg/version"
+	"github.com/pulumi/pulumi-keycloak/provider/pkg/version"
```

```patch
# provider/cmd/pulumi-resource-keycloak/main.go
-	"github.com/pulumi/pulumi-keycloak/provider/v5/pkg/version"
+	"github.com/pulumi/pulumi-keycloak/provider/pkg/version"
```

provider-ci/src/goreleaser.ts
```diff
@@ -96,22 +96,11 @@ export class PulumiGoreleaserPreConfig implements GoreleaserConfig {
...
-      ldflags = [
-        `-X github.com/pulumi/pulumi-${opts.provider}/provider/v${opts["major-version"]}/pkg/version.Version={{.Tag}}`,
-      ];
-    } else {
-      ldflags = [
-        `-X github.com/pulumi/pulumi-${opts.provider}/provider/pkg/version.Version={{.Tag}}`,
-      ];
-    }
+    const ldflags = [
+      `-X github.com/pulumi/pulumi-${opts.provider}/provider/pkg/version.Version={{.Tag}}`,
```
t0yv0, line 102: Confused - will this really work? Doesn't e.g. V5 require the path to be "provider/v5/pkg"?

Oh... version package is now upgraded to its own module. Wicked.

```
module github.com/pulumi-keycloak/provider/pkg/version
```

t0yv0:
Hey folks I've reviewed pulumi-keycloak changes.

1. ensure provider/pkg/version/go.mod created if not exists

2. provider/go.mod receives
    github.com/pulumi/pulumi-keycloak/provider/pkg/version => ./pkg/version
    require github.com/pulumi/pulumi-keycloak/provider/pkg/version v0.0.0

3. provider/resources.go receives

```
    -       "github.com/pulumi/pulumi-keycloak/provider/v5/pkg/version"
    +       "github.com/pulumi/pulumi-keycloak/provider/pkg/version"
```

   And the same change applies for all cmd/*..go etc source files in the provider.

4. In examples, lots of replacements; in particular replace

```
-       "github.com/pulumi/pulumi-keycloak/sdk/v4/go/keycloak"
+       "github.com/pulumi/pulumi-keycloak/sdk/go/keycloak"
```

I'm only really concerned with (4) as a change we might not want to make everywhere.. The examples are supposed to be something that users can copy-paste and run against the latest released provider, so that needs to include `v4` or `v5`. As provider developers we want to be able to run against a local SDK, which is now inconvenient but it's a secondary scenario. Can we bring the v4 etc back in the examples? 

The 1-3 chores are easy to script/automate in sourcemigrator. Perhaps also a replacement or 4 that moves "v4" to "v5" or some such automatically can be scripted also with some ease. Thoughts?

I wouldn't block on scripting the source migrations. This seems like it helps, get this in and improve as we go. Just would like to clarify if we want the examples change.

AaronFriel:
Doesn't #4 prevent us from having to "disable tests due to major version update", or is that unrelated?

The playbook has us setting options like this in major version PRs:

```go
	t.Skip("ignoring due to major version change")
```

But now I'm wondering why we need that if the SDK is linked correctly via `ProgramTestOptions.Dependencies`.

AaronFriel:
https://github.com/pulumi/platform-providers-team/blob/93e5538058a760d1602cbf5994fe1126faa26e10/playbooks/tf-provider-major-version-update.md

> In examples/, temporarily skip any acceptance tests for Go SDKs. We currently lack the ability to maintain working acceptance tests when performing a major version update:

Why is that, I wonder.

AaronFriel:
@t0yv0 There is a case for using `/sdk` (versionless) as the replacement, or even something like `sdk/v9000` or a fake module name like `sdk/current`. With ProgramTest we can test step 1 using `sdk/v4`, and step 2 using `sdk/current` with expect no changes.

That said, I think the same thing could be accomplished by adding a `Dependencies` field to the `EditDirs` config struct, so that each step can have its own overrides. 

What do you think?

t0yv0:
I think this is a regression with the current approach for the end-users interacting with the repo: https://github.com/pulumi/pulumi-keycloak/issues/166

That's my one worry. If we can get that to work again, the rest I think is all fine by me - whichever way ProgramTest accomplishes re-linking to the in-source SDK.

Also wondering if we need alignment from providers team here or we can move forward? I thought their intent was to move off ci-mgmt?

AaronFriel:
I see your point on #4 and making sure examples are copy-able. That makes sense to me, and we can make an issue to augment ProgramTest to do the per-step dependency replacement rules.

t0yv0:
I'm not sure where these statements are coming from TBH so if you could elaborate in an issue or discuss some more here it could be helpful 🙏 

```
temporarily skip any acceptance tests for Go SDKs. We currently lack the ability to maintain working acceptance tests when performing a major version update:
```

```
There is a case for using /sdk (versionless) as the replacement, or even something like sdk/v9000 or a fake module name like sdk/current. With ProgramTest we can test step 1 using sdk/v4, and step 2 using sdk/current with expect no changes.
```

I can see that cross-version testing like https://github.com/pulumi/ci-mgmt/issues/325 can get complicated, but AFAIK we do not carry any such tests at the moment, and simple tests on one version should just work? Is the issue about not upgrading the references in the examples source code?

Perhaps I'm missing some detail here, but I would very much rather err on the side of doing regex/gofmt replacements or manual edits at the time of the (hopefully rate) major version upgrade event than carry extra switches in the provider code itself. When upgrading from v2 to v3 if we could just edit the examples and ProgramTest dependencies to now point to v3, wouldn't this be OK?

AaronFriel:
@t0yv0

Re: 
> temporarily skip any acceptance tests for Go SDKs. We currently lack the ability to maintain working acceptance tests when performing a major version update:

I don't know the origin of this or why it's in the playbook. I think it's false.

> There is a case for using /sdk (versionless) as the replacement, or even something like sdk/v9000 or a fake module name like sdk/current. With ProgramTest we can test step 1 using sdk/v4, and step 2 using sdk/current with expect no changes.

I'm fine retracting this in favor of your argument on (4.) above. My argument was that it's _nice_ as a provider author to be able to be able to update the examples to target new SDKs without having to change the examples dir, and that the version numbers in examples dir will go stale if we don't check them carefully during major version changes. I think they'll go stale because `go.mod` files in the examples dir should have replace directives so that `gopls` and other Go utilities are well behaved (e.g.: so we can run `go mod tidy` on them or `go get -u`):

```
replace github.com/pulumi/pulumi-$PROVIDER/sdk/v3 => ../../sdk
```

t0yv0:
Perfect. I think we're on the same page then. I see your point, for provider author this is tough, but if the users's needs that come first; so perhaps we can compensate for provider author pain with tooling that makes it easier.

- `go mod` CLI has options to edit replacements in go.mod files
-  `gofmt` CLI has options to safely edit Go source trees

t0yv0:
@iwahbe does it make sense to update this PR to take this into account? The other blocker here is we need to sort out if we need alignment from providers team or there's a way we can roll this out to our subset. We'll get the answers to that on Thu.

iwahbe:
I think we can automate (https://github.com/pulumi/upgrade-provider/pull/15) changing examples, and keep full value for customers. This way we don't need to deal updating the examples ourselves, and our users get full copy ability.

AaronFriel:
@iwahbe make sure we document (1) in the playbook, and can we track (2) in an issue?

t0yv0:
Still a bit out alignment here. I'm on board with this:

> Leave full paths in examples, updating them with upgrade-provider to ensure we get coverage.

The second bit feels off to me. Feels too clever, since the provider build would now depend on the state of releases, this is a strange loop dependency.  I'd prefer the major version to be hardcoded, just not in the ci-mgmt but inline in the repo.. say if Makefile could read it from a text file and pass to pulumictl we'd be set. I think I've documented this suggestion in the discussion to https://github.com/pulumi/ci-mgmt/issues/318

> Instead of encoding major versions in ci-mgmt, dynamically calculate them by checking the latest release tag of the relevant provider during build-time. This way, we can launch an "Update GH Workflows" PR immediately after release and it will work. No PR to this repo necessary.

iwahbe:
> The second bit feels off to me. Feels too clever, since the provider build would now depend on the state of releases, this is a strange loop dependency. I'd prefer the major version to be hardcoded, just not in the ci-mgmt but inline in the repo.. say if Makefile could read it from a text file and pass to pulumictl we'd be set. I think I've documented this suggestion in the discussion to https://github.com/pulumi/ci-mgmt/issues/318

The provider build would not depend on the state of the release. The result of "Update GH Workflows" would depend on the state of the release. The idea here is that the repository is the single source of truth on its version number. 

It's not a dependency loop, since we don't initiate workflow updates as part of running releases. 

---

I'm happy to stick this in a file in the repo instead of a tag in the repo. We could honestly stick the `provider-ci/providers/${NAME}/config.yaml` into the provider repo with some known path like `providers/managed.yaml`. This way everything that ci-mgmt knows about `${NAME}` is stored in-repo, instead of centralized. ci-mgmt would only need to know that `${NAME}` should be managed.

t0yv0:
> The idea here is that the repository is the single source of truth on its version number

I'm a bit old-school perhaps but I like to be able to work locally and generate the results I want locally from a set of files regardless how they're arranged into repositories, whether it's a fork, etc, or I cloned the files somewhere out of GitHub, or using GitLab or using my own private repo. If I want to build for "v5.0.0" I want to be able to do that on my local copy w/o modifying the GitHub etc, even if last released version is "v3.0.0".

But in our current praxis it's more for of a nit because it's not going to matter a ton day-to-day which way we do this, and both strategies can be introduced later.

t0yv0:
Not advocating for bringing in providers/managed.yaml just let me be super specific about our current footgun. The generated files in Go SDK as well as Go examples currently *depend on* the major version. The major version is computed out of `pulumictl` floating version. The floating version depends on tags in the repo. Unless VERSION_PREFIX is set via ci-mgmt.

This dependency is the footgun.

The tags in your local repo may not match tags on CI. The computed version may not align on majors between local and CI. any time this is out of whack, tired people are wondering why they can run a test locally but uploading it to CI breaks in a weird way. 

Or why a change works on master, but when someone tags a release the same PR/commit does not pass CI anymore on restarts. I've been bitten by this more than one time.

I've seen providers folks remove out pulumictl entirely and work off a fixed version, but I didn't figure out how they instruct CI which version it's on. This can be interesting.

Or just pinning the major version in repo is interesting. 

In other words the ideal flow for me, when upgrading to a major version:

- edit the source code to indicate which new version I want
- re-generate generated files
- run examples locally - everything tests the new version
- open a PR - CI tests exactly the same thing as tested locally, passes
- merge and done
- no editing out of band ci-mgmt

That is CI validation status is a function of the source file-set, not tags or repo state.

t0yv0:
And yes normal languages do not mention the version anywhere at all in sources but sounds like Go is special in this regard :)

iwahbe:
I think we are talking passed each-other. I don't think I appropriately narrowed the scope of this statement:

> Instead of encoding major versions in ci-mgmt, dynamically calculate them by checking the latest release tag of the relevant provider during build-time.

I'm suggesting the above, only in regards to when `ci-mgmt` generates the set of managed files. This would change nothing about the build process. This would change nothing about the process of making a major upgrade PR. You would:

- edit the files locally
- regenerate the sdk
- run examples/tests
- open a PR
- merge
- tag (release)

The tag step will not change the output of earlier steps. 

All I'm saying, is when the "Upgrade GH Workflows" PR runs, it takes the major version from the repo it will merge into. In effect, I want [this line](https://github.com/pulumi/ci-mgmt/blob/8556261312864d9648cbcf2bc4ac04fba4c9da52/provider-ci/providers/cloudflare/config.yaml#L2) to be a calculated value. Thats it. I'm not suggesting any change to the process in a provider's repo.

PR #321, opened 2023-02-03
Add docs gen Makefile target for hybrid providers

guineveresaenger:
Hybrid providers are terraform-bridged Pulumi providers that have native resources added to them.
Our currently only example of this is the [Pulumi Docker Provider](https://github.com/pulumi/pulumi-docker).
All native resources need to generate their own documentation, and this PR encodes the process into any hybrid
provider Makefile.
The provider's `config.yaml` will need to contain the `docsCmd` field as well as the `hybrid` field in order to
generate a Makefile with a `docs` target.
See also: https://github.com/pulumi/pulumi-docker/pull/479

- Add logic to add a `docs` Makefile target
- generalize approach to denote a hybrid provider rather than docekr specifically
- Generate all providers. Updates major version for Fastly but does not add any extra targets to non-hybrid providers, as expected.

### DIFF
--- a/provider-ci/providers/docker/config.yaml
+++ b/provider-ci/providers/docker/config.yaml
@@ -11,3 +11,5 @@ env:
   PULUMI_GO_DEP_ROOT: /home/runner/work/pulumi-docker
 upstream-provider-org: kreuzwerker
 makeTemplate: bridged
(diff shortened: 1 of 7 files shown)


iwahbe (approved):
LGTM

PR #326, opened 2023-02-06
Add support for patched upstream providers

iwahbe:
Support is directly modeled after the digital ocean repo demoed in the design doc.

This PR adds 3 targets:

1. `upstream`: Ensure that the upstream folder (if it exists) reflects the patch set applied to the upstream commit.
2. `start-patch`: Prepare to create a new patch.
3. `finish-patch`: Set the current committed changes as a new patch.

### DIFF
--- a/provider-ci/providers/aiven/repo/Makefile
+++ b/provider-ci/providers/aiven/repo/Makefile
@@ -144,7 +144,12 @@ else ifeq ("$(wildcard patches/*.patch)","")
 	@echo "patches were expected because upstream exists"
 	@exit 1
 else
+	# Checkout the submodule at the pinned commit.
+	# `--force`: If the submodule is at a different commit, move it to the pinned commit.
+	# `--init`: If the submodule is not initialized, initialize it.
 	git submodule update --force --init
+	# Iterating over the patches folder in sorted order,
+	# apply the patch using a 3-way merge strategy. This mirrors the default behavior of `git merge`
 	cd upstream && \
 		for patch in $(sort $(wildcard patches/*.patch)); do git apply --3way ../$$patch || exit 1; done
 endif
--- a/provider-ci/providers/aiven/repo/Makefile
+++ b/provider-ci/providers/aiven/repo/Makefile
@@ -30,7 +30,7 @@ install_sdks: install_dotnet_sdk install_python_sdk install_nodejs_sdk install_j
 only_build: build
 
 build_dotnet: DOTNET_VERSION := $(shell pulumictl get version --language dotnet)
-build_dotnet:
+build_dotnet: upstream
 	pulumictl get version --language dotnet
 	$(WORKING_DIR)/bin/$(TFGEN) dotnet --out sdk/dotnet/
 	cd sdk/dotnet/ && \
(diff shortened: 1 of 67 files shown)


provider-ci/providers/aiven/repo/Makefile
```diff
@@ -99,15 +104,52 @@ lint_provider: provider
...
+	#	3. Commit your changes. The slugified first line of your commit description will
+	#	be used to generate the patch file name. Only the diff from the latest commit will
+	#	end up in the final patch.
+	#
+	#	4. Run `make finish-patch`.
+	#
+	# It is safe to run `make start-patch` as many times as you want, but any changes
+	# might be reverted until `make finish-patch` is run.
+	@cd upstream && git commit --quiet -m "existing patches"
+endif
```
iwahbe, line 128: This is what the `make start-patch` command looks like. If the instructions don't make sense, please let me know.

t0yv0:
I am wondering whether it would make sense to fork this already based on the ecosystem ownership lists. That is, this edits Makefiles for pulumi-aws but I'm not sure if that's problematic or not. Or can we just maybe drop pulumi-aws from the list of providers managed here? CC @danielrbradley @thomas11

danielrbradley:
@t0yv0 
> I am wondering whether it would make sense to fork this already based on the ecosystem ownership lists. That is, this edits Makefiles for pulumi-aws but I'm not sure if that's problematic or not. Or can we just maybe drop pulumi-aws from the list of providers managed here?

Yes this is potentially problematic - I think we should just drop all providers which are now managed by the providers team for the time being as we're going to implement a different system for managing them.

t0yv0:
That sounds great! If you're all ready to give up on ci-mgmt, we will just drop your providers right away from here.

iwahbe:
I've switched this feature off for repos managed by the providers team. Until we get consensus on how providers and ecosystem will share ci-mgmt, this allows us to move forward as is.

provider-ci/providers/cloudinit/repo/Makefile
```diff
@@ -99,15 +104,52 @@ lint_provider: provider
...
+	#	1. Run this command (`make start-patch`).
+	#
+	#	2. Edit the `upstream` repo, making whatever changes you want to appear in the new
+	#	patch. It's fine to edit multiple files.
+	#
+	#	3. Commit your changes. The slugified first line of your commit description will
+	#	be used to generate the patch file name. Only the diff from the latest commit will
+	#	end up in the final patch.
+	#
+	#	4. Run `make finish-patch`.
```
jazzyfresh, line 123: these are good docs that might also go in the readme or wherever provider build docs go

provider-ci/providers/cloudinit/repo/Makefile
```diff
@@ -99,15 +104,52 @@ lint_provider: provider
...
+	# upstream doesn't exist, so skip
+else ifeq ("$(wildcard patches/*.patch)","")
+	# upstream exists, but patches don't exist. This is probably an error.
+	@echo "No patches found within the patch operation"
+	@echo "patches were expected because upstream exists"
+	@exit 1
+else
+	git submodule update --force --init
+	cd upstream && \
+		for patch in $(sort $(wildcard patches/*.patch)); do git apply --3way ../$$patch || exit 1; done
```
jazzyfresh, line 149: a comment here would be helpful, this is a bit tricky to understand if you don't already know what its supposed to do

jazzyfresh (approved):
left some feedback about documentation, but not a blocker. (i accidentally posted it to a generated file but comments are intended to apply to the template)

provider-ci/providers/cloudinit/repo/Makefile
iwahbe: My idea was that we wouldn't need to put this into a README, beyond you should run "make start-patch". This is a bash comment, not a make comment. It is printed to the screen every whenever `make start-patch` is run.

PR #328, opened 2023-02-07
Remove provider's managed by the provider's team

iwahbe:
In response to this comment: https://github.com/pulumi/ci-mgmt/pull/326#issuecomment-1421221715, we are removing providers that require bespoke CI management.


iwahbe:
Marking as WIP until internal consensus is reached.

PR #333, opened 2023-02-09
fix: Fix conditional check to see if pulumi binary is in PATH

rquitales:
Previous conditional check always returns true, even if the the pulumi binary does not exist. This commit fixes this logical check.

Ran `make` in the `provider-ci` folder after updating `src`.


provider-ci/src/make.ts
```diff
@@ -72,31 +69,36 @@ const indent = "\t";
...
-  return cond.parts.map((part) => part.test + "\n" + render(part.then)).join("\n") + "\n" + cond.end;
+function renderConditional<T>(
+  cond: Conditional<T>,
+  render: (v: T[]) => string
+): string {
+  return (
+    cond.parts.map((part) => part.test + "\n" + render(part.then)).join("\n") +
+    "\n" +
+    cond.end
+  );
```
AaronFriel, line 90: Nit: pull requests are harder to review when they contain unrelated coding style changes.

provider-ci/src/make-bridged-provider.ts
```diff
@@ -57,76 +57,83 @@ export function bridgedProvider(config: BridgedConfig): Makefile {
...
+                "# It is safe to run `make start-patch` as many times as you want, but any changes",
+                "# might be reverted until `make finish-patch` is run.",
+                `@cd upstream && git commit --quiet -m "existing patches"`,
+              ],
+            },
+          ],
+          end: "endif",
+        },
+      ],
+    };
```
AaronFriel, line 136: I'd like this commit reduced to the semantic changes.

AaronFriel (changes requested):
The PR looks good, but contains unrelated changes.

rquitales:
> The PR looks good, but contains unrelated changes.

That's probably due to running `make` (ie all targets). Have removed formatting changes in the last patchset.

iwahbe (approved):
New semantics looks good. I don't see how it affects `make.ts`, so it would be great if that change went away.

AaronFriel (approved):

AaronFriel:
@rquitales were you able to test this out on a repository? I'm noticing that with this change, `make install_plugins` *always* installs the Pulumi CLI:

```shell
$ make provider
make: command: Command not found
[ -x "" ] || curl -fsSL https://get.pulumi.com | sh
=== Upgrading Pulumi 3.50.2-dev.0 to v3.55.0 ===
...
```

When I revert this change, I get the expected behavior:

```shell
$ make install_plugins
[ -x /home/friel/.pulumi-dev/bin/pulumi ] || curl -fsSL https://get.pulumi.com | sh
# end of output, no installation
```

This is using GNU Make 4.2.1, Bash 5.0.17 on Ubuntu Linux 20.04.

rquitales:
@AaronFriel With the reverted change, I believe you would never install `pulumi` as it always returns true even if the `pulumi` binary is not found.

It seems like this change is always installing the CLI since `command` is not installed in that env. If `command` exists, then it won't install Pulumi CLI if it's already existing locally.

Looks like we're running into local env issues, where we can't assume that `command` comes preinstalled.

AaronFriel:
@rquitales `command` is not a program but a shell command. Does make behave correctly when shelling out to run shell commands which aren't binaries?

rquitales:
@AaronFriel seems to be working on both my Mac and a VM on GCP:

```
make install_plugins
[ -x "/opt/homebrew/bin/pulumi" ] || curl -fsSL https://get.pulumi.com | sh
# end of output, no installation
```

AaronFriel:
Interesting: https://stackoverflow.com/questions/12989869/calling-command-v-find-from-gnu-makefile

It looks like `make` can try to be clever.

AaronFriel:
Too clever: https://stackoverflow.com/questions/17547625/how-to-use-shell-builtin-function-from-a-makefile/17550243#17550243

PR #356, opened 2023-03-01
Add update-bridge-ecosystem-providers workflow

aq17:
Mirrors https://github.com/pulumi/ci-mgmt/blob/master/.github/workflows/update-workflows-bridged-providers.yml specifically for team-ecosystem-maintained providers


t0yv0 (approved):
Thanks!

PR #365, opened 2023-03-09
Ensure command -v runs in Make with default shells

AaronFriel:
Adding a shell redirection prevents the make target `install_plugins` from always reinstalling Pulumi.

Background
==========

Optimizations in `make(1)` result in `$(shell command -v SOME_COMMAND)` erroring. E.g.:

```sh
$ make install_plugins
make: command: Command not found
[ -x "" ] || curl -fsSL https://get.pulumi.com | sh
```

This is due to an optimization `make(1)` that avoids invoking a shell and instead attempting to run the command given to `$(shell ...)` directly. However, `command` is not a program that can be run, it's a built-in command of the shell. This results in the error `make: command: Command not found`.

It's the difference between running `exec("command", "-v", ...)` and `exec("sh", "-c", "command -v ...")`.

Relevant StackOverflow threads:
* https://stackoverflow.com/questions/12989869/calling-command-v-find-from-gnu-makefile
* https://stackoverflow.com/questions/17547625/how-to-use-shell-builtin-function-from-a-makefile/17550243#17550243

### DIFF
--- a/provider-ci/providers/aiven/repo/Makefile
+++ b/provider-ci/providers/aiven/repo/Makefile
@@ -96,7 +96,7 @@ install_nodejs_sdk:
 	yarn link --cwd $(WORKING_DIR)/sdk/nodejs/bin
 
 install_plugins:
-	[ -x "$(shell command -v pulumi)" ] || curl -fsSL https://get.pulumi.com | sh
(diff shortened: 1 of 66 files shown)


t0yv0 (approved):

rquitales (approved):
Thanks for digging into this fix!

PR #369, opened 2023-03-13
Upgrade ns1 provider to v3

aq17:


aq17:
Need https://github.com/pulumi/pulumi-ns1/pull/131 to land first

PR #380, opened 2023-03-21
Allow user agent to be set via ldflags for `pulumi-newrelic`

aq17:
Fixes https://github.com/pulumi/pulumi-newrelic/issues/361

### DIFF
--- a/provider-ci/providers/newrelic/repo/.goreleaser.prerelease.yml
+++ b/provider-ci/providers/newrelic/repo/.goreleaser.prerelease.yml
@@ -29,6 +29,8 @@ builds:
   ignore: []
   ldflags:
   - -X github.com/pulumi/pulumi-newrelic/provider/v5/pkg/version.Version={{.Tag}}
+  - -X
+    github.com/newrelic/terraform-provider-newrelic/v2/main.UserAgentServiceName=pulumi
   main: ./cmd/pulumi-resource-newrelic/
 changelog:
   skip: true
--- a/provider-ci/providers/newrelic/repo/.goreleaser.yml
+++ b/provider-ci/providers/newrelic/repo/.goreleaser.yml
@@ -29,6 +29,8 @@ builds:
   ignore: []
   ldflags:
   - -X github.com/pulumi/pulumi-newrelic/provider/v5/pkg/version.Version={{.Tag}}
+  - -X
+    github.com/newrelic/terraform-provider-newrelic/v2/main.UserAgentServiceName=pulumi
   main: ./cmd/pulumi-resource-newrelic/
 changelog:
   filters:
--- a/provider-ci/src/goreleaser.ts
+++ b/provider-ci/src/goreleaser.ts
@@ -83,12 +83,8 @@ interface GoReleaserOpts {
   "major-version": number;
   providerVersion: string;
   skipTfGen: boolean;
-  "extra-ld-flags": boolean;
+  "extra-ld-flags"?: string[];
 }
-
-const extraLdFlags = new Map<string, string[]>([
-    ['newrelic', [`-X github.com/newrelic/terraform-provider-newrelic/v2/main.UserAgentServiceName=pulumi`]],
-]);
 
 export class PulumiGoreleaserPreConfig implements GoreleaserConfig {
   name?: string;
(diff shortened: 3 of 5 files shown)


jazzyfresh (approved):

provider-ci/providers/newrelic/repo/.goreleaser.prerelease.yml
```diff
@@ -29,6 +29,7 @@ builds:
   ignore: []
   ldflags:
   - -X github.com/pulumi/pulumi-newrelic/provider/v5/pkg/version.Version={{.Tag}}
+  - -X main.UserAgentServiceName={{.Tag}}
```
AaronFriel, line 32: Will regenerating the repository files overwrite this?

AaronFriel, line 32: I think the exact ldflag will need to be tweaked, as well. It should be the full path to the module, package, and variable name. The value should also be "pulumi", I think: https://github.com/pulumi/pulumi-newrelic/issues/361

aq17: I guess we need to add another file, i.e. `useragentservicename.go`, in `pulumi-newrelic` also

provider-ci/src/goreleaser.ts
```diff
@@ -113,6 +113,10 @@ export class PulumiGoreleaserPreConfig implements GoreleaserConfig {
       ];
     }
 
+	if (opts.provider === "newrelic") {
+		ldflags.push(`-X github.com/pulumi/pulumi-${opts.provider}/provider/v${opts["major-version"]}/pkg/useragentservicename.UserAgentServiceName="pulumi"`)
```
AaronFriel, line 117: I think we would want to update the newrelic module, not pulumi-newrelic: 


```suggestion
		ldflags.push(`-X github.com/newrelic/terraform-provider-newrelic/v2/main.UserAgentServiceName=pulumi`)
```

Could we write a unit test to verify that works, since `main.UserAgentServiceName` is public?

provider-ci/providers/newrelic/repo/.goreleaser.yml
```diff
@@ -29,6 +29,8 @@ builds:
   ignore: []
   ldflags:
   - -X github.com/pulumi/pulumi-newrelic/provider/v5/pkg/version.Version={{.Tag}}
+  - -X
```
t0yv0, line 32: This might need to be on a single line not two.

provider-ci/src/goreleaser.ts
```diff
@@ -113,6 +113,10 @@ export class PulumiGoreleaserPreConfig implements GoreleaserConfig {
       ];
     }
 
+	if (opts.provider === "newrelic") {
```
t0yv0, line 116: Something funky with indent here? also 

opts.provider === "newrelic" feels pretty tacky, perhaps there could be some opts["extra-ld-flags"] or some such that this could move to newrelic yaml?

provider-ci/providers/newrelic/repo/.goreleaser.yml
aq17: I think it's generating it on two lines 😞

provider-ci/src/goreleaser.ts
```diff
@@ -83,8 +83,13 @@ interface GoReleaserOpts {
   "major-version": number;
   providerVersion: string;
   skipTfGen: boolean;
+  "extra-ld-flags": boolean;
```
AaronFriel, line 86: If this is were a `string[]`, could we just put the extra LD flags in `config.yaml`?

aq17: We could and should

AaronFriel (approved):

provider-ci/src/goreleaser.ts
```diff
@@ -113,6 +114,15 @@ export class PulumiGoreleaserPreConfig implements GoreleaserConfig {
       ];
     }
 
+    if (opts["extra-ld-flags"]) {
+        var flags = opts["extra-ld-flags"]
```
t0yv0, line 118: This breaks compilation for me

```
/Users/t0yv0/code/ci-mgmt/provider-ci/src/goreleaser.ts
  118:9   error  Unexpected var, use let or const instead  no-var
  120:18  error  Unexpected var, use let or const instead  no-var
```

Wonder if wrong TypeScript version I have or?

t0yv0 (approved):

PR #384, opened 2023-04-06
Add workflow to automatically trigger `upgrade-provider` run

aq17:
Add workflow to automatically trigger `upgrade-provider` tool to execute `./upgrade-provider pulumi-{PKG}...` whenever pulumi-bot opens an upstream upgrade PR


provider-ci/src/goreleaser.ts
```diff
@@ -115,12 +115,12 @@ export class PulumiGoreleaserPreConfig implements GoreleaserConfig {
     }
 
     if (opts["extra-ld-flags"]) {
-        const flags = opts["extra-ld-flags"]
-        if (flags) {
-            for (var f of flags) {
-                ldflags.push(f)
-            }
+      const flags = opts["extra-ld-flags"];
```
aq17, line 118: Unrelated, fixing lint errors

provider-ci/src/workflows.ts
```diff
@@ -1060,3 +1060,28 @@ export function ModerationWorkflow(
...
+        types: ["opened"],
+      },
+    },
+    env: {
+      GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}",
+      // If there are missing or extra mappings, they can not have been
+      // introduced by updating the bridge, so for this workflow we'll
+      // ignore mapping errors.
+      PULUMI_EXTRA_MAPPING_ERROR: false,
+      PULUMI_MISSING_MAPPING_ERROR: false,
```
iwahbe, line 1079: This seems dubious to me. It seems like our providers should be free of mapping errors. Then either an upgrade won't need these flags or a mapping error would indicate a bug in the bridge.

Will this job only run during a bridge update? I thought it would also run during provider updates, in which case we definitely want mapping errors.

aq17: This job is supposed to run on provider updates. I admittedly copied the workflow from another one to start, these should be removed

iwahbe (commented):
This looks good to me, but I really don't have a deep understanding of GH actions. Is there an example of trial version of this script working?

aq17:
@iwahbe I can add this workflow to a single provider and test it before rolling out to all providers

aq17:
@iwahbe this should be good now. The same workflow in `pulumi-newrelic` (https://github.com/pulumi/pulumi-newrelic/actions/runs/4640148249/jobs/8211770787) succeeded in dispatching the upgrade provider job in `upgrade-provider` (https://github.com/pulumi/upgrade-provider/actions/runs/4640149034/jobs/8211799799 - still fixing that)

iwahbe (approved):
Seeing a successful deploy, lets try it in prod.

PR #386, opened 2023-04-19
Allow more build parallelism

t0yv0:
The original `-p 1` settings were introduced to conserve RAM in a CI setting. As a developer working locally they make my cores sit idle and builds take longer than they should. This setting lets me override `-p 1` with `-p 6` by setting an env var.

Example: https://github.com/pulumi/pulumi-venafi/pull/107/files

### DIFF
--- a/provider-ci/providers/aiven/repo/Makefile
+++ b/provider-ci/providers/aiven/repo/Makefile
@@ -12,6 +12,7 @@ JAVA_GEN := pulumi-java-gen
 JAVA_GEN_VERSION := v0.5.4
 TESTPARALLELISM := 10
 WORKING_DIR := $(shell pwd)
+PULUMI_PROVIDER_BUILD_PARALLELISM ?= 1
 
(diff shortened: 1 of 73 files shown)


AaronFriel (approved):

provider-ci/providers/aiven/repo/Makefile
```diff
@@ -12,6 +12,7 @@ JAVA_GEN := pulumi-java-gen
 JAVA_GEN_VERSION := v0.5.4
 TESTPARALLELISM := 10
 WORKING_DIR := $(shell pwd)
+PULUMI_PROVIDER_BUILD_PARALLELISM ?=
```
iwahbe, line 15: Can we just omit this if the value is "". AFAIK this is equivalent to "If the value is unset, set it to the empty string" but interpolating an unset value defaults to the empty string.

iwahbe (approved):
1 nit I previously mentioned. Otherwise LGTM

PR #391, opened 2023-04-24
Fix upgrade provider workflow trigger

aq17:
Trigger should be opened issues, not pull requests


iwahbe (approved):
I thought we had a working prototype for this. Why did we think the prototype worked but this appears not to?

aq17:
The prototype used `on: push` as a trigger, to easily rerun the workflow on changes to the test branch's workflow file. This works, but isn't the behavior we want – we should trigger the action on open issues only.

PR #396, opened 2023-04-26
Use Provider Publisher composite action for publish_sdks job

guineveresaenger:
This PR removes the multi-step implementation for package publishing and replaces it with a single composite GitHub Action, which we [maintain as pulumi-package-publisher](https://github.com/pulumi/pulumi-package-publisher).

A successful run of the composite Action can be [found against master on pulumi-kafka](https://github.com/pulumi/pulumi-kafka/actions/runs/4800273904/jobs/8541833586).

My main question is - rather than pin to v0.0.5, there may be benefit to using latest, as we will be using the package publisher action to use the Pulumi CLI tool instead.

Note also we are not touching the Goreleaser step.
 
Fixes https://github.com/pulumi/home/issues/2773.

- First pass at action gen
- Refactor: Use Provider Publish Action for publishing SDKs
- Remove the individual checkout steps
- generate provider workflows

### DIFF
--- a/provider-ci/providers/aiven/repo/.github/workflows/main.yml
+++ b/provider-ci/providers/aiven/repo/.github/workflows/main.yml
@@ -280,131 +280,13 @@ jobs:
         author_name: Failure in publishing binaries
         fields: repo,commit,author,action
         status: ${{ job.status }}
-  publish_java_sdk:
-    continue-on-error: true
-    name: publish_java_sdk
-    needs: publish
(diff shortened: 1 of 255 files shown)


iwahbe:
I would definitely advocate pinning to a version. We are highly tolerant to churn in ci-mgmt, but I don't want anything to change in our provider repos without a (repo local) PR that causes the change.

aq17 (approved):
🎉

PR #403, opened 2023-05-02
Create `.upgrade-config.yml` for each provider repo

aq17:
Once https://github.com/pulumi/upgrade-provider/pull/66 is merged (specifically [this commit](https://github.com/pulumi/upgrade-provider/pull/66/commits/ee97107c5b53f347b8fee34d57ad094abd2f9b0b)), we initialize a config file for each provider, with only `upstream-provider-name` set for now


iwahbe:
Could we do this by defining the config file consumed in https://github.com/pulumi/upgrade-provider/pull/61 for each provider? That way, it remains easy to run the `upgrade-provider` tool locally.

aq17:
> Could we do this by defining the config file consumed in [pulumi/upgrade-provider#61](https://github.com/pulumi/upgrade-provider/pull/61) for each provider? That way, it remains easy to run the `upgrade-provider` tool locally.

Done 👍🏼

provider-ci/src/provider.ts
```diff
@@ -127,3 +132,14 @@ export function generateProviderFiles(config: BridgedConfig) {
...
+export function getUpstreamProviderName(providerName: string) {
+	let nameMap = new Map<string, string>([
+		["f5bigip", "bigip"],
+		["confluentcloud", "confluent"]
+	])
+	if (nameMap.has(providerName)) {
+		providerName = nameMap.get(providerName)!
+	}
+	return "terraform-provider-" + providerName
+}
```
iwahbe, line 145: This information already exists in provider config.

For example:
https://github.com/pulumi/ci-mgmt/blob/d85f5e6d456856652bc708e69ffc48bba5e7ce72/provider-ci/providers/confluentcloud/config.yaml#L8

iwahbe (approved):
Lets make sure we don't merge until our GH action is using a version of `upgrade-provider` that can consume these files.

I believe that this will also require bumping the version of our new CI action that these providers depend on.

t0yv0 (approved):
Looks clean.

PR #405, opened 2023-05-03
Use pulumi/schema-tools

### LINKED ISSUE #404: Switch from `mikhailshilkov/schema-tools` to `pulumi/schema-tools`
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Given https://github.com/pulumi/ci-mgmt/pull/274#issuecomment-1455849744, I suggest we make a hard switch-over.

iwahbe:
Fixes #404

https://github.com/pulumi/pulumi-spotinst/pull/311 servers as an example that the flags I'm passing work.

### DIFF
--- a/provider-ci/src/steps.ts
+++ b/provider-ci/src/steps.ts
@@ -258,7 +258,7 @@ export function InstallSchemaChecker(): Step {
     name: "Install Schema Tools",
     uses: action.installGhRelease,
     with: {
-      repo: "mikhailshilkov/schema-tools",
+      repo: "pulumi/schema-tools",
     },
   };
 }
@@ -555,7 +555,7 @@ export function CheckSchemaChanges(): Step {
     name: "Check Schema is Valid",
     run:
       "echo 'SCHEMA_CHANGES<<EOF' >> $GITHUB_ENV\n" +
-      "schema-tools compare ${{ env.PROVIDER }} master --local-path=provider/cmd/pulumi-resource-${{ env.PROVIDER }}/schema.json >> $GITHUB_ENV\n" +
+      "schema-tools compare -p ${{ env.PROVIDER }} -o ${{ github.event.repository.default_branch }} -n --local-path=provider/cmd/pulumi-resource-${{ env.PROVIDER }}/schema.json >> $GITHUB_ENV\n" +
       "echo 'EOF' >> $GITHUB_ENV",
   };
 }
--- a/provider-ci/providers/aiven/repo/.github/workflows/main.yml
+++ b/provider-ci/providers/aiven/repo/.github/workflows/main.yml
@@ -158,7 +158,7 @@ jobs:
       name: Install Schema Tools
       uses: jaxxstorm/action-install-gh-release@v1.5.0
       with:
-        repo: mikhailshilkov/schema-tools
+        repo: pulumi/schema-tools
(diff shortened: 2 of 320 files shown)


provider-ci/src/steps.ts
```diff
@@ -555,7 +555,7 @@ export function CheckSchemaChanges(): Step {
     name: "Check Schema is Valid",
     run:
       "echo 'SCHEMA_CHANGES<<EOF' >> $GITHUB_ENV\n" +
-      "schema-tools compare ${{ env.PROVIDER }} master --local-path=provider/cmd/pulumi-resource-${{ env.PROVIDER }}/schema.json >> $GITHUB_ENV\n" +
+      "schema-tools compare -p ${{ env.PROVIDER }} -o master -n --local-path=provider/cmd/pulumi-resource-${{ env.PROVIDER }}/schema.json >> $GITHUB_ENV\n" +
```
guineveresaenger, line 558: I have positively no idea how this ever worked on a repo with a `main` default branch (and [it does!?](https://github.com/pulumi/pulumi-artifactory/pull/308), but we should use `github.event.repository.default_branch` I think.

guineveresaenger (approved):
One tweak that may warrant another example PR just to make sure the [github action context](https://docs.github.com/en/actions/learn-github-actions/contexts) I suggested is in fact the correct one. 

It looks like Mikhail's version automagically works on `main`, but I don't know why, and I'm not sure this is true for pulumi/schema-tools.

provider-ci/src/steps.ts
iwahbe: Two comments:
1. It looks like GitHub auto-magically replaces `master` with `main` for GET requests. Its very nice that they do this.
2. We are going to replace `master` with your suggestion, because a provider could have a `main` and a `master` branch. I have no idea what GH will do then, and I don't want to find out.

iwahbe:
> One tweak that may warrant another example PR just to make sure the [github action context](https://docs.github.com/en/actions/learn-github-actions/contexts) I suggested is in fact the correct one.
> 
> It looks like Mikhail's version automagically works on `main`, but I don't know why, and I'm not sure this is true for pulumi/schema-tools.

Second example PR is https://github.com/pulumi/pulumi-spotinst/pull/312. Your suggestion worked perfectly. 🙇

thomas11 (approved):
Coincidentally, related to the `default_branch` change, we're just facing the issue in azure-native that we want to have a long-lived feature branch, but schema-tools is useless and noisy on PRs into the branch because it always compares to master. But that's not for this PR, just wanted to make you aware of this use case.

guineveresaenger:
LGTM! Thank you for including the extra confidence check.

@thomas11 - you probably want to use `github.ref` or github.base_ref` in azure-native actions, since schema-tools takes any branch name. Contexts are vrey confusing, but [this is an okay starting point](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context).

Full disclosure: I have _definitely_ [dumped the full github context](https://stackoverflow.com/questions/70104600/complete-list-of-github-actions-contexts/70107953#70107953) so I could see what was available to me in a given action, since the documentation is so confusing (different information is available based on the context).

PR #411, opened 2023-05-08
Add notification step to provider upgrade

aq17:
- Add notification step to send slack message on provider upgrade success or failure
- Also fixes if conditional to run upgrade action on open issue
- Also removes inputs to upgrade provider action that can be determined via github contexts

### DIFF
--- a/provider-ci/src/workflows.ts
+++ b/provider-ci/src/workflows.ts
@@ -1038,7 +1038,7 @@ export function UpgradeProvider(opts: BridgedConfig): GithubWorkflow {
 			steps.NotifySlackUpgradeFailure()
 		)
         .addConditional(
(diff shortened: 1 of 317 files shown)


provider-ci/src/workflows.ts
```diff
@@ -1029,13 +1029,16 @@ export function UpgradeProvider(opts: BridgedConfig): GithubWorkflow {
     jobs: {
       upgrade_provider: new EmptyJob("upgrade-provider")
         .addStep(
-          steps.UpgradeProviderAction(
-            providerName,
-            opts["provider-default-branch"]
-          )
```
guineveresaenger, line 1035: can we update the PR description to say why we don't need these here anymore?

guineveresaenger (approved):
nice catch on the expansion!

PR #412, opened 2023-05-08
Migrate `azure` to `team: ecosystem`

iwahbe:


t0yv0 (approved):
Thank you! Let's do it! AFAIK this might expose it to ci-mgmt "upgrade-ecosystem-providers" scripts but that may be intentional.

kpitzen (approved):
LGTM, similar to the last one.

provider-ci/providers/azure/config.yaml
```diff
@@ -21,5 +21,6 @@ plugins:
     version: "4.8.2"
   - name: azuread
     version: "5.33.0"
-team: providers
+team: ecosystem
 goBuildParallelism: 2
+javaGenVersion: "0.8.0"
```
guineveresaenger, line 26: This is a change from the default - what is the reason here?

provider-ci/providers/azure/config.yaml
guineveresaenger, line 1: I think it's worth double-checking the `upstream-provider-org` setting, since we are removing the fork in favor of a submodule in pulumi/pulumi-azure/pull/1263. As far as I can tell, it is only used by [this workflow](https://github.com/pulumi/ci-mgmt/blob/master/provider-ci/src/workflows.ts#L491) which is, I believe, in need of retirement? Regardless, there's no fork, pulumi isn't the upstream org.

guineveresaenger (approved):
I'd like to see a follow-up issue for cleanup of the `upstream-provider-org` and the `update_upstream_provider` Workflow job.

provider-ci/providers/azure/config.yaml
guineveresaenger: Oh, I see; this was present in the provider already. 👍

provider-ci/providers/azure/config.yaml
iwahbe: I think you're right, but I'm going to merge that as a separate PR. Right now `upstream-provider-org` is a required field on all providers (even ones without forks or patches). 

I'll open a PR to remove the old action later today.

PR #414, opened 2023-05-09
Fix upgrade-provider.yml slack webhook

aq17:
`SLACK_WEBHOOK_URL` is a github secret in our org, it should not be an env var


guineveresaenger (approved):
d'oh!

Could you run a sample workflow that hits the webhook so we know it works for real this time?

PR #415, opened 2023-05-10
Remove the update-upstream-provider workflow

iwahbe:
This was our first attempt at workflow driven bridged provider updates. This is the workflow that generates PRs with:

     Failed to automatically update upstream provider (probably beacuse of
     new resources or data sources, which must be mapped manually).

We now rely on different actions for automatic provider updates.

Responding to https://github.com/pulumi/ci-mgmt/pull/412#discussion_r1189194649.


aq17 (approved):

guineveresaenger (approved):
thank you for the cleanup! LGTM

PR #419, opened 2023-05-11
Update `javaGenVersion` to latest

iwahbe:
Update to the latest version of java generation across the fleet of providers.

- Because we don't control GCP, I have pinned the version used to team provider's doesn't see any change.


t0yv0:
I think this will break everything. Can you run a test run on one provider? Newer versions of java gen will change how Java code is generated so these PRs would likely fail. We need something to re-generate the Java sdk/java files. Should it be update-provider? We might need to figure out a rollout strategy where we first take the javagen version out of ci-mgmt and into some repo-local config file, and then let update-provider move that, while also re-generating sdk/java.

PR #428, opened 2023-05-15
Fix `upgrade-provider.yml` if condition

aq17:
- Conditional was invalid, causing all open issues to trigger the provider upgrade
- Also bumps the version of `pulumi-upgrade-provider-action` to `v0.0.5`, which fetches all tags after checking out the repo in order to accurately determine the latest version


guineveresaenger (approved):
This should fix the auto-triggering of update runs! 🤞

PR #432, opened 2023-05-17
Add an issue template for ecosystem providers

t0yv0:
This should simplify creating checklists that require doing something for each of our providers.


t0yv0:
I'll merge now but feel free to comment I can fix it up later.

PR #436, opened 2023-05-17
Change schema check to update comments in place

lblackstone:
Previously, the schema-check bot would add a new comment with the results of the check for every new commit. This change upserts the comment instead, which cuts down on noise in PRs.

Docs link: https://github.com/thollander/actions-comment-pull-request#update-a-comment


rquitales (approved):

danielrbradley (approved):
We already use this in azure-native so can confirm this works 👍

PR #442, opened 2023-05-26
Add workflow to upgrade tfbridge version on dispatch

aq17:
This PR does a few things:

- Adds a function that adds a workflow `upgrade-bridge.yml` in all ecosystem-providers, to upgrade the tfbridge version in a provider that can be triggered via a workflow dispatch. 
- Adds a workflow in `ci-mgmt` that triggers a dispatch to all providers to run the `upgrade-bridge.yml` workflow. This allows us to quickly upgrade a single provider if necessary (such as to resolve a P1), or to trigger upgrades for all providers via a single deployment in `ci-mgmt`. 

We can also consider setting up a cron job in `ci-mgmt` to regularly attempt to perform bridge upgrades, instead of waiting until a full provider upgrade is needed, in order to keep PR's more streamlined between tfbridge vs upstream version updates.

Fixes https://github.com/pulumi/home/issues/2589#issuecomment-1564895535

To do: Remove the old `upgrade-bridge.yml` workflow, which is broken/ not maintained.


t0yv0:
I think this looks good! Couple of comments.

There is one existing workflow like this one: https://github.com/pulumi/pulumi-akamai/blob/master/.github/workflows/update-bridge.yml that we might want to deprecate/remove. I don't think it is working right. I tried recently on pulumi-gcp and although that is not one of our providers it didn't work. Last runs I've found are 1 year old.

My other comment is that this by itself does not fix pulumi/home#2589 although it gets pretty close. 

I think in addition to this we need some workflow that triggers it in response to a release or polls for it as you suggest. In terms of latency, I'd be ok with overnight polling for new bridge releases, that'd be a big improvement over status quo.

t0yv0 (approved):
The change LGTM, let's try it and iterate.

aq17:
> I think this looks good! Couple of comments.
> 
> There is one existing workflow like this one: https://github.com/pulumi/pulumi-akamai/blob/master/.github/workflows/update-bridge.yml that we might want to deprecate/remove. I don't think it is working right. I tried recently on pulumi-gcp and although that is not one of our providers it didn't work. Last runs I've found are 1 year old.
> 
> My other comment is that this by itself does not fix [pulumi/home#2589](https://github.com/pulumi/home/issues/2589) although it gets pretty close.
> 
> I think in addition to this we need some workflow that triggers it in response to a release or polls for it as you suggest. In terms of latency, I'd be ok with overnight polling for new bridge releases, that'd be a big improvement over status quo.

Yes, I'm working on creating one workflow in `ci-mgmt` that can be manually triggered to trigger this workflow in all providers (so that we can deploy releases quickly with one click). We can add the nightly cron job after

provider-ci/src/workflows.ts
```diff
@@ -241,99 +241,6 @@ export function PullRequestWorkflow(
   };
 }
 
-interface UpdatePulumiTerraformBridgeWorkflowArgs {
-  providerDefaultBranch: string;
-}
-
-export function UpdatePulumiTerraformBridgeWorkflow(
```
t0yv0, line 248: Curious it's not used in "aws", "Azuread", etc providers that are off ecosystem team? might need to retain for bw-compat for some time.

aq17: It looks like they haven't been used in a while but good point. We can leave them for now to be safe

aq17:
Going to merge this and test the dispatch.

PR #443, opened 2023-06-05
Fix `upgrade-bridge-ecosystem-providers.yml`

aq17:
The github action we use in `ci-mgmt` to send a dispatch to each providers `upgrade-bridge` workflow, https://github.com/marketplace/actions/workflow-dispatch, requires `ref` to be set if it differs from the calling repo branch (which is `master` for this repo, `ci-mgmt`.

This changes it to first attempt to use `master`, and then continue if the ref is not found and to use `main` instead. We should still consider moving all provider default branches from `master` -> `main` eventually.


t0yv0 (approved):

t0yv0:
Better invert this, use main first fall back to master.

PR #446, opened 2023-06-06
Rebuild templating using Go text templates 

AaronFriel:
Rebuild templating using Go text templates

Seeking to resolve the CI disparity between ecosystem managed repositories and pulumi/pulumi-std#29, it quickly became apparent that the style of GitHub Actions - using reusable workflows - in that repository and in pulumi-command would require a major rewrite of the templating in this repository.

Making large changes to the templating in this repository is more difficult as it requires modifying the Node scripts and the type-safe "generator" to reverse engineer CI workflows into a known form.

Directly generating text templates, we skip the intermediate steps and can more naturally write the intended GitHub Actions YAML.

The Go utility in this repository can be run and a diff with the current repository scripts generated via:

```
cd ./package-ci
./scripts/test.sh
```

The committed template, "bridged-provider", is intended to be bit-for-bit identical in output to the current scripts to the extent possible. As a result, there is no semantic diff with the current output.

The two cases where the output was not identical are in the New Relic provider, and the diff is inconsequential wrapping behavior in the current Node scripts that is not present in the Go text template, both appearing in `.goreleaser` files. In this case, the wrapping has no effect on Goreleaser:

```diff
diff --git a/../provider-ci/providers/newrelic/repo/.goreleaser.prerelease.yml b/./test-output/providers/newrelic/repo/.goreleaser.prerelease.yml
index 439adbf7a..a5b009761 100644
--- a/../provider-ci/providers/newrelic/repo/.goreleaser.prerelease.yml
+++ b/./test-output/providers/newrelic/repo/.goreleaser.prerelease.yml
@@ -29,8 +29,7 @@ builds:
   ignore: []
   ldflags:
   - -X github.com/pulumi/pulumi-newrelic/provider/v5/pkg/version.Version={{.Tag}}
-  - -X
-    github.com/newrelic/terraform-provider-newrelic/v2/main.UserAgentServiceName=pulumi
+  - -X github.com/newrelic/terraform-provider-newrelic/v2/main.UserAgentServiceName=pulumi
   main: ./cmd/pulumi-resource-newrelic/
```


package-ci/main.go
```diff
@@ -0,0 +1,7 @@
+package main
+
+import "github.com/pulumi/ci-mgmt/package-ci/cmd"
```
iwahbe, line 3: Can we default `cmd` and `pkg` to `internal/cmd` and `internal/pkg`?

package-ci/pkg/templates/bridged-provider/.goreleaser.yml
```diff
@@ -0,0 +1,63 @@
...
+    goos: windows
+  #{{- else }}#
+  ignore: []
+  #{{- end }}#
+  ldflags:
+  #{{- if ge (index .Config "major-version") 2 }}#
+  - -X#{{if (gt (len .Repository) 22)}}##{{ "\n   " }}##{{end}}# github.com/#{{ .Repository }}#/provider/v#{{ index .Config "major-version" }}#/pkg/version.Version={{.Tag}}
+  #{{- else }}#
+  - -X#{{if (gt (len .Repository) 24)}}##{{ "\n   " }}##{{end}}# github.com/#{{ .Repository }}#/provider/pkg/version.Version={{.Tag}}
+  #{{- end }}#
```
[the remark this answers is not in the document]
AaronFriel: Once we merge this and replace the existing provider workflow, I'd take a pass through this simplifying the messy logic that maintained bitwise identical output.

package-ci/main.go
AaronFriel: I can imagine us wanting to move this or a version of it to the Pulumi CLI, any reason to put this behind internal?

iwahbe: I don't think we are expecting to keep a backwards-compatible interface on the functions we expose. Moving to `internal` enforces that decision until we explicitly decide to expose something.

package-ci/pkg/templates/bridged-provider/.github/workflows/update-bridge.yml
guineveresaenger, line 1: I don't believe we use this workflow anymore, so we can likely leave it off!

aq17: The new workflow is `upgrade-bridge.yml`

provider-ci/providers/docker/config.yaml
```diff
@@ -2,6 +2,7 @@ provider: docker
 major-version: 4
 aws: true
 gcp: true
+gcpRegistry: true
```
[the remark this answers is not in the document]
AaronFriel: It's required for the new template.

package-ci/scripts/test.sh
```diff
@@ -0,0 +1,56 @@
...
+
+set -eo pipefail
+
+rm -rf ./test-output/*
+
+go build -o ./test-output/package-ci
+
+PROVIDER="${1:-""}"
+
+generate() {
```
t0yv0, line 11: Have `Makefile` run this instead of the old system?

t0yv0:
I'm in favor, the devil though is in ensuring uneventful rollout. If the new system still materializes all the workflow files and there's a zero diff (it's byte-for-byte the same) then it'd be excellent we can just swap the generator out without projects noticing.

package-ci/pkg/templates/bridged-provider/.github/workflows/update-bridge.yml
AaronFriel: Can we keep it for now and remove it after merging? Would like to keep the diff of applying the change as small as possible.

package-ci/scripts/test.sh
AaronFriel: Agreed

guineveresaenger (approved):
Since @kpitzen and team cleared this change, let's do it.

Do we want to include instructions for running the `/scripts/test.sh` as part of development? Or should we remove it, as it was part of the PoC?

PR #449, opened 2023-06-07
Upgrade `pulumi-oci` major version to v2

aq17:
Part of https://github.com/pulumi/pulumi-oci/pull/149

### DIFF
--- a/provider-ci/providers/oci/config.yaml
+++ b/provider-ci/providers/oci/config.yaml
@@ -1,5 +1,5 @@
 provider: oci
-major-version: 1
+major-version: 2
 lint: false
 parallel: 1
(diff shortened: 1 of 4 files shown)


t0yv0:
Curious why are CI checks not running at all? Something is fishy here. Also this doesn't look like a full maj-version upgrade, you mention it's part of a multi-step PR sequence can you shed some light on the steps? Thanks.

AaronFriel:
@t0yv0 this is a PR on ci-mgmt, I don't think we have checks in place right now. TBD.

AaronFriel (approved):

iwahbe (approved):

PR #450, opened 2023-06-08
Replace provider-ci with Go text templating

AaronFriel:
Replace provider-ci with package-ci

This commit deletes the provider CI folder and replaces it with package-ci, applying #446, updating the Makefile to conform to the same interface.

Two shared TypeScript files are moved to native-provider-ci:
* provider-ci/src/shared-workflows.ts
* provider-ci/src/action-versions.ts

This is a stacked PR - the last commit is the only commit added in this PR.


aq17 (approved):
LGTM

guineveresaenger (approved):

PR #452, opened 2023-06-08
Upgrade `JAVA_GEN_VERSION` to `v0.9.4`

aq17:
Some providers are failing `build_sdk (java)` tests due to an outdated version being used (i.e. https://github.com/pulumi/pulumi-datadog/actions/runs/5205167636/jobs/9390402783?pr=296) This upgrades to the latest `v0.9.4`


t0yv0 (changes requested):
This is going to break everything as the generated code changes when upgrading Java. This needs to be coupled to something like update-provider to re-generate sdk/java code.

t0yv0:
Same as https://github.com/pulumi/ci-mgmt/pull/419

aq17:
@t0yv0 ah, that makes sense – I'll close this then for now.

PR #480, opened 2023-07-13
Remove the `lint_sdk` step from managed providers

iwahbe:
This removes the `lint_sdk` step from managed providers. This step ran golangci-lint on the go SDK. We do not perform linter for our other language SDKs.

Since this only affects providers whose CI is managed by this repo, all provider SDKs are fully generated. We should not lint generated SDKs.


thomas11 (approved):
LGTM, however - when _do_ we lint generated SDKs to detect issues to feed back into codegen improvements?

iwahbe:
> LGTM, however - when _do_ we lint generated SDKs to detect issues to feed back into codegen improvements?

I don't think we ever **lint** generated SDKs, we test that they **build**, and we do so for *all* of our languages, not just go. As far as we are concerned, if the generated code builds, is backwards compatible, and passes tests it's good.

PR #486, opened 2023-07-30
Add tfconversion providers to CI Management

guineveresaenger:
Part of https://github.com/pulumi/home/issues/3025

This pull request adds five new providers to CI management:

- [null](https://github.com/pulumi/pulumi-null)
- [archive](https://github.com/pulumi/pulumi-archive)
- [local](https://github.com/pulumi/pulumi-local)
- [http](https://github.com/pulumi/pulumi-http)
- [external](https://github.com/pulumi/pulumi-external)

We and others use these providers mainly for terraform conversion work, and as such, we don't want to advertise them in the registry. Their functionality in most Pulumi programs is covered by either Pulumi features, or the language libraries.

- add scaffolding for new providers' CI management
- Skip docs build for archive, external, http, null providers
- Add relevant tokens to config


iwahbe (approved):
Changes look good. Can you explain why we are not publishing these providers in the registry? It is just because we only want them used as part of `pulumi convert --from terraform`?

guineveresaenger:
@iwahbe - I updated the PR description to include this info!

PR #496, opened 2023-08-02
Generate upstream scripts and update configs for tfconvert providers

guineveresaenger:
We missed a few items the first time around. 
This PR adds upstream scripts and upgrade configs for the following providers:

- local
- http
- null
- external
- archive


mjeffryes:
n00b question: I'm not super familiar with how the generation works in this repo. Is this just the output of running `make` again, or did some of the changes trigger the rest to be generated?

t0yv0:
Looks like these changes were made by doing:

```
cd provider-ci && make
```

Oof we might want to have a PR check that doing this doesn't result in dirty state.

t0yv0 (approved):

guineveresaenger:
@mjeffryes - when I ran `make` for the OG changes, these files did not get generated. I ran `make` for a different provider, and suddenly all of these repos' extra files popped up; hence this PR. :(

PR #499, opened 2023-08-02
Update go version to update to from `1.19` to `1.20`

aq17:
We should be targeting `1.20` now, so workflow updates such as https://github.com/pulumi/pulumi-aws-native/pull/1014 don't attempt to downgrade the go version

### DIFF
--- a/tools/sourcemigrator/index.ts
+++ b/tools/sourcemigrator/index.ts
@@ -58,11 +58,11 @@ function updatePulumiCoreRefTo3x(): SourceMigration {
     return sm;
 }
 
-function updateGo_1_19(): SourceMigration {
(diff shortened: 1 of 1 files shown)


mjeffryes:
Are all the managed repos on 1.20 now? or is this going to update a bunch of them next time we apply ci-mgmt? I ask b/c just updating the go.mod doesn't seem sufficient for a version update on it's own; we'd need to run `go mod tidy` to update the sums right?

aq17:
> 

I believe they all migrated over to `1.20` already (a search for `1.19` in all `ci-mgmt` `go.mod` file yields 0 results, so this change is only to prevent unwanted downgrades.

PR #504, opened 2023-08-04
Updating `pulumi-java-gen` to current latest v0.9.5

ringods:
I'm trying to get `ci-mgmt` working again for 3rd party providers. This PR is a first fix to get closer to that target.

Updating the version of `pulumi-java-gen` to current latest v0.9.5. The main fix needed for 3rd party providers is the generation of missing fields in the gradle build file added in [v0.9.4](https://github.com/pulumi/pulumi-java/releases/tag/v0.9.4)

PR #505, opened 2023-08-04
Fixes for null provider

ringods:
Submitting the work of #504 without the version update.

Updates:
* List more of the need build tools
* Removes a duplicate build dependency
* Fixes the build+lint of the `null` provider files


iwahbe (approved):

t0yv0 (approved):
Thanks a ton!

PR #508, opened 2023-08-08
Remove cleanCacheBetweenBuilds option

iwahbe:
This is causing publish failures in pulumi-azure (they only place where this is enabled) because it duplicates hooks. We now do this by default, so there is no observable change in working CI.


guineveresaenger (approved):
LGTM!

PR #518, opened 2023-08-22
Generate workflows for aws

iwahbe:
This PR generated the workflow files for https://github.com/pulumi/pulumi-aws/pull/2708. 

This does not account for aws's Makefile, which is still non-standard. To make this safe to merge, we skip generating a Makefile for pulumi-aws (leaving the current one in place).

Improvements to ci-mgmt:
- Parameterize by test runners (default and publisher).

---

~We are waiting to merge until aws v6 has launched.~ v6 has launched and we have updated pulumi-aws's workflows. Merging now.


guineveresaenger (commented):
This looks great at first glance.

Can you add an explanation why we need the `extraTests` config option? I'm assuming it's because we want to test the provider shim in AWS but nowhere else - however I don't see an extra shim test action being generated; what am I missing?

provider-ci/providers/aws/repo/.github/workflows/run-acceptance-tests.yml
```diff
@@ -0,0 +1,400 @@
...
+              submodules: true
+          - uses: actions/setup-go@v4
+            with:
+              cache-dependency-path: provider/shim/go.sum
+              go-version-file: provider/shim/go.mod
+          - name: go test
+            run: |
+              cd provider/shim
+              go test -v .
+      timeout-minutes: 60
```
iwahbe, line 388: > This looks great at first glance.
> 
> Can you add an explanation why we need the `extraTests` config option? I'm assuming it's because we want to test the provider shim in AWS but nowhere else - however I don't see an extra shim test action being generated; what am I missing?

You're right, we use `extraTests` for testing the shim.

t0yv0 (approved):
🚢

PR #519, opened 2023-08-22
Disable Goreleaser posthooks for GCP

rquitales:
Fixes: https://github.com/pulumi/pulumi-gcp/issues/1161


iwahbe:
Since this change won't persist after the next `cd provider-ci && make providers`, what is the benefit of adding it here (instead of in pulumi-gcp)?

rquitales:
Ah, overlooked that this is generated files being edited.

iwahbe:
You can use a strategy similar to https://github.com/pulumi/ci-mgmt/pull/521 to allow only gcp to not have hooks.

PR #525, opened 2023-08-25
Fix upstream.sh

iwahbe:
This allows running `make upstream.rebase` without FROM or TO.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/scripts/upstream.sh
+++ b/provider-ci/internal/pkg/templates/bridged-provider/scripts/upstream.sh
@@ -87,7 +87,9 @@ start_rebase() {
   fi
 
   git branch -f local
-  git checkout -B pulumi-patch "${FROM}"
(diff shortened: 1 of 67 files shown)


provider-ci/internal/pkg/templates/bridged-provider/scripts/upstream.sh
```diff
@@ -87,7 +87,9 @@ start_rebase() {
   fi
 
   git branch -f local
-  git checkout -B pulumi-patch "${FROM}"
+  # ${FROM} is intentionally unquoted, so that not supplying a ${FROM} argument works.
+  # This way we don't pass an empty string to `git checkout` (which breaks).
+  git checkout -B pulumi-patch ${FROM}
```
mjeffryes, line 92: I'd prefer we just test if the FROM variable is empty; there isn't any good reason for someone to use a value with a space in it, but it still just makes me uneasy.

```suggestion
  if [[ -n "$FROM" ]]; then
     git checkout -B pulumi-patch "$FROM"
  else
     git checkout -B pulumi-patch
  fi
```

PR #528, opened 2023-08-25
Allow checking out submodules as part of repo checkout

iwahbe:
This is needed by https://github.com/pulumi/pulumi-aws/pull/2708.

For an example of what it looks like when `checkoutSubmodules: true`, see https://github.com/pulumi/ci-mgmt/pull/528/commits/be431ea8c5314c34162aac226ba8b176d9f0375f.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/run-acceptance-tests.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/run-acceptance-tests.yml
@@ -16,6 +16,9 @@ jobs:
       uses: #{{ .Config.actionVersions.checkout }}#
       with:
         ref: ${{ env.PR_COMMIT_SHA }}
+        #{{- if .Config.checkoutSubmodules }}#
+        submodules: #{{ .Config.checkoutSubmodules }}#
+        #{{- end }}#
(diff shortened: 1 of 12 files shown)


provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/run-acceptance-tests.yml
```diff
@@ -180,6 +189,9 @@ jobs:
       uses: #{{ .Config.actionVersions.installGhRelease }}#
       with:
         repo: pulumi/schema-tools
+#{{- if .Config.actions.preBuild }}#
+#{{ .Config.actions.preBuild | toYaml | indent 4 }}#
+#{{- end }}#
```
mjeffryes, line 194: Is this just a spot we missed earlier?

mjeffryes (approved):
LGTM

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/run-acceptance-tests.yml
iwahbe: Oops, I forgot to take that out. It's part of a later addition from the massive PR this is coming from.

PR #531, opened 2023-08-29
Create a daily workflow to update native-providers that are opted into ci-mgmt

mjeffryes:
We recently got aws-native back in sync with ci-mgmt; this will help us keep it in sync.

### DIFF
--- a/scripts/generate_native_providers_list.py
+++ b/scripts/generate_native_providers_list.py
@@ -1,4 +1,17 @@
+import argparse
 import os
 import json
 
-print(json.dumps(sorted(os.listdir('../native-provider-ci/providers'))))
+
+excluded_from_auto_pr =["azure-native", "google-native", "kubernetes", "command"]
+
+if __name__ == '__main__':
+    ap = argparse.ArgumentParser()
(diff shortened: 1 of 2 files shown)


scripts/generate_native_providers_list.py
```diff
@@ -1,4 +1,17 @@
+import argparse
 import os
 import json
 
-print(json.dumps(sorted(os.listdir('../native-provider-ci/providers'))))
+
+excluded_from_auto_pr =["azure-native", "google-native", "kubernetes", "command"]
```
iwahbe, line 6: If we are excluding these providers, what set of providers are opted in? Is this only aws-native?

iwahbe (approved):

scripts/generate_native_providers_list.py
mjeffryes: Yes for the moment; I think all the others have drifted, but I'm hoping to sync them up soon and remove them from the list.

thomas11: azure-native will be tricky, start with the others :)

PR #537, opened 2023-08-30
Upgrade providers to go 1.21

mjeffryes:
I'd like to add more of the native providers to the nightly updates. pulumi-command is almost in sync, but it's already been updated to go 1.21. Due to the interaction with sourcemigrator, it seems like the most expedient way to do this is just to bump all the providers to 1.21 

I'm not sure how much of a risk this represents; I'm definitely open to other approaches. Longer term, it seems like we might want to reevaluate how sourcemigrator works to avoid this.


iwahbe (approved):
Looks good to me.

In the future, it would be great to split this PR into multiple commits, separating hand authored commits from their derivative makefile changes.

PR #541, opened 2023-08-31
Native providers: Add codecov upload step after running provider tests

mjeffryes:
Adds a step to native providers workflows that sends coverage reports to codecov

Tested on aws-native here: https://github.com/pulumi/pulumi-aws-native/pull/1058


iwahbe (approved):
If you tested this change, it would be awesome to link to the PR where that happened?

If not, please test before merging :).

Issue #554, opened 2023-09-08
No space left on device blocks Provider upgrades


t0yv0:
Some recent changes in Go SDK generation pushed the builds over the limit of disk space.

- https://github.com/pulumi/pulumi-aws-native/pull/1083
- https://github.com/pulumi/pulumi-aws/pull/2792
- https://github.com/pulumi/pulumi-azure/pull/1350
- https://github.com/pulumi/pulumi-kubernetes/pull/2556
- https://github.com/pulumi/pulumi-google-native/pull/925

```
build_sdk (go)](https://github.com/pulumi/pulumi-azure/actions/runs/6116569557/job/16603257286)
Unhandled exception. System.IO.IOException: No space left on device : '/home/runner/runners/2.308.0/_diag/Worker_20230908-024953-utc.log'
   at System.IO.RandomAccess.WriteAtOffset(SafeFileHandle handle, ReadOnlySpan`1 buffer, Int64 fileOffset)
   at System.IO.Strategies.BufferedFileStreamStrategy.FlushWrite()
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.Diagnostics.TextWriterTraceListener.Flush()
   at System.Diagnostics.TraceSource.Flush()
   at GitHub.Runner.Common.TraceManager.Dispose(Boolean disposing)
   at GitHub.Runner.Common.TraceManager.Dispose()
   at GitHub.Runner.Common.HostContext.Dispose(Boolean disposing)
   at GitHub.Runner.Common.HostContext.Dispose()
   at GitHub.Runner.Worker.Program.Main(String[] args)
System.IO.IOException: No space left on device : '/home/runner/runners/2.308.0/_diag/Worker_20230908-024953-utc.log'
   at System.IO.RandomAccess.WriteAtOffset(SafeFileHandle handle, ReadOnlySpan`1 buffer, Int64 fileOffset)
   at System.IO.Strategies.BufferedFileStreamStrategy.FlushWrite()
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.Diagnostics.TextWriterTraceListener.Flush()
   at GitHub.Runner.Common.HostTraceListener.WriteHeader(String source, TraceEventType eventType, Int32 id)
   at GitHub.Runner.Common.HostTraceListener.TraceEvent(TraceEventCache eventCache, String source, TraceEventType eventType, Int32 id, String message)
[...]

PR #556, opened 2023-09-08
Parameterize by the runner used when building SDKs

iwahbe:
Fixes part of #554 

Extend the `runner` config option to include `buildSdk`. Follow up commits set `runner.buildSdk` to `pulumi-ubuntu-8core` for `aws` and `azure`.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider.config.yaml
+++ b/provider-ci/internal/pkg/templates/bridged-provider.config.yaml
@@ -57,7 +57,8 @@ extraTests:
 #     ...
 runner:
   default: ubuntu-latest
-  publish: ubuntu-latest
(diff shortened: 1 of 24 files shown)


aq17 (approved):

iwahbe:
I'm waiting on https://github.com/pulumi/pulumi-aws/pull/2792 to go fully green, then I'll merge. I don't expect problems here though.

t0yv0 (approved):
I say ship it, maybe, but one thing we'll notice is that CI times will hit a wall if this one runner fleet has insufficient capacity. Also looks like I have a sister PR here for native providers that are also blocked.

t0yv0:
So it doesn't fix #554 btw because native providers. Tough .

PR #557, opened 2023-09-08
Build sdks on larger runner

t0yv0:
Workaround https://github.com/pulumi/ci-mgmt/issues/554 fallout for native providers. Some of these have drifted from ci-mgmt but I notice that pulumi-aws-native is very close so it could definitely benefit it.


iwahbe (approved):
Given how few native providers we have, I'm fine bumping all of them.

PR #558, opened 2023-09-08
Get compatible with pushing wheels to PyPI

t0yv0:
Noticing that some native providers like k8s are laudably close to being cleanly managed by ci-mgmt; so I need to upstream this change here to reduce drift in their local scripts.


mjeffryes (approved):
Thanks for keeping the drift in check!

PR #569, opened 2023-09-13
Dependencies: package publisher 0.11

t0yv0:
As far as I know this is safe to roll out. 0.11 unlocks the capability to publish python wheels which I especially care about for bridged providers (AWS, GCP especially). 

LMK if this is good to go or we need a pilot run.

@pulumi/providers


t0yv0:
Per @guineveresaenger this is blocked until every provider has received bridge update to 3.59.0 and correspondingly pulumi/pkg update to 3.81.0 or above or it will break them.

t0yv0:
Closing in favor of the approach in https://github.com/pulumi/ci-mgmt/pull/570 where we can go provider-by-provider to do piecemeal updates.

PR #575, opened 2023-09-14
Simpler lint action

t0yv0:


t0yv0:
Here's lint job that worked with this method: 

https://github.com/pulumi/pulumi-random/actions/runs/6190963051/job/16808284964#step:5:51

t0yv0:
I'll start from scratch here to make it simpler.

Issue #574, opened 2023-09-14
Lint permanently failing


t0yv0:
Observing this error:

compile: version "go1.21.0" does not match go tool version "go1.21.1"

In pulumi-random it's persistent and blocking release.

PR #576, opened 2023-09-14
Simplify the lint job

t0yv0:
Fixes https://github.com/pulumi/ci-mgmt/issues/574

I think the root cause is some version mismatch between Go tooling and golint version. I found the job really strange though so took the time to make it more in line with the docs on golangci-lint itself and how we lint the bridge.

Use official golangci/golangci-lint-action@v3 action

Pro: caches
Con: drift from Makefile to CI

Do not build the provider in the lint job.

Pro: faster, fewer things to install
Con: need to disarm go:embed directives; the linter checks if the embedded files are not present, which they will not be before a lot of computation 

Do not containerize the lint job.

Pro: use same ubuntu-latest runtime environment as the rest of the jobs.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
@@ -150,7 +150,6 @@ jobs:
         aws s3 cp "${{ env.COVERAGE_OUTPUT_DIR }}/summary.json" "${s3FullURI}" --acl bucket-owner-full-control
   #{{ if .Config.lint -}}#
   lint:
-    container: golangci/golangci-lint:v1.54.1
     name: lint
     runs-on: #{{ .Config.runner.default }}#
     steps:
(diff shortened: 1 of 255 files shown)


t0yv0:
Test run here: https://github.com/pulumi/pulumi-random/actions/runs/6191227056/job/16809091027#step:5:50

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
```diff
@@ -160,22 +159,15 @@ jobs:
...
-    - name: Mark repo as safe directory
-      run: git config --global --add safe.directory /__w/pulumi-${{ env.PROVIDER }}/pulumi-${{ env.PROVIDER }}
-    - name: Unshallow clone for tags
-      run: git fetch --prune --unshallow --tags
 #{{ .Config.actions.setupGo | toYaml | indent 4 }}#
-    - name: Install pulumictl
-      uses: #{{ .Config.actionVersions.installGhRelease }}#
+    - name: disarm go:embed directives to enable lint
+      run: |
+        git grep -l 'go:embed' -- provider | xargs sed -i 's/go:embed/ goembed/g'
```
guineveresaenger, line 165: oof - well if it works, it works! 😅

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
```diff
@@ -160,22 +159,15 @@ jobs:
       with:
         submodules: #{{ .Config.checkoutSubmodules }}#
 #{{- end }}#
-    - name: Checkout Scripts Repo
```
guineveresaenger, line 163: excellent!!

guineveresaenger (approved):
I am a big fan! We
- lose the container ✅ 
- get to use a third-party-maintained tool ✅ 
- and lose come more ci-scripts dependency ✅ 

Have we tried this in e.g. pulumi-random?

t0yv0:
Yes we tried it here https://github.com/pulumi/pulumi-random/actions/runs/6191227056/job/16809091027#step:5:50

iwahbe (commented):
I don't see where this will apply patches to the forked upstream. I believe that we will need to call `make upstream` before this is safe.

t0yv0:
Hmm, I see because that's influencing provider compilation. Let me test run on a repo with forked upstream.

t0yv0:
https://github.com/pulumi/pulumi-cloudflare/actions/runs/6191728298/job/16810539765 

here's cloudflare succeeding after running `make upstream`.

To address that concern.. at the cost of a little bit more makefile-ci drift. We could rework it to pure makefile just would be ideal not to depend on extra targets; perhaps `make lint_provider.only` or some such. I'm tempted to just go with this for now once it's safe to unblock.

t0yv0:
Second lint in 47s https://github.com/pulumi/pulumi-cloudflare/actions/runs/6191728298/job/16810695463

PR #578, opened 2023-09-15
We now have TLS going back to 4.x

t0yv0:
This reflects changes done to pulumi-tls repo today. These need to be in ci-mgmt so that nightly job does not roll them back.


t0yv0:
Merging to avoid rolling back over the weekend, let me know if there's any comments retroactively.

iwahbe (approved):

PR #602, opened 2023-09-21
Remove pulumi-vault from regular upgrade maintenance

guineveresaenger:
Due to https://github.com/pulumi/pulumi-vault/issues/197, pulumi-vault
cannot currently benefit from automation per this repository. Putting it
in a separate folder to await the resolution of the blocking issue and
avoid noise and spurious p1s caused by attempts to upgrade the bridge.

Upstream upgrades will continue to run automatically as per current
state of pulumi-vault's Workflow files.

Resolution for https://github.com/pulumi/pulumi-vault/issues/310.


iwahbe (changes requested):
`Pulumi-vault` can be maintained by this repos automation, just not for `--kind=sdk,bridge` updates.

It looks like this will prevent us from getting workflow updates which we still want AFAIK.

t0yv0:
That's true but that's what we kind of want also, because workflows will be moving on to wheel based publishing and so forth which requires Pulumi v3.80.0 or higher which Vault is currently unable to upgrade to.

Also note that upgrade-provvider currently not working either:
https://github.com/pulumi/upgrade-provider/issues/162

Please reconsider - the is very little value to special case and do extra work for vault. When we are doing work for vault, I recommend (or volunteer) spending a day upgrading upstream to get on the newer azcore version and testing that out, then we can get back to completely normal state.

t0yv0 (approved):

iwahbe (approved):

PR #604, opened 2023-09-21
Cleanup: Remove near-duplicate `main` and `master` Workflow files according to repo default branch

guineveresaenger:
Leverages `providerDefaultBranch` to generate the correct workflow name and triggers so that each repo has either `master.yml` or `main.yml`.

Updates all providers whose default branch is `main` in their config.

Incidentally fixes a few hard coded `master` references in Workflows on repos whose default branch is `main`.


iwahbe (approved):
LGTM

Unless you went through each provider, I wouldn't be surprised if we are missing a `providerDefaultBranch: main` somewhere. This will help us find those errors.

guineveresaenger:
I was able to know which provider was on `main` based on which upgrade-bridge job was failing how last week  😂

PR #607, opened 2023-09-22
[DO NOT MERGE] add shmecrets

guineveresaenger:
Testing lint job.


guineveresaenger:
Looks like Actions have to live on default before we can try them. Ugh.

Issue #344, opened 2023-02-14
Add CI check `actionlint` to check generated GitHub Actions workflows


AaronFriel:

danielrbradley:
Example implementation in Azure Native:

https://github.com/pulumi/pulumi-azure-native/blob/7b05175f2879eb3fb55426e4c8ba8033ccd0a56c/.github/workflows/acceptance-test.yml#L59-L66
```
  lint_workflows:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout Repo
        uses: actions/checkout@v3

      - name: Lint actions
        uses: reviewdog/action-actionlint@v1.36.0
```

Ideally this would be done before pushing out to each individual provider too though.

guineveresaenger:
Looking into this more - we do in fact have a Make target that [runs actionlint and shellcheck on generated files](https://github.com/pulumi/ci-mgmt/blob/master/provider-ci/Makefile#L44).

We do not suggest it as a regular step when manually adding or generating files however; nor does `make providers` depend on it.

The thing to do here is

1. decide if we want to always run the linter when running `make providers`
2. add a linter Workflow to run on all PRs that touch `provider-ci/providers` files.

PR #608, opened 2023-09-22
Run make lint providers as part of CI

guineveresaenger:
Part of #344.

We'll have to merge this first and then see if it does what we want because Actions need to be merged to default.

### DIFF
--- a/.github/workflows/lint-ecosystem-providers.yml
+++ b/.github/workflows/lint-ecosystem-providers.yml
@@ -1,0 +1,17 @@
+on:
+  pull_request:
+    branches:
+      - master
+    paths:
+      - 'provider-ci/providers/**'
+jobs:
+  lint:
+    name: Run actionlint and shellcheck
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.client_payload.pull_request.head.sha }}
+      - name: lint providers
+        run: cd provider-ci && make lint-providers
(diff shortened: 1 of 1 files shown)


mjeffryes (approved):

.github/workflows/lint-ecosystem-providers.yml
```diff
@@ -0,0 +1,17 @@
...
+      - 'provider-ci/providers/**'
+jobs:
+  lint:
+    name: Run actionlint and shellcheck
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.client_payload.pull_request.head.sha }}
```
danielrbradley, line 15: Why do we specify the SHA manually for this job rather than inherit the default checkout behaviour?

.github/workflows/lint-ecosystem-providers.yml
```diff
@@ -0,0 +1,17 @@
...
+      - name: lint providers
+        run: cd provider-ci && make lint-providers
```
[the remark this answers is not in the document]
guineveresaenger: We talked about this a bit elsewhere, but my thinking here is that I want CI to look the same as the local lint target, plus `make` is handling looking over all providers already, and does shellcheck as well.

.github/workflows/lint-ecosystem-providers.yml
guineveresaenger: That's a great question! I'm not sure why we do this elsewhere - I'll simplify to default.

PR #610, opened 2023-09-25
Fix missing deps for sourcemigrator

t0yv0:
Fixes https://github.com/pulumi/ci-mgmt/issues/609

Here's what happened. As part of rolling out wheels to providers, we've coded automatic upgrades as part of the source-migrator steps in TypeScript, and these steps assume additional build dependencies:

- python
- pulumictl

When testing the workflows on gcp,aws,tls,random, the dependencies were added to some but not all of the workflows. That's he omission currently corrected by this PR.

When wheel rollout is complete we remove both the migrations and the dependencies. Separately we might also consider removing update-workflows-ecosystem-providers since the team split is no more and we're now operating under one risk profile/tradeoffs not two.


.github/workflows/rollout-native-provider-workflows.yml
```diff
@@ -48,6 +48,14 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version: stable
+      - name: Setup Python
```
t0yv0, line 51: Hmm still need to add "unhallow clone for tags".

t0yv0:
Thinking out loud, the root cause here is duplication and forgetting to update the copies. Let me quickly look if we can factor out the shared yaml into a callable action using appropriate GHA facilities.

iwahbe (approved):
Looks good to me.

t0yv0:
I'm going to test every changed job here before merging. Alas testing finds issues in passing boolean arguments.

t0yv0:
Successful test for the bridged provider: https://github.com/pulumi/pulumi-ec/pull/145

t0yv0:
Native provider test: https://github.com/pulumi/ci-mgmt/actions/runs/6303007751

t0yv0:
As part of trying to reduce copy-paste here I noticed a few things. 

- rollout-native-provider-workflows.yml has not been used in 8 months, deleting it
- update-native-provider-workflows.yml is used intend, including nightly, and it calls a loop over rollout-single-native-provider-workflow.yml - this is now using the shared logic with bridged providers after the change
- there are a few subtle differences in native providers
   - they do not want to automerge the workflows (can we allow them to?)
   - instead of replacing all workflow files they do a selective replace, that is they're not fully managed

There's perhaps customizations in the original repos drifted from ci-mgmt. This is all kept as-is for now.

PR #613, opened 2023-09-25
Fix typo

t0yv0:
Looks like this dangling reference is being flagged by actions lint, but is not yet qualifying PRs so we missed that.


t0yv0:
https://github.com/pulumi/pulumi-cloudflare/pull/492

iwahbe (approved):

PR #619, opened 2023-09-26
docker: Upgrade javaGenVersion to v0.9.7

guineveresaenger:
Fixes failing workflow upgrades in pulumi-docker


thomas11 (commented):
Why not do all providers like in #623? (I started it before I saw yours.)

I keep running into the same local build failure because v0.5.x is not compatible with Gradle 8.

t0yv0:
I've commented in https://github.com/pulumi/ci-mgmt/pull/623#pullrequestreview-1646801510 

Every change like this needs to be in sync with the repo manually. Otherwise automated workflow update flow breaks and that prevents other features from rolling out.

t0yv0:
This one actually is in sync with the repo so I'm going to check it in.

PR #621, opened 2023-09-26
Linter needs pulumictl

guineveresaenger:
Apparently, the shellcheck in CI is a bit stricter than locally?

### DIFF
--- a/.github/workflows/lint-ecosystem-providers.yml
+++ b/.github/workflows/lint-ecosystem-providers.yml
@@ -5,6 +5,7 @@ on:
 env:
   # Ignore shell check error 1007
   SHELLCHECK_OPTS: "-e SC1007"
+  GITHUB_TOKEN: ${{ secrets.PULUMI_BOT_TOKEN }}
 jobs:
   lint:
     name: Run actionlint and shellcheck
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-ecosystem-providers.yml
```diff
@@ -2,12 +2,20 @@ on:
   pull_request:
     branches:
       - master
+env:
+  # Ignore shell check error 1007
```
iwahbe, line 6: I would love a comment explaining what error 1007 is

iwahbe (approved):
LGTM

.github/workflows/lint-ecosystem-providers.yml
t0yv0: https://www.shellcheck.net/wiki/SC1007

That seems. BAD.

thomas11: Yeah let's not ignore that. We shouldn't need to. The problem is
- The Makefile expression `VERSION=$(VERSION) go generate ...` has an empty `$VERSION` and therefore evaluates to `VERSION= go…` which the linter correctly flags.
- `$VERSION` is `$(shell pulumictl get version)` but pulumictl is not present, so VERSION is empty.

Installing pulumictl should fix this.

iwahbe: @t0yv0 can we also drop the comment explaining what `SHELLCHECK_OPTS: "-e SC1007"` does?

t0yv0:
@mjeffryes superseded this with another PR.

PR #623, opened 2023-09-27
Upgrade pulumi-java

thomas11:
- Bump pulumi-java to v0.9.7. Works with Gradle 8.
- regenerate Makefiles


t0yv0 (changes requested):
This unfortunately will generate a lot of breakage, this can't go as is. We would need to program in a source migration that automatically regenerates the files. There's prior attempts to do this see:

-

PR #625, opened 2023-09-27
Signalfx is now at v7

t0yv0:
This should unblock automated workflow updates for it.

Applied in https://github.com/pulumi/pulumi-signalfx/pull/303

This fixes https://github.com/pulumi/ci-mgmt/issues/622

I've logged https://github.com/pulumi/upgrade-provider/issues/165 - there's some kind of process/sync failure that left signalfx in this unexpected state.

### DIFF
--- a/provider-ci/providers/signalfx/config.yaml
+++ b/provider-ci/providers/signalfx/config.yaml
@@ -5,3 +5,4 @@ plugins:
   - name: aws 
     version: "4.10.0"
 team: ecosystem
+javaGenVersion: "v0.9.4"
--- a/provider-ci/providers/signalfx/config.yaml
+++ b/provider-ci/providers/signalfx/config.yaml
@@ -1,5 +1,5 @@
 provider: signalfx
-major-version: 6
+major-version: 7
 makeTemplate: bridged
 plugins:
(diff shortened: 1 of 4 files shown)


provider-ci/providers/signalfx/config.yaml
```diff
@@ -1,7 +1,8 @@
 provider: signalfx
-major-version: 6
+major-version: 7
 makeTemplate: bridged
 plugins:
   - name: aws 
     version: "4.10.0"
 team: ecosystem
+javaGenVersion: "v0.9.4"
```
thomas11, line 8: Any reason not to go with latest 0.9.7?

thomas11 (approved):
Not sure where the linter failure comes from, otherwise LGTM

provider-ci/providers/signalfx/config.yaml
t0yv0: That's a great question. This would require me to regenerate singlafx manually to 0.9.7 first, and I'm very much short on time. I'm really hoping we could time into automating java upgrades instead so it regenerates itself and keeps evergreen.

PR #628, opened 2023-09-27
Fix "sed no input files lint error"

t0yv0:
https://github.com/pulumi/home/issues/2883#issuecomment-1737427229

Newly introduced fast linter step fails for a few repos:

rabbitmq - sed: no input files
rke - sed: no input files
consul - lint sed: no input files
wavefront - same

This factors it out into a shared workflow and then fixes the issue. I will test on one of these repos to include a test link here.


t0yv0:
Get past the issue in consul: https://github.com/pulumi/pulumi-consul/pull/226

onto a new issue.

provider-ci/Makefile
```diff
@@ -41,7 +41,7 @@ providers/%/repo: bin/provider-ci providers/%/config.yaml
 
 lint-providers/%/repo: $(ACTIONLINT) providers/%/repo
 	@echo Linting $*
-	$(ACTIONLINT) -config-file actionlint.yml providers/$*/repo/.github/workflows/*
+	cd providers/$*/repo &&	git init && ../../../$(ACTIONLINT) -config-file ../../../actionlint.yml && rm -rf .git
```
thomas11, line 44: Curious why we need to initialize a repo here?

And is there no variable for the root dir to use instead of the fragile `../../../`?

thomas11 (approved):
LGTM modulo comments

provider-ci/Makefile
t0yv0: Yeah so this tool is weird. IT has no option to set the root directory. If I run with `.github/workflows/*.yml` then it flags them that lint.yml reference is dangling - I think because it checks them one by one instead of together. So that's broken... Now if I don't give it any args, it starts listing the ci-mgmt workflows not the provider's. 

So I'm tricking it into checking the right workflows by a temp .git root, like check "HERE".

thomas11: OK but this would really have warranted a comment 🥺

t0yv0: Will add retroactively. You're correct.

PR #633, opened 2023-09-28
Enable single step release on bridged providers

mjeffryes:
Adds the pulumi/action-release-by-pr-label actions to bridged provider workflows to enable the workflow described in https://github.com/pulumi/home/issues/3071

Most of the content of this change is in the actions in pulumi/action-release-by-pr-label, so I would encourage you to review them first on this PR:
https://github.com/pulumi/action-release-by-pr-label/pull/2 (Closed b/c these changes are already merged there, but I still want your review and feedback!)

You can see an example PR using this workflow on the release-demo-test repo:
https://github.com/pulumi/release-demo-test/pull/41
The associated builds for that PR:
- [Command-dispatch](https://github.com/pulumi/release-demo-test/actions/runs/6342858380) (turns slash commands into workflows)
- [release-command](https://github.com/pulumi/release-demo-test/actions/runs/6342860289) (implements the /release behavior)
- [main](https://github.com/pulumi/release-demo-test/actions/runs/6342926965) (builds, then tags the release based on the labels added by release-command)
- [release](https://github.com/pulumi/release-demo-test/actions/runs/6342933137/job/17229727194) (triggered from the tag, builds, publishes, then removes labels from the PRs released)

I will also run this process with an upgrade on a single repo (using workflows copied from this PR) before merging this PR to verify that nothing has been lost in translating these changes into ci-mgmt.

resolves https://github.com/pulumi/home/issues/3071


t0yv0:
I like it. I was trying to wrap my head around the needs/release label removal. The PRs will merge and drop out of the active set so is removing the labels still necessary? Perhaps it does not hurt.

t0yv0:
"Clean up release labels" could come back to the PRs and comment "this is now released as v1.2.3". This would be very cool. Just an enhancement idea though not blocking. I'm excited to try this in anger.

t0yv0:
Can an unauthorized user release something from main branch without having access by just adding labels by hand? Or that's not possible? It has to have the labels and be on main e.g. having gone through review?

iwahbe (commented):
This looks good to me. I'm really excited to try this out.

t0yv0:
Can you walk me through concurrency a bit, what would happen if multiple PRs land on master and several of them are need-release and several are not.  Every commit on master is executing master.yml because we currently don't cancel prior jobs (though we should). So this gets N release-if-needed picking up from HEAD, HEAD~1, HEAD~2, HEAD~3. What will happen?

mjeffryes:
> I like it. I was trying to wrap my head around the needs/release label removal. The PRs will merge and drop out of the active set so is removing the labels still necessary? Perhaps it does not hurt.

I'm removing the labels so that we can easily make a query to look for PRs that should have been released, but failed somehow. @guineveresaenger has mentioned a few times how releases can sometimes fail to complete and be forgotten (since it no longer shows up on our dashboards once the PR is merged). I was thinking we could add a pane to our dashboard that shows PRs merged a while ago that still have a needs-release label on them so our triage runner could check on them.

> "Clean up release labels" could come back to the PRs and comment "this is now released as v1.2.3". This would be very cool. Just an enhancement idea though not blocking. I'm excited to try this in anger.
That's a great idea and should not be hard to implement.

>Can an unauthorized user release something from main branch without having access by just adding labels by hand? Or that's not possible? It has to have the labels and be on main e.g. having gone through review?

I'm pretty sure only repo contributors are able to add labels (and the slash command is similarly protected.)

>Can you walk me through concurrency a bit, what would happen if multiple PRs land on master and several of them are need-release and several are not. Every commit on master is executing master.yml because we currently don't cancel prior jobs (though we should). So this gets N release-if-needed picking up from HEAD, HEAD~1, HEAD~2, HEAD~3. What will happen?

I did not optimize for multiple concurrent releases; I think we should generally try to consolidate and only cut a single release at a time, but here's my analysis of what would happen:
- The "release-if-needed" action for each run will only look at the tags on the PR for the merge commit it's building; so each build that finishes for a PR will invoke release-bot with the version tagged on its PR
- Release bot currently just tags the head of main/master, so the tag created for the earlier PR might include things merged later. This is an existing limitation in our release process, but is a little more risky with automation so I think it's worth coming back to this old release-bot PR (https://github.com/pulumi/release-bot/pull/42) to allow specifying the commit to tag.

t0yv0:
> since it no longer shows up on our dashboards once the PR is merged

Let me have a quick look at the dashboard to see if it can query GitHub releases. Joining that information can let us refine the dashboard.

t0yv0:
So we might end up with multiple tags on the HEAD of the main branch, and run into pulumictl deciding randomly which tag is the right one. It can be  a sticky situation but we can refine later once that happens. LGTM!

t0yv0 (approved):

mjeffryes:
Successfully tested end-to-end on https://github.com/pulumi/pulumi-linode/pull/364!

PR #635, opened 2023-09-29
Enable automerge on the nightly workflow updates for native providers

mjeffryes:
Only aws-native (and soon kubernetes(https://github.com/pulumi/ci-mgmt/pull/634)) get these updates, but if they build successfully, we should just let the bot merge them to reduce drift

### DIFF
--- a/.github/workflows/update-native-provider-workflows.yml
+++ b/.github/workflows/update-native-provider-workflows.yml
@@ -33,7 +33,7 @@ jobs:
     with:
       bridged: false
       provider_name: ${{ matrix.provider }}
-      automerge: false
+      automerge: true
       caller_workflow: update-native-provider-workflows
-      
+
 


t0yv0:
This is a good idea but needs careful rollout so meaningful changes are not lost. azure-native is definitely out of scope as  it has a highly custom set of workflows. The rest need to be evaluated on a case-by-case basis.

mjeffryes:
Yeah I don't see us getting azure in sync with ci-mgmt any time soon; aws-native has been in sync for a while though, and kubernetes is low drift as well. no strong opinions here, but I think it would be safe to run with a default to automerge

iwahbe:
Have if verified that each provider that this will sync to is up to date on workflows? If so, then I'm happy to approve.

iwahbe (approved):
Re-reading, it sounds like only aws-native will get these and it is in sync. LGTM

t0yv0 (approved):

PR #637, opened 2023-09-30
Bump Node version to 20.x

mjeffryes:
Our previous version (16.x) is now unsupported. 20.x is "current", but will be graduating to active LTS at the end of the month, so we shouldn't have to think about it again for a long time.


mjeffryes:
example PR https://github.com/pulumi/pulumi-aws/pull/2843

mjeffryes:
I suppose we probably want to run a release with this as well before shipping to all providers

thomas11 (approved):

PR #638, opened 2023-10-01
Bump publish runner of p-azure to avoid "no space left on device"

thomas11:


danielrbradley (approved):
LGTM - is there not some generated files we'd expect to see in the PR?

thomas11:
> LGTM - is there not some generated files we'd expect to see in the PR?

oops, good call - added

PR #651, opened 2023-10-04
Pin additional plugins for azuread

t0yv0:
These are observed to be auto-installed with `make tfgen`. Running a bit ahead, `azure` is only auto-installed with the new converter, but it would make sense that azuread examples depend on the azure provider.

### DIFF
--- a/provider-ci/providers/azuread/config.yaml
+++ b/provider-ci/providers/azuread/config.yaml
@@ -10,3 +10,10 @@ env:
 makeTemplate: bridged
 team: ecosystem
 javaGenVersion: "v0.9.7"
+plugins:
+  - name: time
+    version: "0.0.15"
+  - name: std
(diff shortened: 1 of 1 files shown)


iwahbe (commented):
This needs to happen in https://github.com/pulumi/pulumi-azuread/blob/master/.ci-mgmt.yaml.

t0yv0:
Well. this doesn't make sense. Which one of the two files is authoritative? If the one you point to is authoritative, why is there still a file left here?

t0yv0:
https://github.com/pulumi/pulumi-azuread/pull/490 like this?

iwahbe:
The new authoritative file is created automatically. Cleaning up old files in ci-mgmt will need to be done piece by piece. I expect to do that by the end of the week.

PR #654, opened 2023-10-04
Add a convenience target to locally regenerate workflow files

iwahbe:
This makes it easier to author PRs such as https://github.com/pulumi/pulumi-azuread/pull/490.

PR #645, opened 2023-10-03
Migrate config.yaml from ci-mgmt to the provider repo

iwahbe:
Storing .ci-mgmt.yaml in the provider's repository will enable single PR upgrades of:
- The java version
- Major provider version
- Everything else controlled by the config file

PR #655, opened 2023-10-04
Template PULUMI_CONVERT rollout

t0yv0:
To be reviewed with https://github.com/pulumi/pulumi-azuread/pull/484

Part of https://github.com/pulumi/pulumi-terraform-bridge/issues/1280

As a pre-requisite to PULUMI_CONVERT being enabled, it's necessary to pin Pulumi CLI so that examples continue to be deterministic. This PR borrows targets from pulumi-azure-native that perform CLI pinning. 

### CLI 

Unlike azure-native, it assumes that the CLI version we want to use to convert examples is the same as pulumi/pkg version we want to use to build this provider. This removes a degree of freedom but buys automatic upgrades - the existing mechanisms that update pulumi/pkg (in upgrade-provider) should also update this Pulumi CLI version. 

### Converter

This is not enforced here but https://github.com/pulumi/pulumi-azuread/pull/484 adds TF converter pinning to the CI-mgmt plugins section. This PR makes changes to allow opting into converter and not just resource plugins. There is currently no mechanism to upgrade this, but presumably upgrade-provider can build mechanisms to upgrade all plugins referenced in this config file. Any such upgrade is likely to trigger the need to regenerate SDKs.


provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -19,6 +19,11 @@ WORKING_DIR := $(shell pwd)
 #{{- if .Config.goBuildParallelism }}#
 PULUMI_PROVIDER_BUILD_PARALLELISM ?= -p #{{ .Config.goBuildParallelism }}#
 #{{- end }}#
+#{{- if .Config.pulumiConvert }}#
+PULUMI_CONVERT := 1
```
t0yv0, line 23: Lifting it here allows to temporarily redefine the value, e.g. `make tfgen PULUMI_CONVERT=1`

provider-ci/internal/pkg/templates/bridged-provider/.pulumi.version
```diff
@@ -0,0 +1 @@
+3.86.0
```
mjeffryes, line 1: to coordinate with Ian's change to move config files into the provider repos, you should probably fill this value using a value from config.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -154,4 +158,12 @@ upstream.rebase:
 bin/pulumi-java-gen:
 	$(shell pulumictl download-binary -n pulumi-language-java -v $(JAVA_GEN_VERSION) -r pulumi/pulumi-java)
 
-.PHONY: development build build_sdks install_go_sdk install_java_sdk install_python_sdk install_sdks only_build build_dotnet build_go build_java build_nodejs build_python clean cleanup#{{ if .Config.docsCmd }}# docs#{{end}}# help install_dotnet_sdk install_nodejs_sdk install_plugins lint_provider provider test tfgen#{{ if eq .Config.team "ecosystem" }}# upstream upstream.finalize upstream.rebase#{{end}}#
+.PHONY: development build build_sdks install_go_sdk install_java_sdk install_python_sdk install_sdks only_build build_dotnet build_go build_java build_nodejs build_python clean cleanup#{{ if .Config.docsCmd }}# docs#{{end}}# help install_dotnet_sdk install_nodejs_sdk install_plugins lint_provider provider test tfgen#{{ if eq .Config.team "ecosystem" }}# upstream upstream.finalize upstream.rebase#{{end}}# upgrade_pulumi
+
+upgrade_pulumi:
+	gh release list --repo pulumi/pulumi --exclude-drafts --exclude-pre-releases --limit 1 | cut -f1 | sed 's/^v//' > .pulumi.version
```
mjeffryes, line 164: And with templated version, to upgrade, you would modify the .cimgmt.yml config file and then regen.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -154,4 +158,12 @@ upstream.rebase:
...
+
+.pulumi/bin/pulumi: PULUMI_VERSION := $(shell cat .pulumi.version)
+.pulumi/bin/pulumi: HOME := $(WORKING_DIR)
+.pulumi/bin/pulumi: .pulumi.version
+	curl -fsSL https://get.pulumi.com | sh -s -- --version "$(PULUMI_VERSION)"
```
mjeffryes, line 169: also maybe move these above the PHONEY target?

provider-ci/internal/pkg/templates/bridged-provider/.pulumi.version
t0yv0: Sensible. Can you give me any links here for what I'm coordinating with - which config also? .cimgmt.yaml or .upgrade-config.yaml? Why are there two?

provider-ci/internal/pkg/templates/bridged-provider/Makefile
t0yv0: Well I just ran into this in pulumi-azuread - this seems to be massively bad. How do I regen? If I edit something in the repo, how can I regenerate things in that same repo (or do nothing) instead of having to go back to ci-mgmt?

t0yv0:
Let me play with this a little more. Ideally Makefile would just read the config file.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
t0yv0: Good point!

provider-ci/internal/pkg/templates/bridged-provider/.pulumi.version
mjeffryes: I think it would be .ci-mgmt.yaml (https://github.com/pulumi/ci-mgmt/pull/645/files) moves the config files to individual repos https://github.com/pulumi/ci-mgmt/blob/master/provider-ci/internal/pkg/templates/bridged-provider.config.yaml would be the place to add the default pulumi version. 

Though, after weighing it some more, maybe we prefer to roll out new pulumi versions to all providers at the same time, in which case we would just want the raw value in this file rather than templating it out from config.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
mjeffryes: Ian has your back: https://github.com/pulumi/ci-mgmt/pull/654

mjeffryes:
actually, maybe don't put the version in a file at all; just make it an variable in the Makefile

provider-ci/internal/pkg/templates/bridged-provider/.pulumi.version
t0yv0: If we're updating this version, we need to re-generate schema and SDK. It might also need to match pulumi/pkg reference in provider/go.mod. Actually why don't we compute it from there .

provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -167,5 +171,9 @@ ci-mgmt: .ci-mgmt.yaml
 		--template bridged-provider \
 		--config $<
 
+.pulumi/bin/pulumi: PULUMI_VERSION := $(shell (cd provider && go list -m github.com/pulumi/pulumi/pkg/v3 ) | awk -F' v' '{print $$2}')
+.pulumi/bin/pulumi: HOME := $(WORKING_DIR)
+.pulumi/bin/pulumi: provider/go.mod
+	curl -fsSL https://get.pulumi.com | sh -s -- --version "$(PULUMI_VERSION)"
```
iwahbe, line 177: I like this. It's a clever way to de-duplicate.

provider-ci/internal/pkg/templates/bridged-provider/.pulumi.version
iwahbe: Do we still need this file? It looks like we now do compute it from the go.mod version.

provider-ci/Makefile
```diff
@@ -50,8 +50,7 @@ lint-providers/%/repo: $(ACTIONLINT) providers/%/repo
 	cd providers/$*/repo &&	git init && ../../../$(ACTIONLINT) -config-file ../../../actionlint.yml && rm -rf .git
 	@scripts/shellcheck.sh providers/$*/repo
 
-LINT_RULES := $(addprefix lint-, $(PROVIDER_REPOS))
-lint-providers: $(LINT_RULES)
+lint-providers: lint-providers/aws/repo lint-providers/docker/repo lint-providers/cloudflare/repo
```
iwahbe, line 53: We'll need to do this eventually, but I don't why we need to do this now. Are you optimizing CI time?

t0yv0: It broke the PR verification.

provider-ci/test-workflows/cloudflare/repo/Makefile
```diff
@@ -139,5 +139,12 @@ ci-mgmt: .ci-mgmt.yaml
 		--template bridged-provider \
 		--config $<
 
+.pulumi/bin/pulumi: HOME := $(WORKING_DIR)
+.pulumi/bin/pulumi: provider/go.mod
+	curl -fsSL https://get.pulumi.com | sh -s -- --version "$$( (cd provider && go list -m github.com/pulumi/pulumi/pkg/v3) | awk -F' v' '{print $$2}')"
```
t0yv0, line 144: Now this creates $WORKING_DIR/go because HOME shifted and it puts the Go mod cache there. Ugh.

provider-ci/test-workflows/docker/repo/Makefile
```diff
@@ -141,5 +141,17 @@ ci-mgmt: .ci-mgmt.yaml
...
+.pulumi/bin/pulumi: .pulumi/version
+	curl -fsSL https://get.pulumi.com | sh -s -- --version $(cat .pulumi/version)
+
+# Compute the version of Pulumi to use by inspecting the Go dependencies of the provider.
+.pulumi/version:
+	@mkdir -p .pulumi
+	@if [ -d provider ]; then \
+		cd provider && go list -f "{{slice .Version 1}}" -m github.com/pulumi/pulumi/pkg/v3; \
+	else \
+		echo "3.86.0"; \
```
t0yv0, line 154: This is added to pass spellcheck when there is no provider folder.

t0yv0:
@iwahbe @mjeffryes take another look please, I'm happier with this version and I got it to work for azuread. Functionally it should be ready to go.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -167,5 +171,17 @@ ci-mgmt: .ci-mgmt.yaml
 		--template bridged-provider \
 		--config $<
 
+.pulumi/bin/pulumi: HOME := $(WORKING_DIR)
+.pulumi/bin/pulumi: .pulumi/version
+	curl -fsSL https://get.pulumi.com | sh -s -- --version $(cat .pulumi/version)
+
+# Compute the version of Pulumi to use by inspecting the Go dependencies of the provider.
+.pulumi/version:
```
iwahbe, line 179: We want this to change when we update our `go.mod`.
```suggestion
.pulumi/version: provider/go.mod
```

provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -106,10 +111,9 @@ install_dotnet_sdk:
 install_nodejs_sdk:
 	yarn link --cwd $(WORKING_DIR)/sdk/nodejs/bin
 
-install_plugins:
-	[ -x "$(shell command -v pulumi 2>/dev/null)" ] || curl -fsSL https://get.pulumi.com | sh
+install_plugins:	.pulumi/bin/pulumi
```
iwahbe, line 114: Nit: formatting.
```suggestion
install_plugins: .pulumi/bin/pulumi
```

provider-ci/Makefile
iwahbe: It shouldn't break PR verification.

iwahbe (approved):
LGTM

provider-ci/internal/pkg/templates/bridged-provider/Makefile
t0yv0: Yeah, no, this breaks the shellcheck linter because it's shell-checking in a context where there is no provider/go.mod... I got a little burned out trying to work around this.

danielrbradley (commented):
Initial thoughts while reviewing via phone...

 I like the use of the go.mod instead of the dedicated version file.

Why do we need to set $HOME?

Do we really need to have a fallback if there's no provider folder? Feels like we'll have bigger issues without that! Perhaps just letting the build fail would be okay?

Why not put these two steps together and just do the fetching of the version and the install in the same target?

provider-ci/internal/pkg/templates/bridged-provider/Makefile
iwahbe: We use shellcheck to prevent us making mistakes. We shouldn't allow it to encourage bad practice. Instead of working around shellcheck, can we just disable it checking that line: https://www.shellcheck.net/wiki/Ignore.

t0yv0:
> Why do we need to set $HOME?

Copied blindly from azure-native. Perhaps there was a reason? 

> Do we really need to have a fallback if there's no provider folder

Eh... no... this is a peculiarity of this repo's shellcheck pass. It is really getting in the way. I'll take another stab at not having to do this.

t0yv0:
> Why not put these two steps together and just do the fetching of the version and the install in the same target?

I kind of like the new target. `make pulumi_version` shows me where we're at.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
t0yv0: https://github.com/pulumi/ci-mgmt/issues/660 that does not really work.

Issue #661, opened 2023-10-06
Run GO tests for bridged providers


t0yv0:
The Makefiles should be running tests under `provider/*` as part of verification. There should be a place to put unit tests in the provider. Currently CI ignores any tests of this kind.

PR #662, opened 2023-10-06
Add unit testing to the provider

t0yv0:
Fixes https://github.com/pulumi/ci-mgmt/issues/661

The reason I'm working on this now is that I'd like to really move upgrade tests out of examples and under provider/ in AWS (and soon GCP). https://github.com/pulumi/pulumi-aws/pull/2855

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/Makefile
+++ b/provider-ci/internal/pkg/templates/bridged-provider/Makefile
@@ -119,13 +119,13 @@ install_plugins: .pulumi/bin/pulumi
 lint_provider: provider
 	cd provider && golangci-lint run -c ../.golangci.yml
 
-provider: tfgen install_plugins test_provider
+provider: tfgen install_plugins
 	(cd provider && go build $(PULUMI_PROVIDER_BUILD_PARALLELISM) -o $(WORKING_DIR)/bin/$(PROVIDER) -ldflags "-X $(PROJECT)/$(VERSION_PATH)=$(VERSION)#{{if .Config.providerVersion}}# -X #{{ .Config.providerVersion }}#=$(VERSION)#{{end}}#" $(PROJECT)/$(PROVIDER_PATH)/cmd/$(PROVIDER))
 
 test:
 	cd examples && go test -v -tags=all -parallel $(TESTPARALLELISM) -timeout 2h
 
-test_provider: tfgen
+test_provider:
 	@echo ""
 	@echo "== test_provider ==================================================================="
 	@echo ""
--- a/provider-ci/internal/pkg/templates/bridged-provider/Makefile
+++ b/provider-ci/internal/pkg/templates/bridged-provider/Makefile
@@ -119,11 +119,14 @@ install_plugins: .pulumi/bin/pulumi
 lint_provider: provider
 	cd provider && golangci-lint run -c ../.golangci.yml
 
-provider: tfgen install_plugins
+provider: tfgen install_plugins test_provider
 	(cd provider && go build $(PULUMI_PROVIDER_BUILD_PARALLELISM) -o $(WORKING_DIR)/bin/$(PROVIDER) -ldflags "-X $(PROJECT)/$(VERSION_PATH)=$(VERSION)#{{if .Config.providerVersion}}# -X #{{ .Config.providerVersion }}#=$(VERSION)#{{end}}#" $(PROJECT)/$(PROVIDER_PATH)/cmd/$(PROVIDER))
 
 test:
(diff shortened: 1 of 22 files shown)


provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -119,12 +119,18 @@ install_plugins: .pulumi/bin/pulumi
...
 	cd provider && golangci-lint run -c ../.golangci.yml
 
-provider: tfgen install_plugins
+provider: tfgen install_plugins test_provider
 	(cd provider && go build $(PULUMI_PROVIDER_BUILD_PARALLELISM) -o $(WORKING_DIR)/bin/$(PROVIDER) -ldflags "-X $(PROJECT)/$(VERSION_PATH)=$(VERSION)#{{if .Config.providerVersion}}# -X #{{ .Config.providerVersion }}#=$(VERSION)#{{end}}#" $(PROJECT)/$(PROVIDER_PATH)/cmd/$(PROVIDER))
 
 test:
 	cd examples && go test -v -tags=all -parallel $(TESTPARALLELISM) -timeout 2h
 
+test_provider: tfgen
```
t0yv0, line 128: I've tested this without tfgen dependency on pulumi-azuread, but it breaks because of go:embed files. That's really not great. I wonder if we could fix that up so that code compiles and is able to test itself even if generated files are not present yet. For now leaving this in.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -119,12 +119,18 @@ install_plugins: .pulumi/bin/pulumi
...
+	@echo ""
+	@echo "== test_provider ==================================================================="
```
t0yv0, line 130: Without the visual marker really easy to get lost in the tfgen noise.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
iwahbe: Which `go:embed` files are missing? I don't think we had that problem with pulumi-aws. We should definitely make sure we don't need the `: tfgen` long term.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -119,12 +119,18 @@ install_plugins: .pulumi/bin/pulumi
 lint_provider: provider
 	cd provider && golangci-lint run -c ../.golangci.yml
 
-provider: tfgen install_plugins
+provider: tfgen install_plugins test_provider
```
iwahbe, line 122: This seems inverted. I generally expect that `test_provider: provider` but not vice versa. 

Instead of making `provider` depend on `test_provider`, can we just add a step in `run-acceptence-tests.yml` (probably `prerequisites`) that calls `make test_provider`?

iwahbe:
FYI: If you want to get tests working "right now", you can add the extra test to `extraTests` and run `make ci-mgmt`. We already have an existing extra test for pulumi-aws:

https://github.com/pulumi/pulumi-aws/blob/0b33b02036fcf1cae354aae98a9d278e957e0199/.ci-mgmt.yaml#L69-L94

provider-ci/internal/pkg/templates/bridged-provider/Makefile
t0yv0: I think it was the bridge-metadata one.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
t0yv0: The reasoning here is that `provider` builds a binary from go sources. Those sources better be tested first. 

> can we just add a step in run-acceptence-tests.yml 

Sad here that run-acceptance-tests.yml is duplicated N ways. I'd have to refactor it first to pull out the common bits. I can give this a look.

t0yv0:
This simpler version seems to work in:

https://github.com/pulumi/pulumi-azuread/actions/runs/6437059888/job/17481565040#step:10:37

By running tests after building as a separate step, this bypasses the need to model the dependency in Make. A bit awkward but the least evil quick option at the moment it seems like.  Locally, you just have to "know" you need to run tfgen before.

@iwahbe another look?

iwahbe (approved):
LGTM

Nit: It would be great to add a comment in the `Makefile` explaining that `test_provider` needs `tfgen`, but we haven't modeled the dependency because of performance reasons.

t0yv0:
I think in the ideal state it really should not need tfgen. There is the issue with `go:embed` pointing to non-existing files; this is really a papercut, this can be worked around. Beyond that, why should testing the provider depend on tfgen?

PR #669, opened 2023-10-11
Upgrade setup-dotnet to v3

mikhailshilkov:
Lots of failures across the board:
- https://github.com/pulumi/pulumi-eks/actions/runs/6478619704/job/17590794514
- https://github.com/pulumi/pulumi-gcp/actions/runs/6478614377/job/17590794352
- https://github.com/pulumi/pulumi-eks/issues/922
- https://github.com/pulumi/pulumi-azure/issues/1376
- https://github.com/pulumi/pulumi-awsx/issues/1098

### DIFF
--- a/native-provider-ci/providers/aws-native/repo/.github/workflows/build.yml
+++ b/native-provider-ci/providers/aws-native/repo/.github/workflows/build.yml
@@ -171,7 +171,7 @@ jobs:
         node-version: ${{ env.NODEVERSION }}
         registry-url: https://registry.npmjs.org
     - name: Setup DotNet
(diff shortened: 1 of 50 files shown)


guineveresaenger (changes requested):
We'll need to make changes to https://github.com/pulumi/pulumi-package-publisher to use v3 first. Cut a new tag for that tool, and then update this PR to update the package publisher.

Overall, to make the versions we use less lossy between microservices, we could consider making the action versions Inputs to pulumi-package-publisher, so we don't need multiple PRs every time we migrate to a new action version?

iwahbe (approved):

PR #671, opened 2023-10-11
Use pulumi-ubuntu-8core runner for Azure prerequisites CI step

rquitales:
This PR adds a new parameterized runner for the `prerequisite` job of each bridged CI workflow. The default is for this runner is still `ubuntu-latest`, but `pulumi-ubuntu-8core` is used for pulumi-azure to unblock the out of disk space issue.


guineveresaenger (approved):
thank you!

PR #672, opened 2023-10-12
Remove unused workflow files

iwahbe:
This PR removes `${PROVIDER}/config.yaml` files, since they are no longer used as the source of truth for the values they contain.

Previously, the set of providers was calculated by looking at the set of directories, which obviously will no longer work. Instead we just store the (previously calculated) JSON list directly.


provider-ci/Makefile
```diff
@@ -30,6 +30,11 @@ test:
 format:
 	go fmt ./...
 
+providers/%/config.yaml: F=providers/$*/config.yaml
+providers/%/config.yaml:
+	mkdir -p $(dir $F)
+	curl "https://raw.githubusercontent.com/pulumi/pulumi-$*/master/.ci-mgmt.yaml" > $F
```
t0yv0, line 36: What if the branch is `main`?

t0yv0 (approved):
LGTM with one nit.

provider-ci/Makefile
iwahbe: I just tested that. Github re-interprets `master` as `main`. I can make it explicit.

t0yv0: Oh nice of them!

t0yv0: no need then.

PR #684, opened 2023-10-13
Apply automerge to bridge updates

iwahbe:
We have merged all workflow updates, so this will work.


.github/workflows/update-bridge-ecosystem-providers.yml
```diff
@@ -47,6 +47,7 @@ jobs:
             workflow: upgrade-bridge.yml
             token: ${{ secrets.PULUMI_BOT_TOKEN }} 
             repo: pulumi/pulumi-${{ matrix.provider }}
+            inputs: '{ "automerge": true }'
```
t0yv0, line 50: What is going on with the syntax here.. It has to be JSON escaped?

t0yv0 (approved):

.github/workflows/update-bridge-ecosystem-providers.yml
iwahbe: Yep

PR #685, opened 2023-10-13
Toward improved downstream checks

t0yv0:
Upgrade bridge workflow in each provider is extended to propagate reviewers and extra description, as well as take custom bridge versions as a parameter.

Tested end to end in https://github.com/pulumi/pulumi-aiven/pull/398

Blocked on an updated version of the pulumi-upgrade-provider-action that understands the new params.


iwahbe (approved):
This looks ok to me.

PR #687, opened 2023-10-16
Share prerequisites.yml code

t0yv0:
Coming into this with a need to do a quick edit to prerequisites.yml I'm noticing that it got duplicated with some drift across 5 workflows. So as the first step I'd like to recover explicit sharing with this PR to make it less error prone to make changes.


provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/prerequisites.yml
```diff
@@ -0,0 +1,88 @@
...
+    - name: Tar provider binaries
+      run: tar -zcf ${{ github.workspace }}/bin/provider.tar.gz -C ${{
+        github.workspace }}/bin/ pulumi-resource-${{ env.PROVIDER }}
+        pulumi-tfgen-${{ env.PROVIDER }}
+    - name: Upload artifacts
+      uses: #{{ .Config.actionVersions.uploadArtifact }}#
+      with:
+        name: ${{ env.PROVIDER }}-provider.tar.gz
+        path: ${{ github.workspace }}/bin/provider.tar.gz
+        retention-days: 30
```
t0yv0, line 81: There was drift here, some workflows did not specify retention-days. Enabling it everywhere. Possibly for release.yml it should be longer?

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/prerequisites.yml
```diff
@@ -0,0 +1,88 @@
...
+
+env:
+#{{ .Config.env | toYaml | indent 2 }}#
+
+jobs:
+  prerequisites:
+    name: prerequisites
+    runs-on: #{{ .Config.runner.prerequisites }}#
+    steps:
+    - name: Checkout Repo
```
t0yv0, line 17: There was drift here, in that run-acceptance-tests.yml additionally specified "ref":

```
>         ref: ${{ env.PR_COMMIT_SHA }}
```

In theory this should not be needed? Going to run through a test PR.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/prerequisites.yml
```diff
@@ -0,0 +1,88 @@
...
+      uses: #{{ .Config.actionVersions.installGhRelease }}#
+      with:
+        repo: pulumi/pulumictl
+#{{ .Config.actions.setupPulumi | toYaml | indent 4 }}#
+    - if: github.event_name == 'pull_request'
+      name: Install Schema Tools
+      uses: #{{ .Config.actionVersions.installGhRelease }}#
+      with:
+        repo: pulumi/schema-tools
+#{{- if .Config.actions.preBuild }}#
```
t0yv0, line 41: There was drift here, not every workflow specified this. How this is used, only for AWS provider:

preBuild users jlumbroso/free-disk-space@main

t0yv0:
https://github.com/pulumi/pulumi-cloudflare/actions/runs/6536508001/job/17748366240?pr=521

The first test did not work as PROVIDER env var is undefined. Iterating.

t0yv0:
https://github.com/pulumi/pulumi-cloudflare/pull/522

Eh. This changed the name of the prerequisites job as specified in the required PR checks. So this breaks branch protections...

iwahbe:
> [pulumi/pulumi-cloudflare#522](https://github.com/pulumi/pulumi-cloudflare/pull/522)
> 
> Eh. This changed the name of the prerequisites job as specified in the required PR checks. So this breaks branch protections...

If it's helpful, just make sure that sentinel correctly has a `needs` relationship with the new job. We can remove other branch protections.

t0yv0:
That could be reasonable. I recall some uncertainty on whether sentinel by itself is sufficient to avoid auto merge PRs with failing tests, there was some prior history there. 

I was thinking another workaround here is to not deduplicate the code at GHA level but at macro level, so that it's expanded back as a no-op at GHA level.

iwahbe:
Sentinel is special. It enforces its needs behavior more strictly, so it can work to avoid auto-merging PRs.

https://github.com/pulumi/ci-mgmt/blob/0dd9cf1fdda4b9dae4dbd03528223a54e5b31b0c/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/run-acceptance-tests.yml#L210-L252

t0yv0:
That's a good description there. So in that case, if we trust sentinel, should we remove all checks from the branch protections except the sentinel check? I see the case of a test provider here, CI is waiting for "prerequisites" but also for "tests" jobs separately as required checks as part of the branch protections.  We have a Pulumi stack managing these and it would be something easy to accomplish, would it be reasonable? By the reasoning put forward here, these branch protection checks are redundant because only the sentinel check needs to be verified.

t0yv0:
https://github.com/pulumi/ci-mgmt/issues/693 is it.

iwahbe:
> That's a good description there. So in that case, if we trust sentinel, should we remove all checks from the branch protections except the sentinel check? I see the case of a test provider here, CI is waiting for "prerequisites" but also for "tests" jobs separately as required checks as part of the branch protections. We have a Pulumi stack managing these and it would be something easy to accomplish, would it be reasonable? By the reasoning put forward here, these branch protection checks are redundant because only the sentinel check needs to be verified.

I think it is safe to pair back on branch protections. I hesitate to remove all checks except sentinel, just because I don't fully trust GH to be bug free here. In theory, we only need to depend on sentinel.

t0yv0:
Test succeeded: https://github.com/pulumi/pulumi-cloudflare/pull/524

I rewrote the PR. Instead of using GHA-level sharing with a shared workflow, I lifted this to a macro-level sharing where the shared code is copied at the time of expanding workflow definitions (build time). This makes it possible to share code without GHA knowing it's shared, and therefore without interference with branch protections.

Adds some complexity in here but I think this is a degree of freedom worth having while we figure out how to do GHA better.

provider-ci/test-workflows/aws/.github/workflows/run-acceptance-tests.yml
```diff
@@ -156,7 +156,6 @@ jobs:
     - name: Checkout Repo
       uses: actions/checkout@v3
       with:
-        ref: ${{ env.PR_COMMIT_SHA }}
```
t0yv0, line 159: I believe this should be a safe default, most of our repos do not reference PR_COMMIT_SHA when calling the checkout action.

provider-ci/internal/pkg/generate.go
```diff
@@ -124,6 +138,35 @@ func GeneratePackage(opts GenerateOpts) error {
...
+		}
+		if !strings.HasSuffix(path, ".splice") {
+			return nil
+		}
+		tmpl, err := parseTemplate(templateFS, path)
+		if err != nil {
+			return fmt.Errorf("error parsing template %s: %w", path, err)
+		}
+		var buf bytes.Buffer
+		if err := tmpl.Execute(&buf, tc); err != nil {
```
iwahbe, line 158: This is how splices also get to be templates?

It's not fully recursive, which is nice. I'm calling this out here because I missed it on the first pass and it's an important part of understanding this PR.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/run-acceptance-tests.yml
```diff
@@ -128,85 +128,13 @@ jobs:
...
-        } >> "$GITHUB_ENV"
-    - if: github.event_name == 'pull_request' && github.actor != 'dependabot[bot]'
-      name: Comment on PR with Details of Schema Check
-      uses: #{{ .Config.actionVersions.prComment }}#
-      with:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        comment_tag: schemaCheck
-        message: >+
-          ### Does the PR have any schema changes?
+#{{ .Splices.prerequisites }}#
```
iwahbe, line 136: The prerequisite job will always be called with:

```yaml
 prerequisites:
    name: prerequisites
#{{ .Splices.prerequisites }}#
```

Can we move the `prerequisites:\n\tname: prerequisites` into the splice? We can still accommodate this job's extra `if:` condition:

```yaml
#{{ .Splices.prerequisites }}#
    if: github.event_name == 'repository_dispatch' ||
      github.event.pull_request.head.repo.full_name == github.repository
```

iwahbe (approved):
I think we have room for a little more de-duplication. In general, this is great.

It's unfortunate that we need to work around the limitations of GH Actions, but I agree with the need.

Sorry for taking so long to review.

LGTM

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/run-acceptance-tests.yml
t0yv0: I prefer the little bit of duplication here actually so that it gives a visual marker that there is a job called "prerequisites". I think we're not going to be renaming that anytime soon so this duplication is OK. The body is shared.

provider-ci/internal/pkg/generate.go
t0yv0: Certainly not fully recursive, you cannot use splices in splices. I thought about this and figured YAGNI or we'll build it when we need it.

PR #689, opened 2023-10-16
Upgrade pulumi upgrade provider action v0.0.9

t0yv0:
Routine dependency update: 

Upgrade pulumi upgrade provider action to v0.0.9


t0yv0:
https://github.com/pulumi/pulumi-aiven/actions/runs/6513420885/job/17693006653 tested the pre-release version that is now tagged v0.0.9.

PR #691, opened 2023-10-16
Permit x-repo triggers

t0yv0:
When testing cross-repo dispatch where an action in pulumi-terraform-bridge is supposed to call and start an action in pulumi-aiven, I noticed that the dispatch event is quietly being lost. I think the piece edited here is the missing configuration to get this working.

This is now tested per https://github.com/pulumi/pulumi-terraform-bridge/pull/1442 and aiven.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-bridge.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-bridge.yml
@@ -2,9 +2,10 @@
 
 name: Upgrade bridge
 on:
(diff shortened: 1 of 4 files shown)


t0yv0:
Ah nvm, this is called repository_dispatch. Ignore this while I peruse the docs.

```
  repository_dispatch:
    types:
      - pulumi-cli
```

iwahbe (approved):
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#repository_dispatch makes it seem like this is necessary. 😒 

LGTM

PR #696, opened 2023-10-18
Makefile and upstream.sh fixups

### LINKED ISSUE #694: Upstream rebasing gets into a broken state
Currently, we track if we're in a rebase by touching the file `rebase-in-progress` and then removing it once complete.

- The existance of this file can get out of sync with the actual state of the upstream submodule because we get users to perform manual actions in the submodule repo (e.g. solving conflicts).
- We lack the ability to track if the upstream repo is still in the middle of the rebase.

## Proposal

Query the actual git state of the upstream module to determine if it's in the expected state:

```bash
upstream_state() {
  current_branch=$(cd upstream && git branch --show-current)
  if [[ -z ${current_branch} ]]; then
    echo "clean"
  fi
  if [[ "${current_branch}" != "pulumi-patch" ]]; then
    echo "unexpected_branch: ${current_branch}"
  fi
  # Check for "rebase in progress" in status
  git_status=$(cd upstream && git status --porcelain)
  if [[ ${git_status} == *"rebase in progress"* ]]; then
    echo "rebasing"
  fi
  echo "rebased"
}
```

This can be called and the echoed result captured. We use `set -e;` otherwise bash doesn't propegate this setting into subshell function calls.

```bash
state=$(set -e; upstream_state)
```

This can then be queried on such as:

```bash
describe_state() {
  case $1 in
    clean)
      echo "The upstream repository is in a clean state."
      ;;
    rebasing)
      echo "The upstream repository is currently being rebased."
      ;;
    rebased)
      echo "The upstream repository has been rebased."
      ;;
    unexpected_branch*)
      branch_name=$(echo "$1" | cut -d' ' -f2)
      echo "The upstream repository is on an unexpected branch: ${branch_name}."
      ;;
    *)
      echo "Unknown state: $1"
      ;;
  esac
}
```

danielrbradley:
## [Simplify invoking scripts](https://github.com/pulumi/ci-mgmt/commit/2748725404c8830bdf8df4699c01acb5a310de19)
Add shebang to scripts/upstream.sh to specify its own expected shell and make it executable. Therefore...

1. Remove unneeded shell prefix in make (the shebang makes this obsolete).
2. Remove `@` prefix so we can see what we're running in case something goes wrong. Otherwise diagnosing issues requires reading the makefile and having to manually substitute in variables. This also provides a helpful record of commands in CI for troubleshooting.

## [Unnest pulumi-java-gen binary download command](https://github.com/pulumi/ci-mgmt/commit/c8a250d8388c1dbde140ee947856d22f3a639816) 

Using `$(shell ...)` inside a target is an anti-pattern as it can break the expected execution order, break --what-if, and hides commands being executed from logs.

## [Make patches less noisy](https://github.com/pulumi/ci-mgmt/commit/40caf9a5dd0d7fc15229c75cf21e6b6428b7a6d3) 

1. Remove the file path prefix of "a/" or "b/" in summaries as this isn't useful information.
2. Set source commit hash to all zeros otherwise this changes after every rebase.
3. Remove the git signature which is generally just the version number of git used to extract the patch.
4. Remove stats which just adds a big block of stats at the top of each diff which isn't useful to maintain.

## [Reset back to "HEAD mode" after finishing a rebase](https://github.com/pulumi/ci-mgmt/commit/2d9e7c0d8baba2306cbca406c511677120f1553f) 

Instead of leaving the user on a branch, move back to the referenced submodule commit ready to re-apply the patches.

## [Warn if finalizing mid-rebase](https://github.com/pulumi/ci-mgmt/commit/e8eb388fd4df311b011902ddbca51c2706d4a0dc) 

If we half way through a rebase, it's certainly an error to then attempt to extract the patches - as the commits not yet rebased will be missing from the patch set.

This is probably good enough a fix to close #694


danielrbradley:
Looks like we don't maintain file permissions when copying the script ... just fixing up now.

provider-ci/internal/pkg/templates/bridged-provider/scripts/upstream.sh
```diff
@@ -163,6 +177,7 @@ EOF
 # It is used to make the error message more actionable.
 end_rebase() {
   assert_rebase_in_progress "$1"
+  assert_rebase_not_unfinished "$1"
 
   if [ -d .git/modules/upstream/rebase-merge ]; then
```
mjeffryes, line 182: This test looks to be redundant with the `assert_rebase_not_unfinished` you just added.

mjeffryes (approved):
This really looks like a tool that should standalone and have some tests rather than being copied into every provider repo. However the proposed changes look reasonable.

provider-ci/internal/pkg/generate.go
```diff
@@ -177,6 +177,14 @@ func renderTemplateFile(tmpl *template.Template, outPath string, ctx templateCon
...
 		return err
 	}
+
+	// Make shell scripts executable
+	if strings.HasSuffix(outPath, ".sh") {
+		err = os.Chmod(outPath, 0755)
+		if err != nil {
+			return err
+		}
+	}
```
iwahbe, line 187: Git does not maintain file permissions AFAIK. We shouldn't need this, and I don't believe it does anything for us.

provider-ci/internal/pkg/templates/bridged-provider/scripts/upstream.sh
iwahbe: It isn't. `assert_rebase_in_progress` checks that `make upstream.rebase` has been called.

provider-ci/internal/pkg/templates/bridged-provider/scripts/upstream.sh
```diff
@@ -172,8 +187,9 @@ end_rebase() {
 
   rm patches/*.patch
   cd upstream
-  git format-patch local -o ../patches
+  git format-patch local -o ../patches --no-prefix --zero-commit --no-signature --no-stat
   rm ../rebase-in-progress
+  git submodule update --force --init
```
iwahbe, line 192: This should just call the function to reset upstream.
```suggestion
  apply "$1"
```

This has the nice property of being DRY and asserting that the new patch set applies cleanly.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -138,7 +138,7 @@ tfgen: install_plugins upstream#{{ if .Config.docsCmd }}# docs#{{ end }}#
 
 upstream:
 ifneq ("$(wildcard upstream)","")
-	@$(SHELL) ./scripts/upstream.sh "$@" apply
+	scripts/upstream.sh "$@" apply
```
iwahbe, line 141: I would rather we called `scripts/upstream.sh` as a script, instead of as an executable. If you don't want to use `$(SHELL)`, we can specify bash directly.

```suggestion
	bash ./scripts/upstream.sh "$@" apply
```

iwahbe (changes requested):

provider-ci/internal/pkg/templates/bridged-provider/scripts/upstream.sh
mjeffryes: No I mean the check in line 182 and following.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
mjeffryes: What's the motivation for running it in a subshell?

provider-ci/internal/pkg/templates/bridged-provider/scripts/upstream.sh
```diff
@@ -130,6 +131,19 @@ EOF
   fi
 }
 
+assert_rebase_not_unfinished() {
+  git_status=$(cd upstream && git status --verbose)
```
iwahbe, line 135: Is it possible to get this with `git status --porcelain`? `--verbose` depends on your `git` implementation/version but `--porcelain` is part of `git`'s stable interface.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
iwahbe: AFAIK it will be run in a sub shell regardless. I would prefer we invoke it via the executor to avoid worrying about executability.

provider-ci/internal/pkg/templates/bridged-provider/scripts/upstream.sh
iwahbe: Oh... you're 100% right. Ignore me

provider-ci/internal/pkg/generate.go
danielrbradley: Yup - git maintains these. File archives don't maintain permissions.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
danielrbradley: Executability is what the shebang is for (`#!/usr/bin/env bash`)

[Wikipedia Reference](https://en.wikipedia.org/wiki/Shebang_%28Unix%29)

provider-ci/internal/pkg/templates/bridged-provider/scripts/upstream.sh
danielrbradley: No, porcelain doesn't include rebase status. It's not guarenteed to be there so it's possible we might not always prevent this case, but if it is there, we can be sure that the user is doing something wrong.

provider-ci/internal/pkg/templates/bridged-provider/scripts/upstream.sh
danielrbradley: `[ -d .git/modules/upstream/rebase-merge ]` doesn't work everywhere either - this completely failed to catch my own local rebase. Here's a good discussion of trying to catch a rebase ... essentially, there's no reliable way to do this https://stackoverflow.com/questions/3921409/how-to-know-if-there-is-a-git-rebase-in-progress

mjeffryes: The SO answer you linked suggests that 
```
(test -d "$(git rev-parse --git-path rebase-merge)" || \
 test -d "$(git rev-parse --git-path rebase-apply)" )
```
is probably the most reliable option. 

At any rate, I'd just like to consolidate the old and new checks in a single place.

danielrbradley: Yup, agree, that seems to be a good in-between. Tested locally and this does actually catch the rebase unlike before.

danielrbradley:
@iwahbe I think all the converations should be resolved now with the follow-up changes now in. Were there any other changes outstanding that you see as blocking this getting merged?

provider-ci/internal/pkg/templates/bridged-provider/scripts/upstream.sh
danielrbradley: Remove this whole approache now.

provider-ci/internal/pkg/templates/bridged-provider/scripts/upstream.sh
```diff
@@ -162,18 +163,23 @@ EOF
 # "$1" is the current make command being run.
 # It is used to make the error message more actionable.
 end_rebase() {
-  assert_rebase_in_progress "$1"
+  assert_rebased "$1"
```
iwahbe, line 166: Where is `assert_rebased` defined?

provider-ci/internal/pkg/templates/bridged-provider/scripts/upstream.sh
```diff
@@ -162,18 +163,23 @@ EOF
...
 
-  if [ -d .git/modules/upstream/rebase-merge ]; then
+  rebase_merge_dir=$(cd upstream && git rev-parse --git-path rebase-merge)
+  rebase_apply_dir=$(cd upstream && git rev-parse --git-path rebase-apply)
```
iwahbe, line 169: A comment explaining what this does would be appreciated.

provider-ci/internal/pkg/templates/bridged-provider/scripts/upstream.sh
danielrbradley: Ah - this was meant to go back to `assert_rebase_in_progress`. Fixed

iwahbe (approved):

PR #700, opened 2023-10-27
Support --target-pulumi-version

t0yv0:
Companion to https://github.com/pulumi/pulumi-upgrade-provider-action/pull/21

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-bridge.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-bridge.yml
@@ -12,6 +12,17 @@ on:
         required: false
         type: string
         default: "latest"
+      target-pulumi-version:
+        description: |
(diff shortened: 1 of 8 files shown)


provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-bridge.yml
```diff
@@ -12,6 +12,17 @@ on:
         required: false
         type: string
         default: "latest"
+      target-pulumi-version:
+        description: |
+             Set the version of `pulumi/pkg` and `pulumi/sdk` to depend on for bridged providers. Currently,
+             these versions inform the linked runtime and SDK generation in all languages except Java. Valid
+             options are:
+             - "", "follow": Use the same version as pulumi-terraform-bridge
```
iwahbe, line 20: The final implementation doesn't support `"follow"`.
```suggestion
             - "": Use the same version as pulumi-terraform-bridge
```

iwahbe (approved):
1 Fix. Otherwise LGTM

PR #705, opened 2023-10-31
Add a license check on PRs to providers

iwahbe:
TLDR We should merge this PR to add automatic license checks to our providers. Failing workflows can be addressed in real time without blocking other processes. If this causes to much noise, we can always revert.

### Description
This is necessary to avoid someone manually reviewing each provider upgrade for license compliance. This PR as introduced is a required check.

pulumi/license-check-action is not total, so we also maintain an allow list of paths that are not checked. I have assumed that golang.org and github.com/golang will only contain repos that are valid to include, since they are both open source projects. We own github.com/pulumi, so I have marked that safe as well (it was flagged as unable to find license on the bridge). Other accept-listed repos I have manually checked for a OSS License (Apache, MIT or BSD).

I have confirmed this accept-list works on:
- https://github.com/pulumi/pulumi-digitalocean/actions/runs/6712403292/job/18241901579?pr=524
- https://github.com/pulumi/pulumi-external/actions/runs/6712674064/job/18242730821?pr=49
- https://github.com/pulumi/pulumi-akamai/actions/runs/6712256033/job/18241417416?pr=315
- https://github.com/pulumi/pulumi-aws/actions/runs/6712451436/job/18242044084?pr=2946

### Rollout Plan

When we merge this PR, it will automatically merge to all of our providers via a "Update GitHub Actions workflows" PR (this goes out every night). If the allow-list is complete, and all of our providers are using fully OSS software, then the automated PRs will all merge.

Failing PRs ([visible in real time here](https://github.com/search?q=org%3Apulumi+Update+GitHub+Actions+workflows.+is%3Aopen+AND+NOT+in%3Atitle+%22Bump+%22&type=pullrequests)) will be resolved manually.

License checks will immediately block releases, so at this point any provider with a passing "License Check" workflow is then safe to merge without a manual license check.

### Rollback Plan

I'm nervous about the level of noise this check might cause, especially due to adding new paths to the allow list. If the tool proves too noisy to use, we can revert this PR. The license check will be removed from our providers the next time workflows are updated (nightly).


provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/license.yml
```diff
@@ -0,0 +1,46 @@
...
+  license_check:
+    name: License Check
+    runs-on: #{{ .Config.runner.default }}#
+    steps:
+    - name: Checkout Repo
+      uses: actions/checkout@v3
+      with:
+        ref: ${{ env.PR_COMMIT_SHA }}
+    - name: Unshallow clone for tags
+      run: git fetch --prune --unshallow --tags
```
mjeffryes, line 23: does this check actually need tags?

mjeffryes (approved):

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/license.yml
iwahbe: We calculate the version when running `make upstream` using the local tags. I don't think that we actually use anything that uses the version, but since the makefile generically assumes they are there, this seemed like the safest option until we figure out the actual contract we want to support.

lukehoban:
> I have confirmed this accept-list works on:

Have we confirmed the opposite, that this will correctly still fail an attempt to bring in a dependency on BSL code?

>  so we also maintain an allow list of paths that are not checked

Are these ignored shallowly or deeply?  That is - if a BSL dependency snuck into some `github.com/pulumi/something` repo (but the something repo itself was still Apache 2.0), and then a dependency on that repo was pulled into something being checked here, would we still flag the transitive BSL dependency, even though it was introduced though an ignored dependency?

iwahbe:
> > I have confirmed this accept-list works on:
> 
> Have we confirmed the opposite, that this will correctly still fail an attempt to bring in a dependency on BSL code?

Yes: https://github.com/pulumi/pulumi-external/actions/runs/6713846228/job/18246117223?pr=49. This commit depends on `github.com/hashicorp/terraform/version`, and it fails as expected.

> > so we also maintain an allow list of paths that are not checked
> 
> Are these ignored shallowly or deeply? That is - if a BSL dependency snuck into some `github.com/pulumi/something` repo (but the something repo itself was still Apache 2.0), and then a dependency on that repo was pulled into something being checked here, would we still flag the transitive BSL dependency, even though it was introduced though an ignored dependency?

I would guess that the tool checks every dependency listed in `go.mod`, and thus the ignore list is shallow. The [documentation](https://github.com/google/go-licenses#ignoring-packages) concurs:
> Note that dependencies from the ignored packages are still resolved and checked. This flag makes effect to check, report and save commands.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/license.yml
```diff
@@ -0,0 +1,46 @@
...
+      uses: actions/setup-go@v4
+      with:
+        cache-dependency-path: |
+            sdk/go.sum
+        go-version: 1.21.x
+    - run: make upstream
+    - uses: pulumi/license-check-action@main
+      with:
+        module-path: provider
+        ignore-modules: >-
```
kmosher, line 34: I identified the following for `pulumi-snowflake`:
github.com/apache/arrow/go/v12
github.com/apache/thrift/lib/go/thrift

kmosher (approved):
Tried to do a wider test, across all providers, but I was screwing up my automation somewhere. I think at this point, let's do a test push and see how much bleeding there is.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/license.yml
iwahbe: Accepted in https://github.com/pulumi/ci-mgmt/pull/705/commits/a444b57a9c4cd7ed2e28d41ba003b40fcaf10f0d.

lukehoban (approved):
🚢

PR #711, opened 2023-11-03
Allow custom license ignores

iwahbe:
This PR is best reviewed commit by commit.

- Standardize formatting (06b100170111e156785223b799eb49d972b05d04)
- `license.ignore` needs to be a config option because shimmed providers need a way to allow their own shim. (4b234f9c98cc1816d41ed9d062b07fa60441406e)
- Add a package this is used by pulumi-alicloud (8e3db535e3f8744a5ec7bde02008481060f7c8a3)


provider-ci/providers/cloudflare/config.yaml
```diff
@@ -10,3 +10,6 @@ env:
   CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
   CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
 javaGenVersion: "v0.9.3"
+license:
+  ignore:
+    - github.com/oracle/terraform-provider-oci/shim
```
mjeffryes, line 15: IIUC, we could also work around this by converting this provider from shim to patch right?

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/license.yml
```diff
@@ -42,6 +42,7 @@ jobs:
           github.com/golang,
           github.com/gorhill/cronexpr,
           github.com/jmespath/go-jmespath,
+          github.com/keybase/go-crypto,
```
mjeffryes, line 45: I'm surprised that we need to allow list this one, the license check tool should be able to approve its license

mjeffryes (approved):
Seems reasonable not to block on these, but in both cases, it would be nice if we could resolved the underlying issue that creates the need.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/license.yml
iwahbe: It misses a surprising number of LICENSEs.

provider-ci/providers/cloudflare/config.yaml
iwahbe: We could. Shim's are generally less complexity than a patch, so I'm hesitant to do that. We can come back to this if you feel strongly.

mjeffryes: Oh, I was under the impression that we preferred patched to shimmed. FWIW I think it might be nicer to just convert all shims to a 1 patch provider if only for consistency.

iwahbe: I think we prefer (best to worst):
- plain
- shimmed
- patched
- forked (only vault because we can't upgrade the bridge)

PR #713, opened 2023-11-07
Add devbox config to providers

mjeffryes:
Adds templates for devbox.json and .devcontainer to bridged providers with the goal of simplifying environment setup for building providers. 

As part of this work, I refactored the gradle version to be explicit in the config file so we can ensure that the devbox environment uses the same version.

I'm explicitly not including the pulumi cli or any plugins in the devbox environment. Keeping pulumi up to date in nixpkgs is not something I want to take on right now, and we already have a decent system for loading a pinned version in the Makefile (at least for the bridged providers).

### DIFF
--- a/provider-ci/providers/aws/config.yaml
+++ b/provider-ci/providers/aws/config.yaml
@@ -13,6 +13,9 @@ checkoutSubmodules: true
 # TODO: remove XrunUpstreamTools flag after work to add docs replacement strategies to resources.go is completed
 # Tracked in in https://github.com/pulumi/pulumi-aws/issues/2757
 XrunUpstreamTools: true
+nixpkgs:
+  - name: awscli2
+    version: latest
 plugins:
   - name: archive
     version: "0.0.1"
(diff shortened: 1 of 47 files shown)


provider-ci/providers/aws/config.yaml
```diff
@@ -13,6 +13,9 @@ checkoutSubmodules: true
 # TODO: remove XrunUpstreamTools flag after work to add docs replacement strategies to resources.go is completed
 # Tracked in in https://github.com/pulumi/pulumi-aws/issues/2757
 XrunUpstreamTools: true
+nixpkgs:
```
t0yv0, line 16: Suspect this file is a no-op and https://github.com/pulumi/pulumi-aws/blob/master/.ci-mgmt.yaml is the source of truth instead?

t0yv0 (approved):
Thank you!

provider-ci/providers/aws/config.yaml
mjeffryes: I think that's correct. For this PR I just wanted to put one in to verify that the template works correctly.

PR #714, opened 2023-11-07
Enable setting custom Pulumi CLI version to install for native providers

rquitales:
This PR changes the Pulumi CLI installation step to use `pulumi/actions@v4` which allows us to pin a specific version of the CLI to install. A new config variable: `pulumiCLIVersion` allows us to specify if a version should be pinned.

This PR also pins the Pulumi CLI version to v3.89.0 for the Kubernetes provider.

### DIFF
--- a/native-provider-ci/providers/kubernetes/config.yaml
+++ b/native-provider-ci/providers/kubernetes/config.yaml
@@ -3,6 +3,7 @@ lint: true
 aws: true
 gcp: true
 major-version: 4
+pulumiCLIVersion: v3.89.0
 env:
   AWS_REGION: us-west-2
   PULUMI_TEST_OWNER: moolumi
--- a/native-provider-ci/providers/aws-native/repo/.github/workflows/build.yml
+++ b/native-provider-ci/providers/aws-native/repo/.github/workflows/build.yml
@@ -65,7 +65,7 @@ jobs:
       with:
         repo: pulumi/pulumictl
     - name: Install Pulumi CLI
-      uses: pulumi/action-install-pulumi-cli@v2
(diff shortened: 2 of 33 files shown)


native-provider-ci/providers/kubernetes/config.yaml
```diff
@@ -3,6 +3,7 @@ lint: true
 aws: true
 gcp: true
 major-version: 4
+pulumiCLIVersion: v3.89.0
```
t0yv0, line 6: @iwahbe just double-checking there is no provider-local .ci-mgmt.yaml in native providers yet right? So this needs to go here old-style.

t0yv0:
Directionally I love this. Do you have plans how to upgrade this version of Pulumi? Could be link a backlog item for it? 

Note that for prior art.. azure-native has a way to pin the version using makefile/local version files. I liked that a lot and we now have for bridged providers something like this:

https://github.com/pulumi/pulumi-gcp/blob/master/Makefile#L156

It's a variation on the theme that installs a local Pulumi version equal to the Go pkg dependency. It removes a degree of freedom but provides an upgrade path.

t0yv0 (approved):
CC @iwahbe who's owning builds lately (and @mjeffryes) if there's further plans around this - but nothing here we can't revert easily so LGTM.

iwahbe (approved):

native-provider-ci/providers/kubernetes/config.yaml
iwahbe: That is correct. `.ci-mgmt.yaml` is only for bridged providers right now.

PR #716, opened 2023-11-10
Add bulk merge script dependabot

iwahbe:
The long term plan is to fully automate this job. We could either:

- Put a cron job in each repo, triggering weekly to update it's own dependencies.
- Run a cron from this repo that updates dependencies in every provider.

This manual script is a stop-gap solution, but I believe it is worth checking in as is.

### DIFF
--- a/scripts/merge_dependabot.sh
+++ b/scripts/merge_dependabot.sh
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+
+# This script will combine and set to merge all dependabot PRs in bridged provider repos.
+#
+# To guard against supply side attacks, this will only merge dependabot PRs that were
+# created before last week.
+
+# Ensure iwahbe/gh-combine-prs is installed
+#
+# --force: Ensure up to date, don't error when already installed.
+gh extension install iwahbe/gh-combine-prs --force
+
+# Query syntax is https://docs.github.com/en/search-github/searching-on-github/searching-issues-and-pull-requests.
+#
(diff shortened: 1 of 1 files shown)


scripts/merge_dependabot.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# Query syntax is https://docs.github.com/en/search-github/searching-on-github/searching-issues-and-pull-requests.
+#
+# author:app/dependabot restricts this to PRs authored by dependabot.
+# created:<$(date -v-7d '+%Y-%m-%d') restricts to PRs created before 7 days ago
+QUERY="author:app/dependabot created:<$(date -v-7d '+%Y-%m-%d')"
+
+combine() {
+    provider="$1"
+    D="/tmp/pulumi-$provider"
```
mjeffryes, line 21: nit. should probably use `mktemp -d` for the temporary checkout dir

scripts/merge_dependabot.sh
```diff
@@ -0,0 +1,41 @@
...
+    rm -rf "$D"
+    (cd /tmp && git clone "https://github.com/pulumi/pulumi-$provider.git" "$D")
```
mjeffryes, line 23: more nits. I don't think you actually need to `cd /tmp` before you clone since you're using an absolute path. Also, since you already depend on the `gh client` you could use `gh repo clone` rather than interpolating the url.

mjeffryes (approved):

PR #717, opened 2023-11-13
Disable provider building in tests

VenelinMartinov:
This PR is meant to disable the new behaviour that will be added in https://github.com/pulumi/pulumi-aws/pull/2953 for CI.  

https://github.com/pulumi/pulumi-aws/pull/2953 will add provider building as a prerequisite of running tests, which is great for local development but is not what we want for CI, since it has a separate job for building the provider.

We set the env var `DISABLE_PROVIDER_BUILD` to disable the builds in CI.


t0yv0:
Would it be possible to consider not add any new complexity here but simply adjust the Go logic as appropriate? You could detect that the provider binary is in PATH by os.LookExec or you could detect that the code is in CI context by examining environment variables.

t0yv0 (changes requested):
Please note that the edit fiel

PR #718, opened 2023-11-13
Add no deps provider make target

VenelinMartinov:
This PR adds a new make target for providers `provider_no_deps` - this one will just run the go build command without `tfgen` and `install_plugins`. This is needed since `tfgen` and `install_plugins` have no caching, so will rerun each time and I'd like to use the `go build` in https://github.com/pulumi/pulumi-aws/pull/2953

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/Makefile
+++ b/provider-ci/internal/pkg/templates/bridged-provider/Makefile
@@ -119,6 +119,9 @@ install_plugins: .pulumi/bin/pulumi
 lint_provider: provider
 	cd provider && golangci-lint run -c ../.golangci.yml
 
+# `make provider_no_deps` builds the provider binary directly, without ensuring that 
+# `cmd/pulumi-resource-#{{ .Config.provider }}#/schema.json` is valid and up to date. 
+# To create a release ready binary, you should use `make provider`.
 provider_no_deps:
 	(cd provider && go build $(PULUMI_PROVIDER_BUILD_PARALLELISM) -o $(WORKING_DIR)/bin/$(PROVIDER) -ldflags "-X $(PROJECT)/$(VERSION_PATH)=$(VERSION)#{{if .Config.providerVersion}}# -X #{{ .Config.providerVersion }}#=$(VERSION)#{{end}}#" $(PROJECT)/$(PROVIDER_PATH)/cmd/$(PROVIDER))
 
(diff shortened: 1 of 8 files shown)


t0yv0:
I personally like that a lot and would use it, but adding @iwahbe as reviewer as they focused on re-designing builds recently to cross-check if this is a good idea with respect to further plans here.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -119,9 +119,11 @@ install_plugins: .pulumi/bin/pulumi
 lint_provider: provider
 	cd provider && golangci-lint run -c ../.golangci.yml
 
-provider: tfgen install_plugins
+provider_no_deps:
```
iwahbe, line 122: This is a bit of a sharp edge, so I want to make sure we have a comment explaining that it's there for.
```suggestion
# `make provider_no_deps` builds the provider binary directly, without ensuring that 
# `cmd/pulumi-resource-#{{ .Config.provider }}#/schema.json` is valid and up to date. 
# To create a release ready binary, you should use `make provider`.
provider_no_deps:
```

provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -119,9 +119,11 @@ install_plugins: .pulumi/bin/pulumi
...
 	(cd provider && go build $(PULUMI_PROVIDER_BUILD_PARALLELISM) -o $(WORKING_DIR)/bin/$(PROVIDER) -ldflags "-X $(PROJECT)/$(VERSION_PATH)=$(VERSION)#{{if .Config.providerVersion}}# -X #{{ .Config.providerVersion }}#=$(VERSION)#{{end}}#" $(PROJECT)/$(PROVIDER_PATH)/cmd/$(PROVIDER))
 
+provider: tfgen install_plugins provider_no_deps
```
iwahbe, line 125: The only reason that `provider` depends on `install_plugins` AFAIK is that `tfgen` depends on `install_plugins`. Since we already have `tfgen: install_plugins`, I think we can drop this dependency.
```suggestion
provider: tfgen provider_no_deps
```

iwahbe (approved):
It's unfortunate that we don't have a great caching solution for the schema, but we don't. If this improves the dev loop, LGTM!

I'd like to make sure we have an explanation for the extra target in the makefile when we merge. I have left a suggestion for what that would be, but you are free to expand on it.

PR #719, opened 2023-11-13
Acceptance tests concurrency group

VenelinMartinov:
This PR should add a concurrency group for all providers' acceptance test workflows. This means that only a single workflow of that type can run per PR.  

Tested here:

https://github.com/pulumi/pulumi-aws/pull/2972

https://github.com/pulumi/pulumi-aws/actions/runs/6852384520 got canceled by https://github.com/pulumi/pulumi-aws/actions/runs/6852393776


danielrbradley (approved):
This makes sense to me, would be good to get a 👍 from Anton or Ian too as I've not touched this code in a while.

guineveresaenger:
This also looks good to me. Can we verify this works in the wild against a sample provider, something small like okta or aiven?

I will say that in the past I have benefitted from multiple concurrent runs when CI was doing something odd but the overall benefit here is of course the less waste the better.

guineveresaenger (approved):

iwahbe (approved):
This change looks good to me. LGTM

PR #721, opened 2023-11-15
Devbox improvements

mjeffryes:
A few more minor devbox tweaks

- Add devbox config for ci-mgmt repo itself
- add yarn and pulumictl to the bridged provider template
- add an empty devbox.lock file to the bridged provider template so that the devcontainer will build.


provider-ci/test-workflows/aws/devbox.lock
```diff
@@ -0,0 +1 @@
+{}
```
t0yv0, line 1: This is clever. You want this floating in providers then?

t0yv0 (approved):
Let's try it.

provider-ci/test-workflows/aws/devbox.lock
mjeffryes: yes, since we're only using devbox for local development right now, I think the costs of maintaining a pin outweigh the benefits.

PR #729, opened 2023-11-22
Close outdated pulumi-bot PRs automatically

iwahbe:
pulumi-bot opens lots of PRs. Because of the number of repos we maintain, all PRs that pulumi-bot opens that we want to merge are set to auto-merge. Other PRs are opened for test purposes, but we don't intend to merge these. We want to close these PRs after a couple of days to avoid them cluttering up the repo.

This adds a script (invoked every night on a cron) that cleans up PRs by `pulumi-bot` that are not set to auto-merge and that are older then 3 days old.


t0yv0 (approved):
I love the use of Python instead of Bash.

PR #734, opened 2023-12-04
Unpin Pulumi CLI version for pulumim-kubernetes

rquitales:


native-provider-ci/providers/kubernetes/config.yaml
```diff
@@ -3,7 +3,6 @@ lint: true
 aws: true
 gcp: true
 major-version: 4
-pulumiCLIVersion: v3.89.0
```
mjeffryes, line 6: Why did we pin it before?

rquitales: This was to unblock CI due to: https://github.com/pulumi/pulumi-kubernetes/issues/2664

mjeffryes (approved):

Issue #724, opened 2023-11-20
Stateless Provider Upgrade Tracking and Automation


iwahbe:
We currently run a zapier job connection to an AWS Lambda to kickstart our automation on bridged provider upgrades.

If this job fails, nothing else happens, and we receive no notifications on new providers to update. This happened, and it took us a week to catch. To get us back on track vis a vis provider upgrades and to mitigate this issue permanently, we should transition to a less brittle path.

### New Plan
TODO

### Implementation
TODO

guineveresaenger:
Current implementation: 

https://github.com/pulumi/github-issue-automation

Instructions to add a new provider: https://github.com/pulumi/platform-providers-team/blob/main/playbooks/Onboard%20provider%3A%20internal.md

****************
The gist of the new plan I'd suggest is as follows:

- Remove the Zapier/AWS Lambda roundtrip in favor of a GitHub Action that runs on a cron schedule, asking the GitHub API for a latest release for any given provider. Basically, this changes the model from a push-based approach (which has several points of failure) to a poll-based approach (which, even if it fails, can be idempotently rerun and "recover" missed history).

- https://github.com/pulumi/github-issue-automation/blob/main/lambda/new-release-handler/main.go includes some logic around what we consider a valid release, which should be preserved. Additionally, we should not need to keep a map of terraform providers - this information is available, in-repo, as part of each provider's CI config.

- Using "the existence of an issue that has `upgrade terraform-provider-foo` in its title" as a trigger for upgrades should be removed, although this may be a later refinement.

A few details:

- The cron GHA will create an upgrade issue, and ideally it should immediately trigger an upgrade build, linking this build to the issue created.

- It seems as though the best approach here is to alter https://github.com/pulumi/upgrade-provider to have an option to query the upstream latest release, and use that inside the cron Action. Later, we can refactor and remove the need to rely on detecting open upgrade issues on the Pulumi provider.

guineveresaenger:
> To get idempotency, I think we will need to have detecting open issues depend on the repo in some way, since we need to understand the current version to know if we need to upgrade.

I find this whole process of polling existing issues to be disturbingly brittle, but realized with chagrin that with the git submodule approach, we can't just read the upstream version from go.mod. 🤔 
Maybe write the upstream version to `.ci-mgmt.yaml`? Anyway, this is not immediately necessary, just something to ponder.

iwahbe:
[...]

PR #736, opened 2023-12-05
Add workflow template for cron job checking for upstream upgrades

guineveresaenger:
This PR adds a workflow that runs once a day on a schedule and plans to use the changes in pulumi/upgrade-provider#207 to poll upstream for upgrades, resulting in the creation of a "Upgrade terraform-provider-foo to vx.y.z" issue.

To assist during feature rollout, this PR targets 10 Tier 2 providers (incidentally these are all from a single Zapier handler that we plan to disable for careful rollout) to receive these new upgrade workflows. 

To look at a few example trials of this Workflow, [follow the links in this comment](https://github.com/pulumi/ci-mgmt/issues/724#issuecomment-1841288902)

Part of #724.


provider-ci/internal/pkg/templates/bridged-provider.config.yaml
```diff
@@ -8,6 +8,7 @@ fail-on-missing-mapping: true
 fail-on-extra-mapping: true
 publishRegistry: true
 checkoutSubmodules: false
+use-upstream-version-check: false # TODO: this option should be removed once all providers are onboarded to upstream version checks.
```
t0yv0, line 11: Is this a new option?

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/check-upstream-upgrade.yml
```diff
@@ -0,0 +1,54 @@
...
+    - name: "Set up git identity: name"
+      run: git config --global user.name pulumi-bot
+      shell: bash
+    - name: "Set up git identity: email"
+      run: git config --global user.email bot@pulumi.com
+      shell: bash
+    - name: Run upgrade-provider upstream check
+      id: upstream_version
+      run: |
+        upgrade-provider "$REPO" --kind=check-upstream-version
```
t0yv0, line 33: I see.

t0yv0 (approved):

t0yv0:
🚢

provider-ci/internal/pkg/templates/bridged-provider.config.yaml
guineveresaenger: oof, thank you for catching that! I changed my mind and we won't be using this option, thanks to passing in a list of repos to the template [here](https://github.com/pulumi/ci-mgmt/pull/736/files#diff-aa352508a9683780662f0243439a9b38fc95a4f24455a42e87ae994f01152ce4R1)

Issue #506, opened 2023-08-04
Get Java up-to-date


t0yv0:
Continuing from https://github.com/pulumi/ci-mgmt/pull/504#issuecomment-1665676166

Need to make sure pulumi-java is up-to-date with the latest version and the SDKs regenerated.

If doing manually see checklist below.

For automation options consider prior art in here https://github.com/pulumi/pulumi-azure-native/blob/master/Makefile#L181-L189

Note also there is a nightly job https://github.com/pulumi/ci-mgmt/blob/master/.github/workflows/update-workflows-ecosystem-providers.yml#L10 that syncs providers to latest ci-mgmt with an automated PR. If this job could take care of re-generating the sdk/java files ("build_sdk") then much manual work could be avoided here in bringing these providers up-to-date.

As another alternative perhaps update-provider tool could bring java versions up-to-date. If we wen that direction we'd need to decouple the java version a provider uses from ci-mgmt first; that is it should be specified in every individual repo not in ci-mgmt repo. In that case update-provider could manage it.

## Native Providers

- [x] command

`pulumi-command` uses the same version of java a `pulumi` does:

https://github.com/pulumi/pulumi-command/blob/7c5c4bd68e66950c042c5ce1ad3eb6f9bfda835c/Makefile#L88-L97

## Bridged Providers

Note: The list of bridged providers is not up to date.

- [ ] aiven
- [x] akamai
- [ ] alicloud
- [ ] artifactory
- [ ] auth0
- [ ] azure
- [ ] azuread
- [ ] civo
- [ ] cloudamqp
- [ ] cloudflare
- [ ] cloudinit
- [ ] confluentcloud
- [ ] consul
- [ ] databricks
- [ ] datadog
- [ ] digitalocean
- [ ] dnsimple
- [x] docker
- [x] ec
- [ ] equinix-metal
- [ ] f5bigip
- [ ] fastly
- [ ] github
- [ ] gitlab
- [x] hcloud
- [x] kafka
- [x] keycloak
- [ ] kong
- [ ] libvirt
- [ ] linode
- [ ] mailgun
- [ ] minio
- [ ] mongodbatlas
- [ ] mysql
- [ ] newrelic
- [ ] nomad
- [ ] ns1
- [ ] oci
- [ ] okta
- [ ] onelogin
- [ ] openstack
- [ ] opsgenie
- [ ] pagerduty
- [ ] postgresql
- [ ] rabbitmq
- [ ] rancher2
- [ ] random
[...]

PR #737, opened 2023-12-06
Always build the `bin/pulumi-java-gen` target

guineveresaenger:
Running `make build_sdks` locally often fails on java-gen if there is an old java gen binary left in the (not source controlled) provider/bin folder.
Making the binary build a phony target ensures we re-build it every time an pull the java gen binary of the version specified in the provider config and Makefile.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/Makefile
+++ b/provider-ci/internal/pkg/templates/bridged-provider/Makefile
@@ -185,4 +185,4 @@ ci-mgmt: .ci-mgmt.yaml
 	@mkdir -p .pulumi
 	@cd provider && go list -f "{{slice .Version 1}}" -m github.com/pulumi/pulumi/pkg/v3 | tee ../$@
 
-.PHONY: development build build_sdks install_go_sdk install_java_sdk install_python_sdk install_sdks only_build build_dotnet build_go build_java build_nodejs build_python clean cleanup#{{ if .Config.docsCmd }}# docs#{{end}}# help install_dotnet_sdk install_nodejs_sdk install_plugins lint_provider provider provider_no_deps test tfgen upstream upstream.finalize upstream.rebase ci-mgmt test_provider
+.PHONY: development build build_sdks install_go_sdk install_java_sdk install_python_sdk install_sdks only_build build_dotnet build_go build_java build_nodejs build_python clean cleanup#{{ if .Config.docsCmd }}# docs#{{end}}# help install_dotnet_sdk install_nodejs_sdk install_plugins lint_provider provider provider_no_deps test tfgen upstream upstream.finalize upstream.rebase ci-mgmt test_provider bin/pulumi-java-gen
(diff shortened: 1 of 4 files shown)


iwahbe:
We had code that did this by inspecting the downloaded binary directly in pulumi-aws. It was introduced [here](https://github.com/pulumi/pulumi-aws/pull/2618), but removed when we normalized pulumi-aws.

An even simpler solution might be to postfix the java path by its version. Instead of downloading to a binary called `bin/pulumi-java-gen`, download to a binary called `bin/pulumi-java-gen-$(JAVA_GEN_VERSION)`.

thomas11:
> We had code that did this by inspecting the downloaded binary directly in pulumi-aws. It was introduced [here](https://github.com/pulumi/pulumi-aws/pull/2618), but removed when we normalized pulumi-aws.
> 
> An even simpler solution might be to postfix the java path by its version. Instead of downloading to a binary called `bin/pulumi-java-gen`, download to a binary called `bin/pulumi-java-gen-$(JAVA_GEN_VERSION)`.

Azure Native uses a sentinel file `.pulumi-java-gen.version` containing the version. It's used in this Make target:
```make
bin/pulumi-java-gen: .pulumi-java-gen.version bin/pulumictl
	@mkdir -p bin
	bin/pulumictl download-binary -n pulumi-language-java -v $(shell cat .pulumi-java-gen.version) -r pulumi/pulumi-java
```

Make considers the target out of date and re-downloads pulumi-java **only if** the timestamp of the sentinel file is newer, i.e., someone edited it to bump the version.

danielrbradley:
> Can you talk about this a bit more?

Sure thing! As part of our quality and local development experience improvements we want to continue to iterate on our makefiles to model dependencies as accurately as possible. @VenelinMartinov is also looking at this - to get to the point where the whole provider build can be marked as up-to-date. 

The starting point of these changes is by fixing up the targets at the bottom of the dependency tree on which everthing else depends because, if these are marked as phony, then anything that depends on them will also never be able to be up to date.

> An even simpler solution might be to postfix the java path by its version. Instead of downloading to a binary called `bin/pulumi-java-gen`, download to a binary called `bin/pulumi-java-gen-$(JAVA_GEN_VERSION)`.

This is an interesting idea - I'd be happy with either approach - writing the version to a separate file, or putting the version in the filename would both work.

Putting the version into the filename of the binary saves having another file in the repository, but I'm not sure if changing the name of the binary is something that `pulumictl download-binary` supports yet. The other slight downside is that it's slightly more messy code to always reference `bin/pulumi-java-gen-$(JAVA_GEN_VERSION)` rather than just `bin/pulumi-java-gen`, but that's a minor nit pick.

I've just noticed that we've already got the "version in a file" approach used for downloading the pulumi CLI which depends on `.pulumi/version`. So, to follow that pattern I'd probably lean towards putting the version number into a file at `.pulumi/java-gen-version`.

guineveresaenger:
Closing as this is handled via #506 and #747 .

PR #739, opened 2023-12-08
Update schema tools to use pulumi/schema-tools in native providers

rquitales:


native-provider-ci/src/steps.ts
```diff
@@ -543,7 +543,7 @@ export function CheckSchemaChanges(provider: string): Step {
     name: "Check Schema is Valid",
     run:
       "echo 'SCHEMA_CHANGES<<EOF' >> $GITHUB_ENV\n" +
-      "schema-tools compare ${{ env.PROVIDER }} master --local-path=provider/cmd/pulumi-resource-${{ env.PROVIDER }}/schema.json >> $GITHUB_ENV\n" +
```
rquitales, line 546: New schema-tools CLI command format. Obtained from bridged providers config, and verified locally.

PR #745, opened 2023-12-11
Fix KeyError in close_outdated_bridge_prs.py

mjeffryes:
Fixes https://github.com/pulumi/ci-mgmt/issues/743

In #730, @iwahbe added logic  to avoid closing PRs that have a label on them b/c this probably indicates a human is working on that PR.

Unfortunately, python throws a `KeyError` rather than returning `None` when it can't find a key in a dict, so this broke ci for issues that don't have labels.

This change should fix the code to match the intent.


iwahbe:
I've been writing too much go code lately. Thanks for the fix.

PR #755, opened 2023-12-21
Fix setup pulumi pin

t0yv0:
Noticing that AWS was testing against outdated CLI v3.77.1. This is unfortunate. 

This change gets it back to floating as all the other providers. All the other providers are updated to use the latest recommended GitHub action.

I have slight preference for pinned and weekly updating but that's more setup, for now floating is better than pinned indefinitely in the past.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider.config.yaml
+++ b/provider-ci/internal/pkg/templates/bridged-provider.config.yaml
@@ -37,7 +37,9 @@ env:
 actions:
   setupPulumi:
     - name: Install Pulumi CLI
-      uses: pulumi/action-install-pulumi-cli@v2
+      uses: pulumi/actions@v4
+      with:
+        pulumi-version: "^3"
   setupGo:
     - name: Install Go
       uses: actions/setup-go@v4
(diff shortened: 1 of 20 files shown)


t0yv0:
Should this be removed from .ci-mgmt.yaml in pulumi-aws also? 

https://github.com/pulumi/pulumi-aws/blob/master/.ci-mgmt.yaml#L35

VenelinMartinov (approved):

iwahbe:
> Should this be removed from .ci-mgmt.yaml in pulumi-aws also?
> 
> https://github.com/pulumi/pulumi-aws/blob/master/.ci-mgmt.yaml#L35

You will need to remove ☝️ for the change to take effect. The AWS config here is only used as an example of how the changes we make manifest downstream. It has no effect on the actual provider.

t0yv0:
I've just added a warning headers to these files so it's obvious they're not actually configuring the providers. I think this is good to go then, let me test the change on AWS and if it succeeds I'll merge here.

t0yv0:
https://github.com/pulumi/pulumi-aws/pull/3184 this works in principle.

PR #757, opened 2023-12-27
Add deploy release labels script

iwahbe:
I used this script to systemically set up `needs-release/*` labels in all our providers. I'm checking it in to avoid duplication of work for future label changes.

### DIFF
--- a/scripts/ensure_release_labels.py
+++ b/scripts/ensure_release_labels.py
@@ -1,0 +1,37 @@
+#!/usr/bin/env python3
+
+# This script creates the needs-release/{major,minor,patch} labels necessary for our
+# release process on all provider repos.
+#
+# The script is idempotent.
+
+import subprocess
+import json
+
+
+def cmd(*args: str, prefix="Exec: ") -> str:
+    print(prefix + " ".join(args))
+    return subprocess.run(args,
+                   capture_output=True,
+                   check=True,
+                   encoding="utf-8").stdout
+
+def set_labels(repo: str):
+
+    label = lambda kind, desc: cmd("gh", "label", "create", f"needs-release/{kind}",
+                     "--repo", repo,
+                     "--color", "#C5DEF5",
+                     "--force", # Update if another color; don't fail if no change
+                     "--description", f"When a PR with this label merges, it initiates a release of {desc}",
+                     )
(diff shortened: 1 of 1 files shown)


guineveresaenger (approved):

scripts/ensure_release_labels.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python3
```
t0yv0, line 1: Thank you for using Python instead of bash. We don't have a policy but I'd very much rather we committed to Go or Python for scripting as it requires so much less training to avoid corner case mistakes.

scripts/ensure_release_labels.py
```diff
@@ -0,0 +1,37 @@
...
+#
+# The script is idempotent.
+
+import subprocess
+import json
+
+
+def cmd(*args: str, prefix="Exec: ") -> str:
+    print(prefix + " ".join(args))
+    return subprocess.run(args,
```
t0yv0, line 14: https://docs.python.org/3/library/subprocess.html#subprocess.check_output is my goto in these cases, though you may be specific here because you want to ignore stderr?

scripts/ensure_release_labels.py
```diff
@@ -0,0 +1,37 @@
...
+                   capture_output=True,
+                   check=True,
+                   encoding="utf-8").stdout
+
+def set_labels(repo: str):
+
+    label = lambda kind, desc: cmd("gh", "label", "create", f"needs-release/{kind}",
```
t0yv0, line 21: I'm rusty on Python style but I think this nested def form is preferred:

```
def label(kind, desc):
    return cmd(...)
```

t0yv0 (approved):

scripts/ensure_release_labels.py
guineveresaenger: strong agree; training aside, it just speeds up everyone.

scripts/ensure_release_labels.py
```diff
@@ -0,0 +1,37 @@
...
+                     "--repo", repo,
+                     "--color", "#C5DEF5",
```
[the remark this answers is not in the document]
iwahbe: It's the same color as the `needs-release/*` labels that already existed.

PR #759, opened 2023-12-27
Merge dependabot cron

iwahbe:
Call `./scripts/merge_dependabot.sh` every 4 days.

### DIFF
--- a/.github/workflows/merge-dependabot.yml
+++ b/.github/workflows/merge-dependabot.yml
@@ -1,0 +1,17 @@
+name: merge-dependabot
+
+on:
+  workflow_dispatch: {}
+  schedule:
+    # 4:00AM every 4 days.
+    - cron:  '0 4 01/4 * *'
+env:
+  GITHUB_TOKEN: ${{ secrets.PULUMI_BOT_TOKEN }}
+
+jobs:
+  update_workflows:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone ci-mgmt
+        uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/merge-dependabot.yml
```diff
@@ -0,0 +1,17 @@
+name: merge-dependabot
+
+on:
+  workflow_dispatch:
```
thomas11, line 4: We use `workflow_dispatch: {}` everywhere else that I can see from a quick grep (unless there are actual arguments). Is the `{}` not necessary?

iwahbe: I'm not 100% sure. I added the `{}` to be like other workflows.

.github/workflows/merge-dependabot.yml
```diff
@@ -0,0 +1,17 @@
+name: merge-dependabot
+
+on:
+  workflow_dispatch: {}
+  schedule:
+    # 4:00AM every 4 days.
+    - cron:  '0 4 01/4 * *'
```
guineveresaenger, line 7: TIL you can't say "every fourth day" in cron. 
Another option would be to run every Monday and Thursday, which is slightly more frequent but less crowded around the first of the month? Either way is fine though.

guineveresaenger (approved):
LGTM

thomas11 (approved):

.github/workflows/merge-dependabot.yml
iwahbe: > TIL you can't say "every fourth day" in cron.

Maybe you can, but I don't know how.

PR #764, opened 2023-12-28
Upgrade publish workflow to v0.0.13

iwahbe:
This allows each repo to opt in or out of publishing different SDKs.

This is necessary for pulumi-local, since we are unable to publish the `pulumi-local` python SDK.


iwahbe:
An org-wide search shows that `publishProviderSDKs` was only set in this repo. That makes it safe to move.

VenelinMartinov (approved):

Issue #751, opened 2023-12-19
PyPI is deprecating our current auth method on 1st Jan 2024


VenelinMartinov:
Noticed an error when releasing aws-apigateway: https://github.com/pulumi/pulumi-aws-apigateway/actions/runs/7260038801/job/19779026884#step:13:47

Looks like PyPI is deprecating password auth. It suggests we should switch to "Trusted Publishers" or auth tokens:
```
Notice: Using a username + password pair for authentication against https://upload.pypi.org/legacy/
Error: Starting in 2024, PyPI will require all users to enable Two-Factor Authentication. This will consequently require all users to switch to either Trusted Publishers (preferred) or API tokens for package uploads. Read more: https://blog.pypi.org/posts/2023-05-25-securing-pypi-with-2fa/
Warning: Trusted Publishers allows publishing packages to PyPI from automated environments like GitHub Actions without needing to use username/password combinations or API tokens to authenticate with PyPI. Read more: https://docs.pypi.org/trusted-publishers
```
Steps
- [x] Create API token & add to GH globally
- [ ] Update ci-mgmt based providers
- [ ] Update each non-managed provider
- [ ] Enable 2FA on the pypi account

## Native Providers

- [ ] aws-apigateway
- [ ] aws-native
- [ ] awsx
- [ ] azure-native
- [ ] command
- [ ] eks
- [ ] google-native
- [ ] kubernetes

### Tier 1

- [ ] aws
- [ ] gcp
- [ ] azure
- [ ] azuread

### Tier 2

- [ ] aiven
- [ ] akamai
- [ ] alicloud
- [ ] archive
- [ ] artifactory
- [ ] auth0
- [ ] azuredevops
- [ ] civo
- [ ] cloudamqp
- [ ] cloudflare
- [ ] cloudinit
- [ ] confluentcloud
- [ ] consul
- [ ] databricks
- [ ] datadog
- [ ] digitalocean
- [ ] dnsimple
- [ ] docker
- [ ] ec
- [ ] external
- [ ] f5bigip
- [ ] fastly
- [ ] github
- [ ] gitlab
- [ ] hcloud
- [ ] http
- [ ] kafka
- [ ] keycloak
- [ ] kong
- [ ] libvirt
- [ ] linode
- [ ] local
- [ ] mailgun
- [ ] minio
- [ ] mongodbatlas
- [ ] mysql
- [ ] newrelic
- [ ] nomad
- [ ] ns1
- [ ] null
[...]

PR #767, opened 2023-12-28
PyPi switch to API Token

iwahbe:
According to [pypi](https://pypi.org/help/):

```
To use an API token:
* Set your username to __token__
* Set your password to the token value, including the pypi- prefix
```

Fixes https://github.com/pulumi/ci-mgmt/issues/751

I tested this end to end in pulumi-mysql: https://github.com/pulumi/pulumi-mysql/actions/runs/7350064841/job/20011366694.


guineveresaenger (approved):
LGTM.

...we probably don't need _all_ of these Workflows to have the PyPI token but it doesn't hurt and this must merge.

Also, given the successful sample run, presumably this token is available in GH Secrets everywhere?

iwahbe:
> ...we probably don't need all of these Workflows to have the PyPI token but it doesn't hurt and this must merge.

We don't, but that is the current state of the world and I'd rather not try to tighten it up right now.

> Also, given the successful sample run, presumably this token is available in GH Secrets everywhere?

The token is an org level secret.

PR #769, opened 2024-01-03
Error on missing docs during tfgen

VenelinMartinov:
We run this during upgrade workflows so they should run during normal CI.

Should prevent errors like https://github.com/pulumi/pulumi-keycloak/issues/355


iwahbe (approved):
Looks good to me

PR #770, opened 2024-01-03
Option to set upstream provider org in upgrade config

guineveresaenger:
Part of https://github.com/pulumi/upgrade-provider/issues/221 - refer to issue for details.

Explicitly telling our upgrade machinery the upstream org name allows us to reliably detect upstream upgrades every time.


iwahbe (approved):
LGTM

PR #773, opened 2024-01-04
Require PRs and don't allow direct pushes from admins

blampe:
Prompted by https://pulumi.slack.com/archives/C02FXTZEZ6W/p1704391521664269?thread_ts=1702980308.180469&cid=C02FXTZEZ6W

GitHub allowed me to push directly to a `master` branch because I'm an admin.

I manually disabled that setting on a few repos before realizing we're managing these branch protections from ci-mgmnt.

This PR prevents GitHub admins from pushing directly to `master` by requiring all changes to have a PR.


guineveresaenger (approved):

guineveresaenger (changes requested):
Sorry I take it back.

Would requiring a review not interfere with update automation?

infra/providers/index.ts
```diff
@@ -49,6 +48,11 @@ function defineResources(buildSdkJobName: string, provider: string) {
         new github.BranchProtection(`${provider}-${branch}-branchprotection`, {
             repositoryId: `pulumi-${provider}`,
             pattern: `${branch}`,
+            enforceAdmins: true,
+            requiredPullRequestReviews: [{
+                pullRequestBypassers: ["pulumi-bot"],
+                requiredApprovingReviewCount: 1,
```
guineveresaenger, line 54: Note that this is the default so we don't need to set it explicitly.

I dug a little deeper into this and it appears we _do_ have this very branch protection set up on pulumi-aws, and it means that every time we upgrade the GitHub Workflows [a manual approval is required](https://github.com/pulumi/pulumi-aws/pull/3215).

If we set this globally, it will explode across providers, unless there's some overrides that I'm unaware of.

guineveresaenger (changes requested):
I really think we have to think carefully about adding required PR + review globally.

There is a way to _unset_ the required review to null: https://docs.github.com/en/rest/branches/branch-protection?apiVersion=2022-11-28 but keep having a pull request as a requirement.

Doing so would make certain repos such as pulumi-aws more permissive, however.

infra/providers/index.ts
iwahbe: We have it only on AWS, and it makes everything manual. We cannot set it globally.

guineveresaenger: (I assume you mean for practicality reasons, @iwahbe) - a change that would mostly preserve the current status quo but additionally require a PR for merge to protected branches, is setting `requiredApprovingReviewCount` to none, and special case AWS?

infra/providers/index.ts
```diff
@@ -49,6 +48,11 @@ function defineResources(buildSdkJobName: string, provider: string) {
         new github.BranchProtection(`${provider}-${branch}-branchprotection`, {
             repositoryId: `pulumi-${provider}`,
             pattern: `${branch}`,
+            enforceAdmins: true,
+            requiredPullRequestReviews: [{
+                pullRequestBypassers: ["pulumi-bot"],
```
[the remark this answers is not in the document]
blampe: Great catch! Extremely unexpected.

infra/providers/index.ts
```diff
@@ -50,8 +50,9 @@ function defineResources(buildSdkJobName: string, provider: string) {
             pattern: `${branch}`,
             enforceAdmins: true,
             requiredPullRequestReviews: [{
-                pullRequestBypassers: ["pulumi-bot"],
-                requiredApprovingReviewCount: 1,
+                pullRequestBypassers: ["/pulumi-bot"],
+                // pulumi-bot sometimes needs to be able to merge without approval.
+                requiredApprovingReviewCount: undefined,
```
guineveresaenger, line 55: I think we still need to special case AWS so we don't get unwanted auto merges.

blampe (commented):
Thanks for the sanity check @guineveresaenger! I setup a test repo just to confirm the behavior and this is exactly right:

> "allowed to bypass pull request requirements" means that it allows pulumi-bot to not make a pull request; not that any pull request review requirements do not apply.

I've updated this to explicitly not set the required number of approvals, with a note explaining why. The net result, as far as I can tell, is that `pulumi-bot` will continue to be able to merge PRs without approval, and admins won't be able to directly push to main. We keep aws as an exception.

While I'm in here I'm going to setup deployments for this stack.

infra/providers/index.ts
iwahbe: Yes. It would be prohibitively expensive to manually approve all automated PRs. If we could set `requiredApprovingReviewCount: 1` and except automated PRs, that would be great.

guineveresaenger (approved):
The logic looks good to me, thank you!

If you're setting up Deployments, can we add some information (here or otherwise) how this stack gets deployed? For folks who don't use Deployments on the daily this may be a nice to have.

PR #774, opened 2024-01-04
Add minio to denylist for upstream checks

guineveresaenger:
Due to https://github.com/pulumi/pulumi-minio/issues/200 we are currently uable to upgrade the Pulumi MinIO provider.

We should not continue to check for upstream updates that we cannot take.

Part of https://github.com/pulumi/upgrade-provider/issues/221.


iwahbe (approved):
LGTM

P.S. If we have any more PRs of this format, I'm going to suggest that we move this list out of templates into `.ci-mgmt.yml`: `disableUpstreamCheck` or something similar.

guineveresaenger:
for sure! I think I'm happy to do that as a cleanup/refactor, after we've moved out Tier 1 from this code.

PR #775, opened 2024-01-11
Parameterize upgrade-bridge job kind

dixler:
Part of https://github.com/pulumi/pulumi/issues/15140

This parameterizes the kind run on upgrade-provider to enable downstream codegen tests to only upgrade pulumi (they currently may upgrade the bridge which could cause the bridge to block pulumi/pulumi releases).

We could copy the job as well to an `upgrade-pulumi.yml`, but I'm unfamiliar with the repository so I'm unsure of the consequences of such an approach.


VenelinMartinov (approved):
The name of the action is kind of unfortunate but the change seems fine to me.

t0yv0:
this looks fine. Can always revert if doesn't work.

PR #778, opened 2024-01-12
Upgrade all pu/pu dependencies

blampe:
Our automation around pu/pu bumps doesn't catch every directory, e.g. `examples/go.mod` is often stale.

This is a source of dependabot noise. Updating pu/pu often has a side-effect of resolving security issues (since pu/pu is constantly bumping minimum revs). For the directories that don't bump pu/pu we end up needing followup from dependabot.

Re: special casing for p/kubernetes, AFAICT there's no behavioral difference with what we have right now. p/kubernetes skips the `go mod download` step, but `go mod tidy` does this automatically; it also runs `make ensure` at the end, but that just runs `go mod tidy`.

### DIFF
--- a/native-provider-ci/providers/aws-native/repo/.github/workflows/weekly-pulumi-update.yml
+++ b/native-provider-ci/providers/aws-native/repo/.github/workflows/weekly-pulumi-update.yml
@@ -74,25 +74,7 @@ jobs:
 
         git checkout -b update-pulumi/${{ github.run_id }}-${{ github.run_number }}
 
-        cd provider
-
-        go get github.com/pulumi/pulumi/pkg/v3
-
-        go get github.com/pulumi/pulumi/sdk/v3
-
-        go mod download
-
-        go mod tidy
-
-        cd ../sdk
-
-        go get github.com/pulumi/pulumi/sdk/v3
-
-        go mod download
-
-        go mod tidy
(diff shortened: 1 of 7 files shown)


native-provider-ci/providers/aws-native/repo/.github/workflows/weekly-pulumi-update.yml
```diff
@@ -74,25 +74,7 @@ jobs:
...
-        cd ../sdk
-
-        go get github.com/pulumi/pulumi/sdk/v3
-
-        go mod download
-
-        go mod tidy
-
-        cd ..
+        for MODFILE in $(find . -name go.mod); do pushd $(dirname $MODFILE); go get github.com/pulumi/pulumi/pkg/v3 github.com/pulumi/pulumi/sdk/v3; go mod tidy; popd; done
```
VenelinMartinov, line 77: Would this add the dependencies if they weren't there before? Looks like we used to only update pulumi/sdk for the sdk directory.

blampe: Yes, however the `go mod tidy` step removes any unused dependencies.

VenelinMartinov (approved):

PR #782, opened 2024-01-16
Fix upgrade-bridge job

t0yv0:
After https://github.com/pulumi/ci-mgmt/pull/775 the downstream checks are no longer working for testing the bridge because it invokes upgrade-provider with `kind=""`. Note that `inputs` are not defined when processing `repository_dispatch`, so the code has to work around that.


iwahbe (approved):
LGTM

PR #783, opened 2024-01-16
Enable upstream upgrade checks on GCP, AWS, and Azure. Update AWS test files.

guineveresaenger:
These are the final three providers for https://github.com/pulumi/ci-mgmt/issues/724.
After this merges, we will remove the Zapier job and retire https://github.com/pulumi/github-issue-automation.


iwahbe (approved):
LGTM! 🎉

PR #784, opened 2024-01-17
Update provider onboarding instructions

guineveresaenger:
Updates the instructions for adding a new bridged provider to our internal automation ecosystem.

This pull request falls short of adding full documentation for possible config settings, which is more involved and needs a follow-up, but it corrects out of date information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,20 +59,11 @@ Adjust for your provider and filesystem.
 
 To add a new provider:
 
-1. Create a new directory and config file for the provider. From the root of the repository, run:
-
-   ```bash
-   # Change the value of PROVIDER_NAME below:
-   PROVIDER_NAME=foo && \
-      mkdir provider-ci/providers/${PROVIDER_NAME} && \
-      touch provider-ci/providers/${PROVIDER_NAME}/config.yaml
-   ```
-
-1. In the `config.yaml` you created, add the configuration to be applied to the generated GitHub Actions workflows for the provider:
+1. A new provider needs a top-level `.ci-mgmt.yaml` in its _own_ repository with the following basic configuration:
 
    ```yaml
    # Required values:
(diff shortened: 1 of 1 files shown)


guineveresaenger:
Followup from https://github.com/pulumi/platform-providers-team/pull/135.

README.md
```diff
@@ -84,28 +75,33 @@ To add a new provider:
    setup-script: testing/setup.sh # Path to a script that's used for testing bootstraps
    ```
 
-1. Generate the configuration:
+1. In _this_ repository (`ci-mgmt`), generate the configuration:
```
iwahbe, line 78: What is the point of this step?

guineveresaenger: To see if you're set up correctly? It's certainly optional, compared to before, but there's value in having the files immediately, and verifying you're g2g. I can add some more words to that effect.

README.md
```diff
@@ -74,28 +74,27 @@ To add a new provider:
...
 
+1. If you would like to manually generate the configuration for verification, you can do so in this repository.
+
+   ```bash
+   cd provider-ci && make provider NAME=foo
+   ```
+
+   The generated files will be writen to `provider-ci/providers/foo/repo/`.
+   These files will not be committed to Git.
+
```
iwahbe, line 97: I would instead suggest running:
```bash
go run github.com/pulumi/ci-mgmt/provider-ci@master generate \
	--name pulumi/pulumi-$(PROVIDER_NAME) \
	--out . \
	--template bridged-provider \
	--config .ci-mgmt.yaml
```

This will dump your config files into the repo you are currently in (letting you get started without waiting for a nightly PR). 

The above script is taken from the Makefile we generate for each provider.

iwahbe (approved):
LGTM.

PR #723, opened 2023-11-18
Test PRs targeting branches other than master

t0yv0:
In context of https://github.com/pulumi/pulumi-aws/pull/3025 it would be great to be able to maintain multiple "main" branches, such as "v5" branch in AWS, and have tests qualify this branch and PRs targeting it. This change specifically enables testing PRs that target feature-* or v* branches.

PR #785, opened 2024-01-18
Run acceptance tests on all branch targets

blampe:
I found it surprising when trying to [stack](https://stacking.dev) some PRs that we don't run tests unless the PR is explicitly against master. This is [awkward](https://github.com/pulumi/pulumi-docker/pull/948#issuecomment-1897531359) to work with.

This PR changes our workflow to trigger for PRs against any branch. I might be missing some context around #723 and why we chose to opt-in only some patterns, but if there are automated flows that shouldn't run tests I recommend we opt those out explicitly.


t0yv0:
Yeah without this change you could work around by calling your branch feature-x . Maybe the thinking here was that the tests are expensive or some such. But I also find it a little surprising so your change LGTM.

t0yv0 (approved):

PR #788, opened 2024-01-22
Add gomod Make dependency

t0yv0:
The Makefile uses a trick to get a Pulumi CLI version that matches the provider/go.mod dependency on pkg. However since we did not explain to Make that reading go.mod is literally a dependency here, @guineveresaenger machine got into a state where `make` locally was using an out-of-date Pulumi CLI compared to CI. This is probably benign just yet, but with turning on example conversion via Pulumi CLI we can easily get this to produce discrepancies in generated examples between local and CI.

### DIFF
--- a/provider-ci/Makefile
+++ b/provider-ci/Makefile
@@ -56,6 +56,10 @@ lint-providers/%/repo: $(ACTIONLINT) providers/%/repo providers/%/config.yaml
 	# directory it either starts checking ci-mgmt workflows itself or checks workflows
 	# file-by-file and incorrectly reports dangling cross-references.
 	cd providers/$*/repo &&	git init && ../../../$(ACTIONLINT) -config-file ../../../actionlint.yml && rm -rf .git
+	# Allow Makefile to reference go.mod.
+	mkdir -p providers/$*/repo/provider && touch providers/$*/repo/provider/go.mod
+	# Assume .pulumi/version exists
+	mkdir -p providers/$*/repo/.pulumi && touch providers/$*/repo/.pulumi/version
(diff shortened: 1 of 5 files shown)


t0yv0:
Ah this reminds me why I did not put it in in the first place :) This defeats the spellcheck.

t0yv0:
With a bit of a workaround, I can get the CI to pass. Let me know if this looks OK.

provider-ci/Makefile
```diff
@@ -56,6 +56,10 @@ lint-providers/%/repo: $(ACTIONLINT) providers/%/repo providers/%/config.yaml
 	# directory it either starts checking ci-mgmt workflows itself or checks workflows
 	# file-by-file and incorrectly reports dangling cross-references.
 	cd providers/$*/repo &&	git init && ../../../$(ACTIONLINT) -config-file ../../../actionlint.yml && rm -rf .git
+	# Allow Makefile to reference go.mod.
+	mkdir -p providers/$*/repo/provider && touch providers/$*/repo/provider/go.mod
+	# Assume .pulumi/version exists
+	mkdir -p providers/$*/repo/.pulumi && touch providers/$*/repo/.pulumi/version
```
iwahbe, line 62: I'm not sure why we need this line, since we didn't need it before. If it's needed to make the linter happy, so be it.

iwahbe (approved):
This looks fine to me. Catching corner cases with better error messages is always a good thing.

provider-ci/Makefile
t0yv0: Yes exactly. This is to make linter happy locally for me.

PR #795, opened 2024-01-30
Disable AWS Native

danielrbradley:
We're needing to diverge from the generic pattern due to test restructuring so are off-boarding from ci-mgmt for the time being.

Connected to https://github.com/pulumi/pulumi-aws-native/pull/1292

### DIFF
--- a/scripts/generate_native_providers_list.py
+++ b/scripts/generate_native_providers_list.py
@@ -3,7 +3,7 @@ import os
 import json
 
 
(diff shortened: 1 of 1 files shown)


blampe (approved):
Let's file an issue somewhere to track re-convergence (or whatever else we have in mind for managing these going forward).

danielrbradley:
Coverting back to draft as I've managed to avoid needing to do this for the time being, though will likely look at doing this again very soon.

PR #799, opened 2024-01-31
Update gradle-build-action version for native providers

mjeffryes:
As noted in https://github.com/pulumi/pulumi-kubernetes/pull/2795#issuecomment-1919552728, we need to bump the version used here for repositories that are managed under ci-mgmt/native-provider-ci

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -41,6 +41,8 @@ jobs:
         with:
           path: ci-scripts
           repository: pulumi/scripts
+      - name: Validate native-providers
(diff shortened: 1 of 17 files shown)


rquitales (approved):

blampe (approved):

blampe:
@mjeffryes don't forget to `make`!

mjeffryes:
> @mjeffryes don't forget to `make`!

Doh! Looks like we don't even build the native-providers in the PR checks or that would have caught this.

Ran make and also updated the workflow

PR #807, opened 2024-02-07
Show which patch failed

t0yv0:
When calling make upstream that fails on applying a patch, I would love too see immediately which patch failed. This helps fix that one.


guineveresaenger (approved):
Sounds like a useful enhancement to me!

PR #812, opened 2024-02-09
Pin pulumi/scripts checkout

t0yv0:
This is a slow-moving legacy dependency, I'm hoping that pinning it can help actions/checkout to be better at caching or not failing around exhausting GH tokens as in:

FAILED https://github.com/pulumi/pulumi-azuredevops/pull/293
FAILED https://github.com/pulumi/pulumi-azuredevops/pull/292


t0yv0:
Not sure. I'm hoping having a ref will activate caching of some sort. Generally I prefer pin everything to make builds reproducible.

guineveresaenger (approved):
Seems fine - I bow to your experience/hunch

PR #814, opened 2024-02-09
Pin pulumictl

t0yv0:
Use a specific tag for pulumiCtl. This speed up installing it and avoids putting pressure on the API rate limits for determining the latest release of a dependency in our builds such as https://github.com/pulumi/pulumi-azuredevops/pull/292

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider.config.yaml
+++ b/provider-ci/internal/pkg/templates/bridged-provider.config.yaml
@@ -78,6 +78,7 @@ actionVersions:
   goReleaser: goreleaser/goreleaser-action@v2
   installGhRelease: jaxxstorm/action-install-gh-release@v1.10.0
   checkout: actions/checkout@v4
+  pulumictlTag: v0.0.46
   cleanupArtifact: c-hive/gha-remove-artifacts@v1
   createOrUpdateComment: peter-evans/create-or-update-comment@v1
   downloadArtifact: actions/download-artifact@v2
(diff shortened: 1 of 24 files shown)


provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
```diff
@@ -30,6 +30,7 @@ jobs:
     - name: Install pulumictl
       uses: #{{ .Config.actionVersions.installGhRelease }}#
       with:
+        tag: #{{ .Config.pulumictlTag }}#
```
guineveresaenger, line 33:
```suggestion
        tag: #{{ .Config.actionVersions.pulumictlTag }}#
```

provider-ci/internal/pkg/templates/bridged-provider.config.yaml
```diff
@@ -77,6 +77,7 @@ actionVersions:
   googleAuth: google-github-actions/auth@v2
   goReleaser: goreleaser/goreleaser-action@v2
   installGhRelease: jaxxstorm/action-install-gh-release@v1.5.0
+  pulumictlTag: v0.0.46
```
guineveresaenger, line 80: This is nested under `actionVersions` - that is fine with me because it's adjacent but see below

guineveresaenger (changes requested):
I think you'll need to nest the config value one level.

guineveresaenger:
nice find though! :tada:

guineveresaenger (approved):

Issue #772, opened 2024-01-04
Address workflow warnings


iwahbe:
The generated release workflow has [21 warnings](https://github.com/pulumi/pulumi-azure/actions/runs/7406005148), mostly related to deprecations in GitHub.

The most common errors are:

- The following actions uses node12 which is deprecated and will be forced to run on node16: jaxxstorm/action-install-gh-release@v1.5.0, actions/upload-artifact@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/
- The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

We should address these now, instead of waiting for GH to turn them into errors.

arunsathiya:
>The set-output command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Regarding this, I have a WIP PR here and should be ready for review soon: https://github.com/pulumi/ci-mgmt/pull/789

Also being taken care of in various other Pulumi-owned repositories: https://github.com/search?q=is:pr+author:arunsathiya+org:pulumi&type=pullrequests

VenelinMartinov:
The c sharp warnings are because of https://github.com/pulumi/pulumi/issues/15144

VenelinMartinov:
I'll reopen until I verify we've silenced all warnings we can

VenelinMartinov:
https://github.com/thollander/actions-comment-pull-request is still warning about nodejs version deprecation, looks a bit stale: https://github.com/thollander/actions-comment-pull-request/pull/307

PR #818, opened 2024-02-13
Update github actions

VenelinMartinov:
Should mostly fix https://github.com/pulumi/ci-mgmt/issues/772 and silence some warnings.


iwahbe (approved):
Looks good to me. Since you changed native-provider-ci, I think we should get someone from that team to approve also. 

CC @mjeffryes

VenelinMartinov:
Yeah, I thought I might as well update all actions, LMK if I should revert the native ones.

mjeffryes (approved):

PR #822, opened 2024-02-14
Consume pulumi-package-publisher@v0.0.14

iwahbe:
Upgrades to https://github.com/pulumi/pulumi-package-publisher/releases/tag/v0.0.14, fixing the problem introduced in #818.


iwahbe:
> Please don't forget native providers too

ci-mgmt managed native providers don't actually use pulumi-package-publisher. https://github.com/pulumi/ci-mgmt/pull/822/commits/c26683cbde88b23428783eca8ae37fe7ac53fd49 removed the variable that implied that they did.

PR #827, opened 2024-02-22
Free disk space for go releaser

VenelinMartinov:
Seems to be quite disk space hungry: https://github.com/pulumi/pulumi-azure/actions/runs/8005595075/job/21867100794

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/release.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/release.yml
@@ -144,6 +144,12 @@ jobs:
       #{{- end }}#
     runs-on: #{{ if .Config.runner.publish }}##{{- .Config.runner.publish }}##{{ else }}##{{- .Config.runner.default }}##{{ end }}#
     steps:
+    - name: Free Disk Space (Ubuntu)
+      uses: jlumbroso/free-disk-space@main
+      with:
+        # this might remove tools that are actually needed,
+        # if set to "true" but frees about 6 GB
+        tool-cache: false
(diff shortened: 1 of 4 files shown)


provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
```diff
@@ -181,6 +181,13 @@ jobs:
       #{{- end }}#
     runs-on: #{{ if .Config.runner.publish }}##{{- .Config.runner.publish }}##{{ else }}##{{- .Config.runner.default }}##{{ end }}#
     steps:
+    - name: Free Disk Space (Ubuntu)
+      uses: jlumbroso/free-disk-space@main
+      with:
+        # this might remove tools that are actually needed,
+        # if set to "true" but frees about 6 GB
```
guineveresaenger, line 188: let's explain why we're setting it to `false` rather than saying what it would do if we were setting it to `true`:

```suggestion
        # we are setting this to false to avoid removing tools that are actually needed,
        # but if set to "true" it could free up about 6 GB
```

guineveresaenger (approved):
I love the overall approach!

A couple concerns but non-blocking

1. the [action being used](https://github.com/jlumbroso/free-disk-space) is not being actively maintained. This could be because it's extremely stable - there are no open bugs - but just putting it out there.
2. would it be too much to have our own uninstall script that is a bit more tailored to our needs? The suggested action is a bundle of ` sudo apt-get remove -y` - in the future we might want to add more explicit uninstalls?

Personally I might prefer an explicit run of uninstalls over someone else's choices here but not enough to block.

PR #829, opened 2024-02-27
Clean go sdk build artifacts

VenelinMartinov:
This save some disk space and allows AWS to run on normal runners.

Tested in https://github.com/pulumi/pulumi-aws/pull/3469


iwahbe (approved):
Just to confirm my understanding: this makes it so when we build each module (checking that it compiles), we clean up the resulting artifact after we build.

This seems reasonable to me.

VenelinMartinov:
Yup, exactly

PR #830, opened 2024-02-27
Disable auto plugin acquisition

t0yv0:
This disables automatic plugin acquisition when PULUMI_CONVERT is enabled. Instead the build will encounter errors until all the plugin are specified in .ci-mgmt.yaml explicitly. I can think of some downsides, like these pins now need to be managed and extended during routine upstream upgrades which is not automated, but at the moment it seems that we'll waste less time having more deterministic builds than managing the pins. We can re-evaluate over time.


iwahbe (approved):
LGTM

At some point we should consider letting each provider declare what plugins they need but let ci-mgmt set their versions. We can do that later.

PR #835, opened 2024-02-28
Localize plugin set

### LINKED ISSUE #832: Localize plugin set
https://github.com/pulumi/pulumi-random/pull/724#issuecomment-1967898153

Making PULUMI_HOME=./.pulumi  local during schema generation can avoid leaking provider versions across projects.

EronWright:
This PR updates the provider makefile to use a localized plugin directory (and the localized Pulumi installation) for all code generation tasks.

Note that the other tasks (e.g. `install_sdks`, `test`) do not use the localized Pulumi installation or plugin directory.

Closes #832

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/Makefile
+++ b/provider-ci/internal/pkg/templates/bridged-provider/Makefile
@@ -42,31 +42,35 @@ only_build: build
 
 build_dotnet: DOTNET_VERSION := $(shell pulumictl get version --language dotnet)
 build_dotnet: export PULUMI_HOME := $(WORKING_DIR)/.pulumi
+build_dotnet: export PATH := $(WORKING_DIR)/.pulumi/bin:$(PATH)
 build_dotnet: upstream
 	pulumictl get version --language dotnet
-	PATH=${PWD}/.pulumi/bin:$$PATH PULUMI_CONVERT=$(PULUMI_CONVERT) PULUMI_DISABLE_AUTOMATIC_PLUGIN_ACQUISITION=$(PULUMI_CONVERT) $(WORKING_DIR)/bin/$(TFGEN) dotnet --out sdk/dotnet/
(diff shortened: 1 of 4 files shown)


provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -41,28 +41,32 @@ install_sdks: install_dotnet_sdk install_python_sdk install_nodejs_sdk install_j
...
 build_dotnet: upstream
 	pulumictl get version --language dotnet
-	PULUMI_CONVERT=$(PULUMI_CONVERT) PULUMI_DISABLE_AUTOMATIC_PLUGIN_ACQUISITION=$(PULUMI_CONVERT) $(WORKING_DIR)/bin/$(TFGEN) dotnet --out sdk/dotnet/
+	PATH=${PWD}/.pulumi/bin:$$PATH PULUMI_CONVERT=$(PULUMI_CONVERT) PULUMI_DISABLE_AUTOMATIC_PLUGIN_ACQUISITION=$(PULUMI_CONVERT) $(WORKING_DIR)/bin/$(TFGEN) dotnet --out sdk/dotnet/
 	cd sdk/dotnet/ && \
 		printf "module fake_dotnet_module // Exclude this directory from Go tools\n\ngo 1.17\n" > go.mod && \
 		echo "$(DOTNET_VERSION)" >version.txt && \
 		dotnet build /p:Version=$(DOTNET_VERSION)
 
+build_go: export PULUMI_HOME := $(WORKING_DIR)/.pulumi
```
t0yv0, line 53: Very nice, TIL this nice Make syntax. Any reason that PATH= extension below is also not lifted to this syntax? Might be easier on the eye, though increases the scope of PATH= which I think is OK here.

t0yv0 (approved):

provider-ci/internal/pkg/templates/bridged-provider/Makefile
EronWright: This feature is called [Target-specific Variable Values](https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html).

I copy-pasted the PATH manipulation from `tfgen` target, but yeah I'll try to clean it up.

t0yv0 (approved):
🚢

PR #836, opened 2024-02-28
Add provider bin to PATH for test target

EronWright:
In provider makefiles, the `test` target doesn't seem to work because the examples cannot find the built provider.  The fix is simply to add the provider build directory to the path. That's what the CI script does when running the tests (see [code](https://github.com/pulumi/ci-mgmt/blob/1971f54d81598729424c636745ba961ee6432783/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/run-acceptance-tests.yml#L271-L272)).

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/Makefile
+++ b/provider-ci/internal/pkg/templates/bridged-provider/Makefile
@@ -131,6 +131,7 @@ provider_no_deps:
 
 provider: tfgen provider_no_deps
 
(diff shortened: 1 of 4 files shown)


t0yv0 (approved):
LGTM, I think it can't hurt. In the ideal world the Go tests would manipulate their own path expectations or even build the binaries on demand since some users are interacting with them outside the Makefile possibly but this clearly makes it easier to use Make.

PR #843, opened 2024-03-07
Add dockerbuild provider

blampe:
We discussed rollout plans for the new `buildx.Image` resource and decided to release it as its own standalone provider.

Naming is still TBD but I'm penciling in dockerbuild for now.


mjeffryes (approved):
I strongly dislike "docker-native" as the name (I'm voting for dockerbuild.), but by all means land and iterate.

PR #845, opened 2024-03-12
Reclaim some disk space in release jobs

mjeffryes:
The Kubernetes builds recently started running out of disk during publish. Hoping that this will fix the problem.

Also address some deprecation warnings from goreleaser.
Fixes: https://github.com/pulumi/pulumi-kubernetes/issues/2867

### DIFF
--- a/native-provider-ci/providers/aws-native/repo/.github/workflows/build.yml
+++ b/native-provider-ci/providers/aws-native/repo/.github/workflows/build.yml
@@ -355,6 +355,15 @@ jobs:
       uses: actions/setup-go@v5
       with:
         go-version: ${{ env.GOVERSION }}
+    - name: Clear GitHub Actions Ubuntu runner disk space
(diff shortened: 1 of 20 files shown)


native-provider-ci/providers/aws-native/repo/.github/workflows/build.yml
```diff
@@ -355,6 +355,15 @@ jobs:
       uses: actions/setup-go@v5
       with:
         go-version: ${{ env.GOVERSION }}
+    - name: Clear GitHub Actions Ubuntu runner disk space
+      uses: jlumbroso/free-disk-space@v1
+      with:
+        tool-cache: false
+        dotnet: false
+        android: true
+        haskell: true
```
t0yv0, line 364: Ah but we'll use Haskell one day!

t0yv0 (approved):

PR #851, opened 2024-03-18
chore: add commands for debugging tfgen

corymhall:
This PR adds two commands to make debugging `tfgen` for bridged providers easier.

1. `tfgen_build_only`: Only builds to `tfgen` binary, doesn't run generate. When debugging if you need to rebuild tfgen, you don't need to re-run all the other stuff too.
2. `debug_tfgen`: Starts `tfgen` as a debug server. Instructions on how to debug `tfgen` are being added to the docs in https://github.com/pulumi/pulumi-hugo/pull/4067


VenelinMartinov (approved):
🚀

Issue #815, opened 2024-02-09
Downstream check failures should not create P1s and alert


t0yv0:
The alert are very noisy and use up time of the providers engineer on rotation.  Downstream checks initiate PRs in provider repos to pre-test pre-release bridge and/or pulumi/{pkg,sdk} versions. They should be allowed to fail builds and or checks without creating P1. Bonus points if we can make these tests run without creating PRs in the first place. 

Examples:

- https://github.com/pulumi/pulumi-aws/issues/3405
- https://github.com/pulumi/pulumi-azure/issues/1704
- https://github.com/pulumi/pulumi-gcp/issues/1659
- https://github.com/pulumi/pulumi-azuread/issues/796
- https://github.com/pulumi/pulumi-random/issues/693

VenelinMartinov:
https://github.com/pulumi/home/pull/3327 partially addresses this - it doesn't create issues for faliures after the PR is created but not failures before that.

t0yv0:
This is highly encouraging, great find! Do you have an idea where the rest of it is controlled, or need help poking around?

VenelinMartinov:
I think we need an additional check in the lambda above - essentially when the workflow fails, we need to check it if was triggered by a downstream test workflow.

https://github.com/pulumi/home/pull/3327 added a check if the PR that triggered the workflow starts with [DOWNSTREAM TEST] - that is not sufficient since the workflow which opens that PR can fail itself - that workflow is not triggered from the PR. Example: https://github.com/pulumi/pulumi-gcp/issues/1819

t0yv0:
https://github.com/pulumi/home/tree/master/infrastructure/github-automation/workflow-issue-ops perhaps is relevant

t0yv0:
I'm going to reopen this based don the following evidence:

** https://github.com/pulumi/pulumi-aws/issues/3691 workflow failure: Upgrade bridge

- https://github.com/pulumi/pulumi-aws/actions/runs/8405275816/job/23017784736 This is a downstream codegen test for pulumi/pulumi#15770
  https://github.com/pulumi/pulumi-aws/actions/runs/8411074794/job/23030245058 This is a downstream codegen test for pulumi/pulumi#15770
  https://github.com/pulumi/pulumi-aws/actions/runs/8411612958/job/23031395056 This is a downstream codegen test for pulumi/pulumi#15772. (run-id: 8411612189)
[...]

PR #853, opened 2024-03-19
chore: add `upgrade-bridge-test` as possible repository_dispatch type

corymhall:
When you trigger a workflow via a repository_dispatch, you have to specify a custom `type`. This type appears in the workflow_run webhook event so it can be used to filter different types of runs.

I'm adding a new type `upgrade-bridge-test` which can be used by the downstream test workflows. This will allow us to [filter out those workflow runs](https://github.com/pulumi/home/pull/3359) from creating p1 tickets.

re https://github.com/pulumi/ci-mgmt/issues/815


VenelinMartinov (approved):
Great stuff!

PR #855, opened 2024-03-21
Add downstream test job to CI

VenelinMartinov:
Fixes https://github.com/pulumi/ci-mgmt/issues/856

This makes ci-mgmt run acceptance-tests on [pulumi-xyz](https://github.com/pulumi/pulumi-xyz) and verifies that they all succeed.

It should prevent errors like https://github.com/pulumi/ci-mgmt/pull/837


corymhall (approved):
🚀

PR #860, opened 2024-03-25
Introduce examples cache 

t0yv0:
This should unlock some speedups especially in large providers like pulumi-aws by reusing available results for the examples generation pass and reducing the wait time on pulumi-converter-terraform.

Generalized form of https://github.com/pulumi/pulumi-aws/pull/3550

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.splice
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.splice
@@ -7,6 +7,12 @@
     - name: Unshallow clone for tags
       run: git fetch --prune --unshallow --tags
 #{{ .Config.actions.setupGo | toYaml | indent 4 }}#
+    - name: Cache examples generation
+      uses: actions/cache@v4
+      with:
+        path: |
+          .pulumi/examples-cache
+        key: ${{ runner.os }}-${{ hashFiles('provider/go.sum') }}
     - name: Install pulumictl
       uses: #{{ .Config.actionVersions.installGhRelease }}#
       with:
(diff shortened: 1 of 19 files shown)


iwahbe (approved):
I'm a bit nervous here, since I think that the cache key effectively needs to be the SHA of the entire commit (omitting /sdk).

I'm willing to test in prod here, since any break *will not* be customer facing.

t0yv0:
The cache key at GHA level is approximate. Inside that directory, the system will be storing subfolders with a much more elaborate cache key. So serving false results from a cache hit is much less of a possibility unless the elaborate cache key is imprecise. I think it is a little bit imprecise still, we could use MarshallableProviderInfo hash as part of it, but it does account for a lot of things.

iwahbe:
> The cache key at GHA level is approximate. Inside that directory, the system will be storing subfolders with a much more elaborate cache key. So serving false results from a cache hit is much less of a possibility unless the elaborate cache key is imprecise. I think it is a little bit imprecise still, we could use MarshallableProviderInfo hash as part of it, but it does account for a lot of things.

I'm saying that to ensure no false cache hits, the *elaborate* cache key needs to be more general. Here are two examples:
- The cache key needs to include `resources.go` since that is how we do property name overrides and those effect generated examples. For larger providers, we do name overrides in other files (`labels.go` for GCP, `tags.go` for AWS), so the *elaborate* key really needs to include each `.go` file in `provider/`. 
- The cache key needs to include the upstream provider (not captured in go.mod for patched or shimmed providers) as well as the `patches` folder (which can make arbitrary changes to upstream).

t0yv0:
I think all the information for resources.go as well as upstream patches is condensed to MarshallableProviderInfo; when we are converting HCL to PCL that's literally it - we're passing MPI and example sources to a given Pulumi CLI and converter versions. The PCL->{Py,TS} part has not been externalized yet but is essentially similar. I think if we include MarshallableProviderInfo hash we'll be good for all practical purposes. I'll do that before landing this.

t0yv0:
This has been tested to some success in pulumi-aws with some more bridge changes which are now in and released, so I am merging 🤞 hoping for no disruption here.

PR #861, opened 2024-03-25
Drop old source migrations

t0yv0:
I noticed that running workflow update in pulumi-aws fails to recognize that pyproject.toml is already rolled out and runs tfgen again making it very slow. It's probably a good idea to clean up all the old "source migrations" as they have done their job all the repositories have the edits applied.

### DIFF
--- a/tools/sourcemigrator/index.ts
+++ b/tools/sourcemigrator/index.ts
@@ -26,129 +26,25 @@ interface MigrateResult {
     filesEdited: Number;
 }
 
-function updateExamplesFromCore31DotNet6(): SourceMigration {
-    let pattern = new RegExp("[<]TargetFramework[>]netcoreapp3.1[<][/]TargetFramework[>]");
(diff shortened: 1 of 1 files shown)


tools/sourcemigrator/index.ts
```diff
@@ -26,129 +26,25 @@ interface MigrateResult {
...
-                    if (replaced) {
-                        child.execSync("go mod tidy", {cwd: path.dirname(f)});
-                    }
-                    return replaced;
-                }).length;
-            return {filesEdited: filesEdited};
-        },
-    };
-    return sm;
-}
```
mjeffryes, line 83: I wonder if it would be worth retaining one or more of these as examples and just removing them from the active migrations list.

mjeffryes (approved):

tools/sourcemigrator/index.ts
t0yv0: Yeah I had the same thought. I'll keep one.

iwahbe (approved):

PR #862, opened 2024-03-26
Add a config section check for bridged providers

VenelinMartinov:
fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1585

This PR adds a check in bridged provider CI for configuration options in README.md.
CI will now comment on PRs if any configuration options are missing in the README.md

Tested in https://github.com/pulumi/pulumi-github/pull/615 and https://github.com/pulumi/pulumi-xyz/pull/31

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/prerequisites.splice
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/prerequisites.splice
@@ -78,11 +78,11 @@
         message: >+
           ### Is README.md missing any configuration options?
 
-          ${{ env.MISSING_CONFIG }}
+          ${{ env.MISSING_CONFIG || 'No missing config!' }}
 
 
-          Please add a description for each of these options to `README.md`.
-          Details about them can be found in either the upstream docs or `schema.json`.
+          ${{ env.MISSING_CONFIG && 'Please add a description for each of these options to `README.md`.' }}
+          ${{ env.MISSING_CONFIG && 'Details about them can be found in either the upstream docs or `schema.json`.' }}
 
     - name: Tar provider binaries
       run: tar -zcf ${{ github.workspace }}/bin/provider.tar.gz -C ${{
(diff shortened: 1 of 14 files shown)


provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/prerequisites.splice
```diff
@@ -57,6 +57,33 @@
 
           Maintainer note: consult the [runbook](https://github.com/pulumi/platform-providers-team/blob/main/playbooks/tf-provider-updating.md) for dealing with any breaking changes.
 
+    - if: github.event_name == 'pull_request'
+      name: Check Configuration section
+      run: |
+        sed -n '/## Configuration/,$p' README.md | sed -n '/## Reference/q;p' >> config_section.txt
+        jq -r '.config[] | keys[]' < provider/cmd/pulumi-resource-${{ env.PROVIDER }}/schema.json >> keys.txt
+        EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
```
t0yv0, line 65: /dev/urandom ?

t0yv0:
I'm not a great fan of bash-in-yaml for something like this but it works for the moment.

t0yv0 (approved):

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/prerequisites.splice
VenelinMartinov: I copied it from https://github.com/pulumi/ci-mgmt/blob/43d79c11a73a2b3301165f69a24ccfddc14a0909/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/prerequisites.splice#L39 which was probably copied from this SO: https://stackoverflow.com/a/74232400

Looks like GH uses the sequence as an EOF marker and it's random so it doesn't match anything in the actual string.

PR #866, opened 2024-03-27
Accurate go cache for prerequisites

t0yv0:
Running some experiments in pulumi-aws I found that for simple PRs that edit tests or resources.go it's really nice to have accurate Go cache. Specifically prerequisites job cache should be keyed on upstream/go.sum and provider/go.sum; if that is done than cache hits mean no additional go package downloading is happening, greatly speeding up builds (I got ranges from 3min to 20min, with 3 min on hot caches). 

I think some of the prior thinking here was that using an approximate cache key would allow to reuse some of the cache even when bridge or upstream upgrades are happening. However when inaccurate cache is downloaded, Go proceeds to download the actual dependences, which in AWS case are significant, causing us to run out of disk space. To compensate, we installed a disk space cleaner job that eats 2.5 minutes of build time negating any gains we might have had. Alas. 

Accurate cache is probably the more intuitive way forward here.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/prerequisites.splice
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/prerequisites.splice
@@ -16,7 +16,13 @@
       run: git fetch --prune --unshallow --tags
     - name: Prepare upstream code
       run: make upstream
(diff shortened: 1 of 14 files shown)


VenelinMartinov (approved):
The disk space cleaner job was added when we moved to the smaller CI instances. Feel free to get rid of it if CI works without it!

PR #867, opened 2024-03-27
Fix update workflows PRs title

VenelinMartinov:
Seems like in https://github.com/pulumi/ci-mgmt/pull/855 we regressed and update PRs are now called `falseUpdate GitHub Actions workflows.`

example: https://github.com/pulumi/pulumi-gcp/pull/1878

this should fix it and correctly return them to `Update GitHub Actions workflows.`


iwahbe (approved):
My kingdom for a GH action type checker. Anyway...

LGTM

PR #868, opened 2024-03-27
Fix array indices in config check

VenelinMartinov:
My jq syntax in https://github.com/pulumi/ci-mgmt/pull/862 was off yielding 0s and 1s for some providers:
https://github.com/pulumi/pulumi-azure/pull/1917#pullrequestreview-1963315541

The issue was that their schema had a defaults section in the config block.

This PR fixes the jq syntax to correctly ignore that bit.


danielrbradley (commented):
Not really sure I have the context here to understand how the `[]` -> `.variables` fixes this. How do we normally test this kind of change in ci-mgmt?

iwahbe:
It looks like it's still failing, now in a new way.

VenelinMartinov:
Tested this on azure and xyz - both work now and don't yield the 0s and 1s

iwahbe (approved):

PR #877, opened 2024-04-09
Flag config check

VenelinMartinov:
Flag the check added in https://github.com/pulumi/ci-mgmt/pull/862 to only run for providers which have enabled it.

We should fix the warnings for a provider and then enable the check to guard against regressions.


iwahbe (approved):
2 suggestions:

1. Can we turn this on for one of our test workflows?
2. If we are going to only turn this on when we know a provider is fully up to data on their README, can we make it non-advisory?

VenelinMartinov:
Made non-advisory and test in https://github.com/pulumi/pulumi-civo/pull/488

PR #883, opened 2024-04-15
chore: create upstream patch files with `--no-numbered`

corymhall:
Currently the patch files are generated with subjects that contain the patch number (i.e. `Subject: [PATCH 02/52] Add S3 legacy bucket to resources`). This makes diffs too verbose since everytime I add or remove a patch I have a diff on _every_ one of the patches.

The patches have the patch number in the name of the file already so the number in the subject is redundant.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/scripts/upstream.sh
+++ b/provider-ci/internal/pkg/templates/bridged-provider/scripts/upstream.sh
@@ -176,7 +176,7 @@ end_rebase() {
 
   rm patches/*.patch
   cd upstream
-  git format-patch local -o ../patches --zero-commit --no-signature --no-stat
+  git format-patch local -o ../patches --zero-commit --no-signature --no-stat --no-numbered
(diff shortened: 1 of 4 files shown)


t0yv0 (changes requested):
I'm for this in principle but it needs to go to ci-mgmt as this edit will be wiped out:

```# WARNING: This file is autogenerated - changes will be overwritten if not made via https://github.com/pulumi/ci-mgmt
```

corymhall:
> I'm for this in principle but it needs to go to ci-mgmt as this edit will be wiped out:

This is in ci-mgmt 😀

t0yv0:
Ah it trips me up that the templates have this "warning" copy.

t0yv0:
Affects Ian's squad so adding as a reviewer. LGTM>

t0yv0 (approved):

iwahbe (approved):
LGTM

PR #892, opened 2024-04-17
Onboard kubernetes-cert-manager to ci-mgmt

blampe:
[kubernetes-cert-manager](https://github.com/pulumi/pulumi-kubernetes-cert-manager) is an older provider whose releases are currently [broken](https://github.com/pulumi/pulumi-kubernetes-cert-manager/actions/runs/7996599325).

Rather than hacking together a one-time fix to unblock releases, it would be better to manage this repo via ci-mgmt so it can stay up-to-date with the rest of the ecosystem. 

This can merge after https://github.com/pulumi/pulumi-kubernetes-cert-manager/pull/38, which updates the repo to work with these workflows.


mjeffryes (approved):
Thank you for doing this! Maybe we can add 
https://github.com/pulumi/pulumi-kubernetes-coredns
https://github.com/pulumi/pulumi-kubernetes-ingress-nginx
the same way.

blampe:
> Thank you for doing this! Maybe we can add https://github.com/pulumi/pulumi-kubernetes-coredns https://github.com/pulumi/pulumi-kubernetes-ingress-nginx the same way.

Definitely. It'll probably take a day or two each to work through various build issues, but very doable.

PR #900, opened 2024-04-25
Rework versioning

danielrbradley:
Stop using pulumictl to re-calculate the build version by scanning the local git repo.
- Use a hard-coded dev version for local builds: `[MAJOR].0.0-alpha.0+dev`
- Use [pulumi/provider-version-action](https://github.com/pulumi/provider-version-action) for github builds.

Changes rational:
- Ensure we have PROVIDER_VERSION set whenever we call `make ...` in CI
  - Except for the upgrade provider workflow where we can just run with the dev version so it's more similar to local development as we're just pushing to open a PR.
- Use the version action when calling goreleaser.
- Remove "unshallow clone" steps as we no longer need a full checkout for pulumictl to calculate the version.
  - Leave "unshallow clone" in the check-upstream-upgrade as unsure if the upgrade-provider tool requires all tags to be checked out.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.splice
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.splice
@@ -1,3 +1,6 @@
+    - uses: pulumi/provider-version-action@v1
+      with:
+        set-env: 'PROVIDER_VERSION'
     - name: Checkout Scripts Repo
       uses: #{{ .Config.actionVersions.checkout }}#
       with:
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.splice
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.splice
@@ -58,9 +58,6 @@
       run: make install_plugins
     - name: Update path
       run: echo "${{ github.workspace }}/bin" >> "$GITHUB_PATH"
-    - name: Set PACKAGE_VERSION to Env
-      run: echo "PACKAGE_VERSION=$(pulumictl get version --language generic)" >>
-        "$GITHUB_ENV"
     - name: Build SDK
       run: make build_${{ matrix.language }}
     - name: Check worktree clean
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.splice
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.splice
@@ -7,8 +7,6 @@
         path: ci-scripts
         repository: pulumi/scripts
         ref: #{{ .Config.pulumiScriptsRef }}#
-    - name: Unshallow clone for tags
-      run: git fetch --prune --unshallow --tags
 #{{ .Config.actions.setupGo | toYaml | indent 4 }}#
     - name: Cache examples generation
       uses: actions/cache@v4
--- a/provider-ci/test-workflows/cloudflare/.github/workflows/release.yml
+++ b/provider-ci/test-workflows/cloudflare/.github/workflows/release.yml
@@ -47,8 +47,6 @@ jobs:
         path: ci-scripts
         repository: pulumi/scripts
         ref: deca2c5c6015ad7aaea6f572a1c2b198ca323592
-    - name: Unshallow clone for tags
-      run: git fetch --prune --unshallow --tags
     - name: Install Go
       uses: actions/setup-go@v5
       with:
@@ -181,8 +179,6 @@ jobs:
         path: ci-scripts
         repository: pulumi/scripts
         ref: deca2c5c6015ad7aaea6f572a1c2b198ca323592
-    - name: Unshallow clone for tags
-      run: git fetch --prune --unshallow --tags
     - name: Prepare upstream code
       run: make upstream
     - name: Install Go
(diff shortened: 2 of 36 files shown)


danielrbradley:
The downstream xyz test is failing because its `ci-mgmt.yml` specifies its major version as `1`, but pulumictl was then returning version `0.0.1-alpha.1714037557+8fe5ba22.dirty` because it hasn't yet been released. When the updated build uses the version `1.0.0-alpha.0+dev` it removes the `getDataSource` method because the bridge metadata has:

```json
            "xyz_data_source": {
                "current": "xyz:index/dataSource:DataSource",
                "past": [
                    {
                        "name": "xyz:index/getDataSource:getDataSource",
                        "inCodegen": true,
                        "majorVersion": 0
                    }
                ]
            }
```

Therefore it seems like moving to version 1 actually causes the method to be removed.

I need to identify if pulumictl was incorrect here, if the new behaviour is correct, and how we should fix the xyz test.

VenelinMartinov (approved):
lgtm, have you tested the release workflow with these changes anywhere? Might be worth giving it a try before rolling out to all providers.

provider-ci/test-workflows/cloudflare/.github/workflows/release.yml
```diff
@@ -166,6 +164,9 @@ jobs:
     steps:
     - name: Checkout Repo
       uses: actions/checkout@v4
+    - uses: pulumi/provider-version-action@v1
```
t0yv0, line 167: I'm nervous about this a little - has this been code-reviewed, tested on a small provider? Rolling out straight to AWS? Also... Will this calculate the same version across the build steps *even if* tags are added or removed concurrently with the build? Something more foolproof would call this once, save an artifact and ref that artifact in all subsequent phases.

danielrbradley:
@iwahbe yup - rolling this out first to p-azure:
- https://github.com/pulumi/pulumi-azure/pull/1977

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.splice
```diff
@@ -1,11 +1,12 @@
+    - uses: pulumi/provider-version-action@v1
+      with:
+        set-env: 'PROVIDER_VERSION'
     - name: Checkout Scripts Repo
       uses: #{{ .Config.actionVersions.checkout }}#
       with:
         path: ci-scripts
         repository: pulumi/scripts
         ref: #{{ .Config.pulumiScriptsRef }}#
```
t0yv0, line 9: The removal of UNshallow clone for tags is a very good speedup!

t0yv0 (approved):
If reviewed/tested already let's try this! The hermeticity with respect to concurrent tags is a pre-existing problem .

iwahbe (approved):
I wouldn't be shocked if something breaks here, but I think the speedup is worth the potential pain. Lets give it a shot.

danielrbradley:
I'll let p-azure operate with this until Monday to see if any issues arise, then merge. Thanks!

provider-ci/test-workflows/cloudflare/.github/workflows/release.yml
t0yv0: This is the way to go for sure, possibly separate change: > This is why in azure-native we have a separate job at the beginning that generates the version once that we then refer to.

t0yv0:
I also want to point out that using a Go binary like pulumictl was a better architecture (more portable) than locking the logic in a GitHub action, though we might getting an upgrade on the implementation *quality*.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.splice
danielrbradley: Indeed - I expect this will save nearly 3 minutes per job in an AWS workflow run! 🚀

blampe:
@danielrbradley how would this handle back porting fixes if the version is derived from ci-mgmt's major version? 

For example I have a use case right now where I'd like to make some patch releases to docker v4 _and_ v3 to update docstrings for a resource to point folks to the new provider.

I was planning on cutting a v3 back port branch and tagging that. In this new world, I'm guessing I would check in a local `.ci-mgmt.yaml` to that branch with a major version 3 and everything should just work?

danielrbradley:
@blampe ah good thought ... 

1. This will work perfectly for tagged releases as it will use the tag as the version.
2. This already fixes building old versions locally as it has the major version baked into the makefile rather than scanning git.
3. We could extend the version action to understand version branches e.g. if the current branch is `v2` then the major version for dev builds will also be `2.x...`

t0yv0:
> As for the architecture choice

This works perfectly for internal needs but users with non-GitHub setups are out of luck, moves us to being more coupled to GHA not less.

danielrbradley:
> > As for the architecture choice
> 
> This works perfectly for internal needs but users with non-GitHub setups are out of luck, moves us to being more coupled to GHA not less.

Moving this to a Go tool wouldn't help with that - it would just make the Go tool coupled to GitHub instead. This setup means only the GH Action is tied to GH and the interface for another CI would be to set the PROVIDER_VERSION based on the context of the run.

The only alternative is what we had previously which was coupled to git instead .. but this requires a full checkout which is not as reliable a context for a release and has performance impact.

t0yv0:
Perfect, thanks for explaining, so the new strategy really only works in GitHub actions which is a perfect vehicle then. I guess we still have pulumictl around for the alternative if it's ever needed. Thanks!

PR #902, opened 2024-04-25
Exclude ci-scripts from check-worktree-is-clean

blampe:
`ci-scripts` is flagging itself for being untracked, so let's ignore it. This seems more straightforward than updating `.gitignore` everywhere.

```
Run ./ci-scripts/ci/check-worktree-is-clean
error: working tree is not clean, aborting!
HEAD detached at pull/901/merge
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	ci-scripts/
```

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -44,14 +44,13 @@ jobs:
         with:
           path: ci-scripts
           repository: pulumi/scripts
-        env:
-          GITHUB_WORKSPACE: ${{ runner.temp }}
+      - run: mv ci-scripts ../ci-scripts # actions/checkout#197
       - name: Validate native-providers
         run: cd native-provider-ci && make all
       - name: Validate that our test workflows match checked in workflows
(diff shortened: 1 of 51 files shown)


.github/workflows/pull-request.yml
```diff
@@ -44,12 +44,13 @@ jobs:
         with:
           path: ci-scripts
           repository: pulumi/scripts
+      - run: mv ci-scripts ../ci-scripts # actions/checkout#197
```
rquitales, line 47: We could also just add ci-scripts to `.git/info/exclude` if we don't want to move any files (eg. `echo "ci-scripts" >> .git/info/exclude`). Then we don't need to update the relative path where the scripts are called from.

rquitales (approved):

.github/workflows/pull-request.yml
blampe: This was a great idea!

PR #904, opened 2024-04-29
Use git worktree clean action

danielrbradley:
Avoid using ci-scripts for checking if the worktree is clean - use the new dedicated action instead.

The motivation for this change is to make it easier to allow specific changes to be allowed once we start injecting versions into SDKs during generation.

This leaves our only usage of ci-scripts for running `ci-scripts/ci/publish-tfgen-package`.

Use new version action in native providers too (missed from https://github.com/pulumi/ci-mgmt/pull/900). Once this is merged, each native provider can update its makefile to use the version variable rather than calling `pulumictl get version`. Then we can can remove the fetching of tags.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -39,18 +39,12 @@ jobs:
         uses: jaxxstorm/action-install-gh-release@v1.11.0
         with:
           repo: pulumi/pulumictl
-      - name: Checkout Scripts Repo
-        uses: actions/checkout@v4
-        with:
-          path: ci-scripts
-          repository: pulumi/scripts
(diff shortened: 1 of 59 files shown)


t0yv0 (approved):

danielrbradley:
Deploying to pulumi-azure for testing: https://github.com/pulumi/ci-mgmt/actions/runs/8896633308

danielrbradley:
Re-testing via random:
1. https://github.com/pulumi/ci-mgmt/actions/runs/8898114739/job/24434560647
2. https://github.com/pulumi/pulumi-random/pull/909

PR #905, opened 2024-04-29
Use native GitHub approvals for community PRs

blampe:
Inspired by https://github.com/pulumi/pulumi/pull/16083.

GitHub [natively](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) allows maintainers to manually kick off tests on PRs from community members. It's straightforward and doesn't hide test results behind a URL.

This PR removes all the logic around `/run-acceptance-tests` and checking for outside forks in favor of GitHub's native functionality.

Caveat: GitHub doesn't currently provide an API for configuring fork approval settings (https://github.com/integrations/terraform-provider-github/issues/2108), and the default is "Require approval for first-time contributors" whereas the current workflow is equivalent to "Require approval for all outside collaborators".

I've changed the pulumi org's default permission to always require approval for outside PRs, and this seems to have the intended effect.


native-provider-ci/providers/aws-native/repo/.github/workflows/run-acceptance-tests.yml
```diff
@@ -66,7 +49,6 @@ jobs:
       uses: actions/checkout@v4
       with:
         lfs: true
-        ref: ${{ env.PR_COMMIT_SHA }}
```
blampe, line 69: This was being appended as part of `addDispatchConditional`. In theory it's not needed -- GitHub will checkout a merge commit of the PR by default, which is good. We got rid of this in some other actions as part of https://github.com/pulumi/ci-mgmt/pull/687 which makes me more confident it's safe to remove.

EronWright (approved):

danielrbradley (approved):
This is a great simplification!

By changing the org-level setting does this mean we don't need to set it to always require approvals on a per-repo basis?

blampe:
> By changing the org-level setting does this mean we don't need to set it to always require approvals on a per-repo basis?

I think so, yes. pulumi/kubernetes was configured for "first-time contributors" before I tweaked the org-level setting, but afterwards it showed "all outside collaborators". Seems to be the case with all the other repos I spot-checked as well.

tgummerer (approved):

iwahbe:
@blampe Can we cary this forward?

blampe:
> @blampe Can we cary this forward?

@iwahbe unclear, probably needs more thought (especially for provider repos since we almost always require some kind of credentials during tests).

Some additional context in https://github.com/pulumi/pulumi/pull/16083 and https://pulumi.slack.com/archives/C02FXTZEZ6W/p1714422853089509.

PR #910, opened 2024-05-03
Fix nightly SDK gen for aws-native

t0yv0:
Fixes failures such as https://github.com/pulumi/pulumi-aws-native/issues/1514

### DIFF
--- a/native-provider-ci/src/steps.ts
+++ b/native-provider-ci/src/steps.ts
@@ -102,9 +102,8 @@ export function CommitAutomatedSDKUpdates(provider: string): Step {
       "git add sdk\n" +
       `git commit -m "Regenerating SDKs based on ${dir} @ \${{ steps.vars.outputs.commit-hash }}" || echo "ignore commit failure, may be empty"\n` +
       "git add .\n" +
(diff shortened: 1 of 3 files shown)


t0yv0:
Tested in https://github.com/pulumi/pulumi-aws-native/actions/runs/8944147845

native-provider-ci/src/steps.ts
```diff
@@ -103,7 +103,8 @@ export function CommitAutomatedSDKUpdates(provider: string): Step {
       `git commit -m "Regenerating SDKs based on ${dir} @ \${{ steps.vars.outputs.commit-hash }}" || echo "ignore commit failure, may be empty"\n` +
       "git add .\n" +
       `git commit -m "Regenerating based on ${dir} @ \${{ steps.vars.outputs.commit-hash }}"\n` +
-      "git push origin generate-sdk/${{ github.run_id }}-${{ github.run_number }}",
+      "git push origin generate-sdk/${{ github.run_id }}-${{ github.run_number }}" +
+      ` || echo "ignore commit failure, may be empty"\n`,
```
mjeffryes, line 107: should this be on the `git commit` command above (like we do for the SDK commit)? If it's really the push command we want to ignore failures on, maybe the message should be different?

mjeffryes (approved):

PR #912, opened 2024-05-07
Add worktreeAllowedChanges config option

danielrbradley:
- Default to having no effect.
- This is needed as we start to onboard providers into embedding version numbers in SDKs.


flostadler (approved):
LGTM!

PR #917, opened 2024-05-09
Download & restore Go SDK before pushing

danielrbradley:
- Ensure we have the SDK with the latest version number included before publishing.


VenelinMartinov (approved):
looks good!

Issue #915, opened 2024-05-08
Embed version in SDKs


danielrbradley:
Part of https://github.com/pulumi/home/issues/3372

1. Enable embedding version number in each language's SDK

```go
RespectSchemaVersion: true
```

2. Configure ci-mgmt

Versions are stable within the same major version for local development but will be updated during the build in CI.

```yaml
# These files are updated with the real version number
worktreeAllowedChanges: |-
  sdk/**/pulumi-plugin.json
  sdk/dotnet/Pulumi.*.csproj
  sdk/go/*/internal/pulumiUtilities.go
  sdk/nodejs/package.json
  sdk/python/pyproject.toml

publish:
  goSdk:
    usePush: true
```

## Native Providers

PR for the managed providers: https://github.com/pulumi/ci-mgmt/pull/925

- [x] aws-apigateway (not managed) [PR](https://github.com/pulumi/pulumi-aws-apigateway/pull/161)
- [x] aws-native [CI PR](https://github.com/pulumi/pulumi-aws-native/pull/1538) & [Schema PR](https://github.com/pulumi/pulumi-aws-native/pull/1539)
- [x] awsx (not managed) [PR](https://github.com/pulumi/pulumi-awsx/pull/1297)
- [x] azure-native (not managed) [PR](https://github.com/pulumi/pulumi-azure-native/pull/3289)
- [x] command [CI PR](https://github.com/pulumi/pulumi-command/pull/434) & [Schema PR](https://github.com/pulumi/pulumi-command/pull/438)
- [x] docker-build [CI PR](https://github.com/pulumi/pulumi-docker-build/pull/66) & [Schema PR](https://github.com/pulumi/pulumi-docker-build/pull/68)
- [x] eks (not managed) [PR](https://github.com/pulumi/pulumi-eks/pull/1232)
- [x] google-native (deprecated)
- [x] kubernetes [CI PR](https://github.com/pulumi/pulumi-kubernetes/pull/3008) & [Schema PR](https://github.com/pulumi/pulumi-kubernetes/pull/3028)
- [x] [kubernetes-cert-manager](https://github.com/pulumi/pulumi-kubernetes-cert-manager) [CI PR](https://github.com/pulumi/pulumi-kubernetes-cert-manager/pull/60) & [Schema PR](https://github.com/pulumi/pulumi-kubernetes-cert-manager/pull/67)

## Bridged Providers

PR: https://github.com/pulumi/ci-mgmt/pull/920

[...]

PR #920, opened 2024-05-10
Embed SDK version

danielrbradley:
1. Change default config to use push-based Go SDK publishing
2. Use gopatch to enable RespectSchemaVersion in all languages
3. Rebuild SDKs if we've updated the resources.go file

Related to https://github.com/pulumi/ci-mgmt/issues/915 and https://github.com/pulumi/home/issues/3372

### DIFF
--- a/tools/sourcemigrator/index.ts
+++ b/tools/sourcemigrator/index.ts
@@ -77,7 +77,7 @@ function respectSchemaVersion(): SourceMigration {
         shell.exec(`gofumpt -w ./provider/resources.go`);
         // Check if we've made changes
         const gitStatus = shell.exec(`git status --porcelain`).stdout;
-        if (gitStatus.trim() !== "") {
+        if (gitStatus.includes("provider/resources.go")) {
           shell.exec(`make tfgen build`);
         }
       } finally {
--- a/tools/sourcemigrator/index.ts
+++ b/tools/sourcemigrator/index.ts
@@ -61,6 +61,17 @@ interface MigrateContext {
 //     }
 // }
 
+function checkError(res: shell.ShellString): shell.ShellString {
+  if (res.code !== 0) {
+    throw new Error(`Command failed: ${res.stderr}`);
+  }
+  return res;
+}
+
+function run(command: string): shell.ShellString {
+  return checkError(shell.exec(command));
+}
+
 function respectSchemaVersion(): SourceMigration {
   return {
     name: "Respect Schema Version",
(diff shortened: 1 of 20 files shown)


iwahbe:
@danielrbradley Can you pick a bridged provider and deploy this code all the way through a release?

danielrbradley:
Rolling out to pulumi-azure as a test:
- Action to create PR: https://github.com/pulumi/ci-mgmt/actions/runs/9059851553
- Azure PR: https://github.com/pulumi/pulumi-azure/pull/2028

danielrbradley:
A couple of warts we might want to address that I spotted during the first test rollout:

- The SDK commit actually contains all the SDKs, not only the Go SDK. This is because we just point the Go SDK publish action as the module - which is `sdk/` - where the go.mod lives - which also includes all the other SDKs. This isn't a massive issue, but perhaps we could extend the action to allow globbing to select which files to include/exclude.
- Linked to the above issue, we also include the `go.tar.gz` file as we currently always download the assets into the sdk folder rather than a temp directory ... which results in it getting included when we publish the `sdk/...` go module.

iwahbe:
> A couple of warts we might want to address that I spotted during the first test rollout:
> 
> * The SDK commit actually contains all the SDKs, not only the Go SDK. This is because we just point the Go SDK publish action as the module - which is `sdk/` - where the go.mod lives - which also includes all the other SDKs. This isn't a massive issue, but perhaps we could extend the action to allow globbing to select which files to include/exclude.
> * Linked to the above issue, we also include the `go.tar.gz` file as we currently always download the assets into the sdk folder rather than a temp directory ... which results in it getting included when we publish the `sdk/...` go module.

I think it's worth addressing both before we deploy to all providers. I don't want to regress in download size, so we **need** to address the second point before deploying. For simplicity sake, I would like to address the first point as well, but it's not necessary.

> but perhaps we could extend the action to allow globbing to select which files to include/exclude

Instead, can we just hard-code the list of paths we want to include? Right now, any pulumi provider should have all of their go code within `sdk/{go.mod,go.sum,go}`. `pulumi/publish-go-sdk-action@v1` can just know that.

iwahbe (commented):
> A couple of warts we might want to address that I spotted during the first test rollout:
> 
> * The SDK commit actually contains all the SDKs, not only the Go SDK. This is because we just point the Go SDK publish action as the module - which is `sdk/` - where the go.mod lives - which also includes all the other SDKs. This isn't a massive issue, but perhaps we could extend the action to allow globbing to select which files to include/exclude.
> * Linked to the above issue, we also include the `go.tar.gz` file as we currently always download the assets into the sdk folder rather than a temp directory ... which results in it getting included when we publish the `sdk/...` go module.

I think it's worth addressing both before we deploy to all providers. I don't want to regress in download size, so we **need** to address the second point before deploying. For simplicity sake, I would like to address the first point as well, but it's not necessary.

> but perhaps we could extend the action to allow globbing to select which files to include/exclude

Instead, can we just hard-code the list of paths we want to include? Right now, any pulumi provider should have all of their go code within `sdk/{go.mod,go.sum,go}`. `pulumi/publish-go-sdk-action@v1` can just know that.

danielrbradley:
> I think it's worth addressing both before we deploy to all providers. I don't want to regress in download size, so we **need** to address the second point before deploying. For simplicity sake, I would like to address the first point as well, but it's not necessary.

Done - make the action customisable with globs for the files to include, then customised the list of files for our use case within ci-mgmt as this might change in the future as we do things like move the go.mod one day or publish to a different repo.

tools/sourcemigrator/index.ts
```diff
@@ -64,29 +61,65 @@ interface MigrateResult {
...
+          `go run github.com/uber-go/gopatch@latest -p "${patchPath}" ./provider/resources.go`
+        );
+        // Format the code - twice to ensure that the code is formatted correctly
+        run(`go install mvdan.cc/gofumpt@latest`);
+        run(`gofumpt -w ./provider/resources.go`);
+        run(`gofumpt -w ./provider/resources.go`);
+        // Check if we've made changes
+        const gitStatus = run(`git status --porcelain`).stdout;
+        if (gitStatus.includes("provider/resources.go")) {
+          run(`make tfgen build`);
```
VenelinMartinov, line 93: Do you not need to rebuild the sdks too?

danielrbradley: `make build` includes building the SDKs

PR #931, opened 2024-05-16
Fix native Java & Go SDK publish

danielrbradley:
- Fix Publish Go Sdk dependency - typo between underscore and hyphen.
- Fix Java publish by using the consistent version rather than calling `pulumictl get version`.
- Fix cf2pulumi Chocolatey version - use consistent version rather than calling `pulumictl get version`.
- Remove unused steps and workflows.


iwahbe (approved):
Testing in prod :)

PR #932, opened 2024-05-17
Fix docs dependency in native release

danielrbradley:
tag_sdk no longer exists and has been replaced by pubish_go_sdk

### DIFF
--- a/native-provider-ci/providers/aws-native/repo/.github/workflows/release.yml
+++ b/native-provider-ci/providers/aws-native/repo/.github/workflows/release.yml
@@ -564,7 +564,7 @@ jobs:
           !*.tar.gz
   dispatch_docs_build:
     runs-on: ubuntu-latest
-    needs: tag_sdk
+    needs: pubish_go_sdk
     steps:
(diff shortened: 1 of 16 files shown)


native-provider-ci/providers/aws-native/repo/.github/workflows/release.yml
```diff
@@ -564,7 +564,7 @@ jobs:
           !*.tar.gz
   dispatch_docs_build:
     runs-on: ubuntu-latest
-    needs: tag_sdk
+    needs: pubish_go_sdk
```
thomas11, line 567: typo

thomas11 (changes requested):

native-provider-ci/providers/aws-native/repo/.github/workflows/release.yml
danielrbradley: Noooo ... so this is actually working because the actual job name is wrong 🙃 ... will fix both

thomas11 (approved):

PR #933, opened 2024-05-17
Don't hit GitHub API for schema comparisons

blampe:
The latest `schema-tools` supports reading from disk instead of fetching from GitHub.

### DIFF
--- a/native-provider-ci/src/steps.ts
+++ b/native-provider-ci/src/steps.ts
@@ -324,6 +324,7 @@ export function InstallSchemaChecker(provider: string): Step {
     uses: action.installGhRelease,
     with: {
       repo: "pulumi/schema-tools",
+      tag: "master",
     },
   };
 }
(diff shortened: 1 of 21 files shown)


native-provider-ci/src/steps.ts
```diff
@@ -553,12 +553,10 @@ export function CheckSchemaChanges(provider: string): Step {
     if: "github.event_name == 'pull_request'",
     name: "Check Schema is Valid",
     run:
+      "git show ${{ github.event.repository.default_branch }}:provider/cmd/pulumi-resource-${{ env.PROVIDER }}/schema.json > ${{ runner.temp }}/schema.json;\n" +
```
danielrbradley, line 556: Will this work when we only have a shallow clone of the repository? Does the shallow clone include the default_branch?

danielrbradley (commented):
As of recently, we've stopped checking out the entire repository as this is a slow operation for some providers to do in every step. We need to ensure this change works with a shallow clone as downloading the whole git history is slower than fetching the single file.

danielrbradley (changes requested):
As an alternative, I'd suggest adding the complexity to the schema-tools to check if the base is available in git directly, then falling back to the github API if it's a shallow clone.

Issue #936, opened 2024-05-24
Prepare provider-ci for non-bridged providers


danielrbradley:
## Long term aims

- Remove duplication of provider-ci and native-provider-ci
- Reduce the cost & risk of changing CI.
- Resolve fragmentation of CI & makefile features.

## Refactoring goals

- Re-use specific parts of the bridged-provider template.
- Make templates composable - so `bridged-provider` or `go-provider` can be composed of smaller templates.

### Make it simpler to run the provider-ci binary directly

Make the 4 required arguments to `provider-ci` optional:
1. `--config` - default to `.ci-mgmt.yaml`.
2. `--out` - default to the current directory ('.')
3. `--name` - default to the config `.provider` field.
4. `--template` - add `template` to the config file

### Simplify configuration

Remove default `env` configuration. Put the common keys into a splice instead of in the default config.
- Add tool versions and just append any additional provider-specified env.
- Remove obsolete keys e.g. `TRAVIS_OS_NAME`
- [Search shows](https://github.com/search?q=org%3Apulumi+path%3A.ci-mgmt.yaml+env&type=code&p=3) that providers only ever add to the ENV config but never override existing values. Only exception being [pulumi-docker setting PULUMI_GO_DEP_ROOT to the same path, but differently](https://github.com/pulumi/pulumi-docker/blame/4311d91410cabd8aba4308d1c945a903e38a9643/.ci-mgmt.yaml#L22C7-L22C7)

### Remove tool versions from env

We're generating all the files ourselves, so there's no reason to push the version into the `env` just to be read out when installing each tool. We can just put the version numbers directly into step template.

## Compatibility

We can use github search to investigate which parts of the configuration actually get used: https://github.com/search?q=org%3Apulumi+path%3A.ci-mgmt.yaml&type=code

PR #937, opened 2024-05-24
Make generate cli arguments optional

danielrbradley:
Related to https://github.com/pulumi/ci-mgmt/issues/936 - this is a pre-cursor to building out composable templates.

## Unbundle the config parsing from the generate method

We need to parse the config before calling generate.

## Make it simpler to run the provider-ci binary directly

Make the 4 required arguments to `provider-ci` optional:

```
Usage:
  provider-ci generate [flags]

Flags:
  -c, --config string     local config file to use (default ".ci-mgmt.yaml")
  -h, --help              help for generate
  -n, --name string       repository name to generate (default "{config.repository}" or otherwise "pulumi/pulumi-{config.provider}")
  -o, --out string        directory to write generate files to (default ".")
  -t, --template string   template name to generate (default "{config.template}" or otherwise "bridged-provider")
```

This avoids the need to generate a makefile template to make it easier to run the tool to pull templates:

```diff
- go run github.com/pulumi/ci-mgmt/provider-ci@master generate \
-		--name pulumi/pulumi-$(PACK) \
-		--out . \
-		--template bridged-provider \
-		--config $<
+ go run github.com/pulumi/ci-mgmt/provider-ci@master generate
```

Existing providers have all the required information available except for the `template` field.

### DIFF
--- a/provider-ci/internal/cmd/generate.go
+++ b/provider-ci/internal/cmd/generate.go
@@ -1,22 +1,63 @@
 package cmd
 
 import (
+	"fmt"
+
 	"github.com/pulumi/ci-mgmt/provider-ci/internal/pkg"
 	"github.com/spf13/cobra"
 )
 
-var generateOpts = &pkg.GenerateOpts{}
+type generateArguments struct {
+	RepositoryName string
+	OutDir         string
+	TemplateName   string
+	ConfigPath     string
+}
+
+var generateArgs generateArguments
 
 // generateCmd represents the generate command
 var generateCmd = &cobra.Command{
 	Use:   "generate",
 	Short: "Generate repository files.",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		err := pkg.GeneratePackage(pkg.GenerateOpts{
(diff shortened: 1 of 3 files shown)


iwahbe (commented):
This feels like two changes:

- Make CI args optional.
- Use a single config file.

It would be easier to review (and cleaner to decide on) if they were separate PRs.

---

> Change the original design from having a default config per template to a single global default config. This is because:

> 1. The config should be a minimal collection of possible keys.
> 2. The config keys needs to not clash between different sub-templates.

I agree with (1).

I don't understand (2). It sounds like you are proposing a unified namespace, where each template pics and chooses from it. Is that correct?

Does this design keep the restriction that only 1 template can be used at a time? If so, I don't see the of (2).

danielrbradley:
> This feels like two changes:
> 
> * Make CI args optional.
> * Use a single config file.
> 
> It would be easier to review (and cleaner to decide on) if they were separate PRs.

@iwahbe this is now done - this PR only makes the arguments optional.

flostadler (approved):
LGTM

Issue #594, opened 2023-09-19
Consider quoting all provider names


guineveresaenger:
We encountered an issue with `pulumi-null`'s provider name being evaluated to the null value in Actions.

https://github.com/pulumi/ci-mgmt/pull/592 is attempting to fix that; however we should probably quote all provider names in our Actions as per [code review suggestion](https://github.com/pulumi/ci-mgmt/pull/592#discussion_r1330337342).

danielrbradley:
Fixed as part of https://github.com/pulumi/ci-mgmt/pull/943

PR #943, opened 2024-05-29
Simplify tool version & env config

danielrbradley:
See commits for impact of individual changes.

Part of https://github.com/pulumi/ci-mgmt/issues/936 - attempting to simplify our shared config ahead of onboarding native providers with the same shared config but different templates.

## Remove `actions.setupGo` and `actions.setupPulumi` from the default config

These are never overridden by providers, we were only using the config file for templating. This makes these two actions match with how we write all the other `setupX` actions - inline, with config variables for the action version and tool version.

This helps reduce the number of possible configuration variations we need to consider when making changes.

## Remove tool versions from ENV

We only set them as ENV vars to be able to access them as variables in the GH action script. Now we're templating everything, we can just set these values inline to remove a layer of indirection and reduce the number of global mutable variables which can be hard to trace how they're used.

## Remove PROVIDER env

Similar to the tool versions above, this was only used in the templates, not as actual ENV vars. In addition, this is confusing because the makefile also has a variable called PROVIDER which shadows this ENV var, but contains the name of the provider binary. There's no other obvious references to this ENV variable.

## Remove TRAVIS_OS_NAME env

This was previously used by the [pulumi/scripts publish-tfgen-package script](https://github.com/pulumi/scripts/blob/9aeffd86afa793846dc4f001a6f374b3f135bfef/ci/publish-tfgen-package#L29)

We now use the https://github.com/pulumi/pulumi-package-publisher action which does not require this variable to be set.


provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/command-dispatch.yml
```diff
@@ -1,11 +1,6 @@
 # WARNING: This file is autogenerated - changes will be overwritten if not made via https://github.com/pulumi/ci-mgmt
 
 env:
-  #{{- if eq .Config.provider "null" }}#
-  PROVIDER: "#{{ .Config.provider }}#"
-  #{{- else }}#
-  PROVIDER: #{{ .Config.provider }}#
-  #{{- end }}#
```
guineveresaenger, line 8: Have we verified that this change will generate a string for `pulumi-null` provider? GHA will fail if it's not quoted.

It also occurs to me that this adjustment will no longer work because we need to be able to interpolate `.Config.provider` and this PR combines `PROVIDER` into `.Config.provider`.

See also: https://github.com/pulumi/ci-mgmt/issues/594 which we can close if we find a satisfactory adjustment for pulumi-null.

provider-ci/internal/pkg/templates/bridged-provider.config.yaml
```diff
@@ -28,28 +29,14 @@ env:
...
-      with:
-        pulumi-version: "^3"
-  setupGo:
-    - name: Install Go
-      uses: actions/setup-go@v5
-      with:
-        go-version: 1.21.x
-        cache-dependency-path: |
-          sdk/go.sum
+actions: {}
```
guineveresaenger, line 39: Could we get rid of this field then?

guineveresaenger (commented):
This looks pretty great. I have one concern about pulumi-null, see comments.

I would advise to merge this with all commits intact, not a squash commit. THe individual commit descriptions are great information to keep.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/command-dispatch.yml
danielrbradley: Yup, I've reviewed all usages of the `.Config.provider` for where we need quoting. Everywhere we use it now in YAML, we're combining it with a prefix or suffix, so it's never just setting the value "null" in YAML.

Examples:
- Quoted within a command: `find ${{ github.workspace }} -name "pulumi-*-#{{ .Config.provider }}#" -print -exec chmod +x {} \;`
- Unquoted but suffixed: `name: #{{ .Config.provider }}#-provider.tar.gz`
- Unquoted but suffixed: `role-session-name: #{{ .Config.provider }}#@githubActions`
- Unquoted within command, and prefixed: `schema-tools compare -p #{{ .Config.provider }}# -o ${{ github.event.repository.default_branch }} -n --local-path=provider/cmd/pulumi-resource-#{{ .Config.provider }}#/schema.json` (arguably should quote all arguments in commands)
- Interpolated in command: `jq -r '.config | select(.variables) | .variables | keys[]' < provider/cmd/pulumi-resource-#{{ .Config.provider }}#/schema.json >> keys.txt`
- Prefixed: `pulumi-tfgen-#{{ .Config.provider }}#`
- Prefixed in command: `run: pulumictl create docs-build pulumi-#{{ .Config.provider }}#`
- Interpolated into string: `commit-message: Resync build for pulumi-#{{ .Config.provider }}#`
- Interpolated into string: `title: Fix up build for pulumi-#{{ .Config.provider }}#`

provider-ci/internal/pkg/templates/bridged-provider.config.yaml
danielrbradley: Removing this breaks some of the templates. There's two optional properties which are document by the comment below: `actions.preBuild` and `actions.preTest`.

It seems like the templates assume that the parent `actions` container is available because removing it causes the error:

```
Error: error rendering template templates/bridged-provider/.github/workflows/prerequisites.splice: template: templates/bridged-provider/.github/workflows/prerequisites.splice:41:15: executing "templates/bridged-provider/.github/workflows/prerequisites.splice" at <.Config.actions.preBuild>: nil pointer evaluating interface {}.preBuild
```

Leaving just the `{}` seems to fix this 🤷

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
```diff
@@ -45,13 +40,21 @@ jobs:
         aws-access-key-id: ${{ secrets.AWS_CORP_S3_UPLOAD_ACCESS_KEY_ID }}
         aws-region: us-west-2
         aws-secret-access-key: ${{ secrets.AWS_CORP_S3_UPLOAD_SECRET_ACCESS_KEY }}
-#{{ .Config.actions.setupGo | toYaml | indent 4 }}#
+    - name: Install Go
+      uses: #{{ .Config.actionVersions.setupGo }}#
+      with:
+        go-version: "#{{ .Config.toolVersions.go }}#"
+        cache-dependency-path: |
+          sdk/go.sum
```
blampe, line 48: This should probably be `**/go.sum` to include the provider's dependencies in the cache key.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
```diff
@@ -45,13 +40,21 @@ jobs:
         aws-access-key-id: ${{ secrets.AWS_CORP_S3_UPLOAD_ACCESS_KEY_ID }}
         aws-region: us-west-2
         aws-secret-access-key: ${{ secrets.AWS_CORP_S3_UPLOAD_SECRET_ACCESS_KEY }}
-#{{ .Config.actions.setupGo | toYaml | indent 4 }}#
+    - name: Install Go
+      uses: #{{ .Config.actionVersions.setupGo }}#
+      with:
+        go-version: "#{{ .Config.toolVersions.go }}#"
```
blampe, line 46: It would be better to specify `go-version-file: provider/go.mod` here, with a config option to override that path. Otherwise we're installing this global version only to re-install the version specified in the toolchain.

provider-ci/internal/pkg/templates/bridged-provider.config.yaml
```diff
@@ -10,15 +10,16 @@ fail-on-extra-mapping: true
 publishRegistry: true
 checkoutSubmodules: false
 testMasterAndReleaseWorkflows: false
+toolVersions:
+  dotnet: "6.0.x"
+  go: "1.21.x"
+  java: "11"
+  gradle: "7.6"
+  node: "20.x"
+  pulumi: "^3"
```
blampe, line 19: Why are we letting the CLI version float?

danielrbradley: We always want to be testing latest. This is just moved from [line 45](https://github.com/pulumi/ci-mgmt/pull/943/files/864cf32041d97cb75232edfa278748945fc61797#diff-fbd139b766b0a06128fb3f2ccdc8a50a4058c8bf7dea78ce1e5890c6330ce8aaL45)

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
danielrbradley: Yeah, that would probably make sense to use that as the source of truth. This change is just tidying the config file. This is taken directly from the config [lines 47-52](https://github.com/pulumi/ci-mgmt/pull/943/files/864cf32041d97cb75232edfa278748945fc61797#diff-fbd139b766b0a06128fb3f2ccdc8a50a4058c8bf7dea78ce1e5890c6330ce8aaL47-L52)

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
danielrbradley: Quite possibly. I'm not sure why only the sdk has been selected previously. Looks like it was added by @iwahbe in https://github.com/pulumi/ci-mgmt/pull/521 with the comment:

> I have borrowed the caching strategy from pulumi-aws. Instead of repeating the same block in 29 places, I have defined the action in "bridged-provider.config.yaml" and then
templated it out from there.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/command-dispatch.yml
guineveresaenger: Excellent, thank you for looking into this!

provider-ci/internal/pkg/templates/bridged-provider.config.yaml
guineveresaenger: right. AWS being special again...

guineveresaenger (approved):
LGTM.

@blampe, I believe we can tackle the sdk path concerns in a followup, since this is a refactor/cleanup?

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
iwahbe: I don't remember why it was `sdk/go.sum`. Don't block on me.

PR #945, opened 2024-05-30
Fix `make upstream.rebase` documentation

iwahbe:
In effect `make upstream` (`apply` in `upstream.sh`) is called as the final step in `make upstream.finalize`. Because `make upstream` resets the upstream submodule to the currently staged commit, if `TO` is set to any value that is not HEAD it will be ignored (but look like it worked).


guineveresaenger (approved):
Thank you! 😃

PR #947, opened 2024-05-30
Only require new sentinel check

danielrbradley:
Depends on #944 and #946 being fully rolled out before this is run.

Once this is deployed, we will be free to restructure workflow jobs without needing to worry about breaking the required checks.

### DIFF
--- a/infra/providers/index.ts
+++ b/infra/providers/index.ts
@@ -15,29 +15,7 @@ function hasManagedBranchProtection(provider: string): boolean {
 function defineResources(buildSdkJobName: string, provider: string) {
     const requiredChecks: string[] = [
         "Update Changelog",
-
-        "prerequisites",
-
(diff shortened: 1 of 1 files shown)


infra/providers/index.ts
```diff
@@ -15,29 +15,7 @@ function hasManagedBranchProtection(provider: string): boolean {
...
-        buildSdkJobName + " (dotnet)",
-        buildSdkJobName + " (go)",
-        buildSdkJobName + " (java)",
-        buildSdkJobName + " (nodejs)",
-        buildSdkJobName + " (python)",
-        "test (dotnet)",
-        "test (go)",
-        "test (java)",
-        "test (nodejs)",
-        "test (python)",
```
blampe, line 38: I love that this makes room for more arbitrary test sharding.

blampe (approved):

PR #949, opened 2024-05-30
Remove the redundant branch protection

iwahbe:
The `github` provider knows how to look up the default branch for a repository. This avoids creating 2 branch protection rules for every repo.

### DIFF
--- a/infra/providers/index.ts
+++ b/infra/providers/index.ts
@@ -12,7 +12,7 @@ function hasManagedBranchProtection(provider: string): boolean {
     return !provider.includes("azure-native");
 }
 
-function defineResources(buildSdkJobName: string, provider: string) {
+function nativeProviderProtection(buildSdkJobName: string, provider: string) {
     const requiredChecks: string[] = [
         "Update Changelog",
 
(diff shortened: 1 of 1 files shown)


iwahbe:
Merging this will require some alias work. I'll do that after this PR gets approval but before the PR merges.

iwahbe:
I'm working on aliases now so we don't delete and then re-create branch protection that we want.

guineveresaenger (approved):
This looks like a good cleanup to me.

Can we roll this out on one provider as a canary?

infra/providers/index.ts
```diff
@@ -73,36 +73,31 @@ function tfProviderProtection(provider: string) {
...
-            enforceAdmins: true,
-            requiredStatusChecks: [{
-                strict: false,
-                contexts: requiredChecks,
-            }],
-            requiredPullRequestReviews: [{
-                // We want to make sure that pulumi-bot can auto-merge PRs, so we
-                // explicitly remove review requirements.
-                requiredApprovingReviewCount: 0,
+    const repo = `pulumi-${provider}`;
```
mjeffryes, line 76: nit. we could do the same thing for the native providers

infra/providers/index.ts
```diff
@@ -73,36 +73,31 @@ function tfProviderProtection(provider: string) {
...
+            requiredApprovingReviewCount: 0,
+
+        }],
+    }, {
+        deleteBeforeReplace: true,
+    })
 }
 
-for (let bridgedProvider of [...tfProviders].filter(hasManagedBranchProtection)) {
+for (let bridgedProvider of tfProviders) {
```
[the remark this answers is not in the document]
iwahbe: The filter currently looks like this:

https://github.com/pulumi/ci-mgmt/blob/9c9517b3e1d49367b1c666c29a965d1a939934c8/infra/providers/index.ts#L9-L13

In general, bridged providers *should* use the workflows we define.

infra/providers/index.ts
iwahbe: I'd like to do that as a separate PR. I'll follow up after this merges.

iwahbe:
I'll try to roll this out to one provider with `--target` and manually verify that it worked as expected.

iwahbe:
> > Can we roll this out on one provider as a canary?
> 
> I'll try to roll this out to one provider with `--target` and manually verify that it worked as expected.

Deploying with `pu up --target **aiven**` deletes only the unused branch as expected.

PR #951, opened 2024-05-30
Cleanup branch protection aliases for bridged providers

iwahbe:
Followup on #949: Remove redundant branch protections from native providers.

### DIFF
--- a/infra/providers/index.ts
+++ b/infra/providers/index.ts
@@ -87,14 +87,9 @@ function tfProviderProtection(provider: string) {
             // We want to make sure that pulumi-bot can auto-merge PRs, so we
             // explicitly remove review requirements.
             requiredApprovingReviewCount: 0,
-
         }],
     }, {
         deleteBeforeReplace: true,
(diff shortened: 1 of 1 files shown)


mjeffryes (approved):

guineveresaenger (approved):
nit: can we mention we're also doing an alias operation on native providers here?

infra/providers/index.ts
```diff
@@ -39,31 +39,29 @@ function nativeProviderProtection(buildSdkJobName: string, provider: string) {
...
+            requiredApprovingReviewCount: 1,
+        }],
+        requiredStatusChecks: [{
+            strict: false,
+            contexts: requiredChecks,
+        }]
+    }, {
+        deleteBeforeReplace: true,
+        // TODO: Delete after pulumi up has happened.
+        aliases: [{name: branchAlias(provider)}],
```
iwahbe, line 63: > nit: can we mention we're also doing an alias operation on native providers here?

We are doing an alias operation on the native providers here.

PR #953, opened 2024-05-30
Clarify the usage of blocked vs upstream

iwahbe:
The difference between the `blocked` and the `awaiting-upstream` issue isn't universally clear. From
slack ([internal](https://pulumi.slack.com/archives/C02FXTZEZ6W/p1716474534804139)), it sounds like we would like this breakdown:

- `blocked`: The issue cannot be resolved without 3rd party action.
- `awaiting-upstream`: The issue cannot be resolved without action in another repository (which may or may not be owned by Pulumi).

Let's decide what we want these labels to mean and then make the description on the labels useful and authoritative.

### DIFF
--- a/infra/providers/index.ts
+++ b/infra/providers/index.ts
@@ -1,3 +1,4 @@
+import * as pulumi from "@pulumi/pulumi";
 import * as github from "@pulumi/github";
 import * as fs from 'fs';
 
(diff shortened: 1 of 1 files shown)


mjeffryes (approved):
Looks good to me assume this is where we decide to land it.

guineveresaenger:
I'm somewhat in favor of https://github.com/pulumi/team-management/pull/360, as project management is an internal process.

PR #954, opened 2024-05-31
Ensure release labels via pulumi

iwahbe:
This replaces the imperative script we currently rely on, which hasn't been run in a while and is currently out of date.

Stacked on top of https://github.com/pulumi/ci-mgmt/pull/953


infra/providers/index.ts
```diff
@@ -121,6 +121,18 @@ class ProviderLabels extends pulumi.ComponentResource {
     }
 }
 
+class BridgedProviderLabels extends ProviderLabels {
```
mjeffryes, line 124: it wouldn't hurt to just apply these to all provider repos. @danielrbradley is trying to get all our providers using ci-mgmt for their workflows, and I hope that not too far downstream of that we'll be able to adopt the one step release process for non-bridged providers.

mjeffryes (approved):

infra/providers/index.ts
iwahbe: I wouldn't want to apply these labels to native provider repos unless they also support one step releases. If I did, I think people (myself included) would assume that they could merge a PR with `needs-release/patch` and it would release on its own.

danielrbradley: What do we mean by one-step?

I don't think there's much difference in the release process right now.

danielrbradley: Is there a doc or a reference on how the needs-release label is implemented?

iwahbe: I think this is as close to a reference as you are going to get:

https://github.com/pulumi/ci-mgmt/blob/236437ea328a6f3dc4fbfa4924fc968fc619842a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml#L170-L185

iwahbe: Our automation tags a PR as `needs-release/minor` as part of an automated minor release. When we merge the PR with that label, it runs the release on its own.

mjeffryes: documented some in our [playbook](https://github.com/pulumi/platform-providers-team/blob/main/playbooks/Release.md#single-step-release) and on the action itself.

danielrbradley: Yeah, think we can replicate that to native providers. Don't see much tricky stuff there 👍

PR #958, opened 2024-06-03
Refactor: Use GHA native dispatch for registry metadata build kickoff

guineveresaenger:
This pull request removes the dependency on `pulumictl` for the bridged provider registry docs dispatch.

### Before this change
Currently, the [Create Docs Build](https://github.com/pulumi/ci-mgmt/blob/master/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/release.yml#L24) step in the Release Workflow calls into a [Pulumictl command](https://github.com/pulumi/pulumictl/blob/master/cmd/pulumictl/create/docs-build/cli.go) that triggers the [Publish Provider Update](https://github.com/pulumi/registry/blob/master/.github/workflows/publish-provider-update.yml) Workflow, building metadata, in pulumi/registry.

`publish.yml`(provider) -> `pulumictl`(bespoke command) -> `publish-provider-update`(registry)

### After this change
Use `peter-evans/repository-dispatch@v3` to dispatch the `resource-provider` event that [already exists on the registry `publish-provider-update` Workflow](https://github.com/pulumi/registry/blob/master/.github/workflows/publish-provider-update.yml#L16), removing the need for pulumictl.

The registry's Action has a few fields we do not use. See [example run](https://github.com/pulumi/registry/actions/runs/9350133653/job/25732951804#step:2:4). Filed https://github.com/pulumi/registry/issues/4671 to track. The repository dispatch payload therefore only contains:

```
"project": "${{ github.repository }}", - this is somewhat unnecessary but makes for a safe refactor without having to change the registry workflow file
"shortname": "#{{ .Config.provider }}#", - Just the provider name without `pulumi-` prefix
"ref": "${{ github.ref_name }}" - This is the tag name of the release.
```

*Note:* This pull request changes bridged providers only since we are currently working on unifying templating for native and bridged providers. I'm happy to add native provider changes to this effect if reviewers would prefer.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/release.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/release.yml
@@ -21,19 +21,18 @@ jobs:
     needs: tag_sdk
     runs-on: #{{ .Config.runner.default }}#
     steps:
-    - name: Dispatch Metadata build
-      uses: peter-evans/repository-dispatch@v3
-      with:
-    - env:
-        token: ${{ secrets.PULUMI_BOT_TOKEN }}
-        repository: pulumi/pulumi-${{ env.PROVIDER }}
-        event-type: resource-provider
-        client-payload: |-
-          {
-            "project": pulumi-${{ env.PROVIDER }},
-            "shortname": ${{ env.PROVIDER }},
-            "ref": ${{ env.GITHUB_REF_NAME }}
-          }
+      - name: Dispatch Metadata build
+        uses: peter-evans/repository-dispatch@v3
+        with:
+          token: ${{ secrets.PULUMI_BOT_TOKEN }}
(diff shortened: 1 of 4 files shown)


iwahbe (approved):
This looks extremely safe to me. Let's make sure that we have a test run to show before we merge for the whole fleet.

danielrbradley (approved):

guineveresaenger:
Update: [This ran correctly against pulumi-dbtcloud](https://github.com/pulumi/pulumi-dbtcloud/blob/b7080511d6930aa0f558ebb35eec5d6f21f7187f/.github/workflows/release.yml#L127): 
Registry build here: https://github.com/pulumi/registry/actions/runs/9455770701/job/26046268147

danielrbradley (approved):
Nice simplification!

PR #960, opened 2024-06-05
Set missing docs error to true in the Makefile

### LINKED ISSUE #959: Set `PULUMI_MISSING_DOCS_ERROR` to true in Makefile
We've seen a few `upgrade-provider` failures due to our `prerequisites` CI Workflow setting missing docs to "warning" instead of "error" like `upgrade-provider` does. So we sometimes see upgrade failures due to missing docs.

We should set `PULUMI_MISSING_DOCS_ERROR to `true` everywhere, for every provider - locally and in CI.

The best way to do this is by setting it in the Makefile.

A recent bridge upgrade test only failed on one provider (pulumi-openstack), so this is a safe change to make that should not result in lots of ops churn.

guineveresaenger:
Fixes #959.

Verified this works as expected via https://github.com/pulumi/pulumi-openstack/pull/604

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/Makefile
+++ b/provider-ci/internal/pkg/templates/bridged-provider/Makefile
@@ -22,6 +22,7 @@ PULUMI_CONVERT := 1
 #{{- else }}#
 PULUMI_CONVERT := 0
 #{{- end }}#
+PULUMI_MISSING_DOCS_ERROR := true
 
 # Override during CI using `make [TARGET] PROVIDER_VERSION=""` or by setting a PROVIDER_VERSION environment variable
 # Local & branch builds will just used this fixed default version unless specified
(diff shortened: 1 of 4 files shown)


provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -22,6 +22,7 @@ PULUMI_CONVERT := 1
 #{{- else }}#
 PULUMI_CONVERT := 0
 #{{- end }}#
+PULUMI_MISSING_DOCS_ERROR := true
```
iwahbe, line 25: Does this not work?
```suggestion
export PULUMI_MISSING_DOCS_ERROR := true
```

iwahbe (approved):
LGTM!

provider-ci/internal/pkg/templates/bridged-provider/Makefile
guineveresaenger: it might! I followed the pattern as established for consistency.

PR #964, opened 2024-06-10
Add option for freeing disk space

danielrbradley:
Stacked on top of #942 

- Provide dedicated option instead of relying on providers to use custom steps which are run after tool installs.
- Run as first step so we don't delete things that have just been installed.
- Default to off to maintain existing behaviour. This can take around 3 minutes to run, so we should only run it if we're having issues.

```yaml
# Set to true to clear disk space before running prerequisites workflow.
freeDiskSpaceBeforeBuild: false
```

This fixes oddities in Azure and AWS where we've added steps in the `preBuild` and `preTest` to clear disk space but then have to install tools (e.g. dotnet) a second time as they get deleted.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider.config.yaml
+++ b/provider-ci/internal/pkg/templates/bridged-provider.config.yaml
@@ -10,6 +10,10 @@ fail-on-extra-mapping: true
 publishRegistry: true
 checkoutSubmodules: false
 testMasterAndReleaseWorkflows: false
+# Set to true to clear disk space before running prerequisites workflow.
+freeDiskSpaceBeforeBuild: false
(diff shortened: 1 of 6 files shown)


iwahbe:
Could you change the target branch to #942's branch. That way we will see only the diffs that this PR introduces?

danielrbradley:
> Add option for freeing disk space

Planning to rebase tomorrow once #942 merges. The diff can be seen here (it's just the last commit): https://github.com/pulumi/ci-mgmt/pull/964/commits/a4090107f17c21a7b0297ec2a49b8729f5e30fdc

iwahbe (approved):

PR #965, opened 2024-06-10
Encapsulate tool setup into a composite action

danielrbradley:
Stacked on top of https://github.com/pulumi/ci-mgmt/pull/942

Primary benefit is to reduce duplication within templates. This was not possible when using splices because splices cannot reference splices.

Additionally, we want to make installing our standard versions of our tools available for use from custom test jobs as used in AWS and other providers so they don't have to hard code their own install actions.

- Go was not previously explicitly installed during the build_sdk job but is used when running `make build_go`, so this adds it to ensure we're using a consistent version.


iwahbe (approved):
This looks good to merge, but with the same caveats as https://github.com/pulumi/ci-mgmt/pull/942; it needs to have a full CI -> master -> release workflow test before deploying to the whole fleet.

danielrbradley:
Deploying to test via AWS which has custom workflows in its config: https://github.com/pulumi/ci-mgmt/actions/runs/9480279540/job/26120591175

PR #971, opened 2024-06-12
Add `awaiting/bridge` as a label to brided provider repos.

iwahbe:
This PR adds a label to each bridged provider repo: `awaiting/bridge`.

We can use this to better track provider bugs that should be fixed in the bridge.

### DIFF
--- a/infra/providers/index.ts
+++ b/infra/providers/index.ts
@@ -72,14 +72,26 @@ function tfProviderProtection(provider: string) {
     new BridgedProviderLabels(provider);
 }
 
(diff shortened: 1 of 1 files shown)


mjeffryes (approved):
We should maybe have one of these for awaiting/codegen, awaiting/core, etc. 1

t0yv0 (approved):
It's a bit weird to go and label closed issues with this but I might do that.. I'm looking for a way to run the numbers on the bug root-cause trends with the help of these labels.

PR #973, opened 2024-06-13
Install pulumi cli in jobs

VenelinMartinov:
We were not actually installing the CLI and just using the ambient GHA version = (

setup-tools expects pulumicli as a string: https://github.com/pulumi/pulumi-github/blob/43b283d3734a435d5c714723ab8c1ddefc2be3c3/.github/actions/setup-tools/action.yml#L38

We also now install the dev version

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider.config.yaml
+++ b/provider-ci/internal/pkg/templates/bridged-provider.config.yaml
@@ -20,7 +20,7 @@ toolVersions:
   java: "11"
   gradle: "7.6"
   node: "20.x"
-  pulumi: "^3"
+  pulumi: "dev"
   python: "3.11.8"
 env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Setup tools
         uses: ./.github/actions/setup-tools
         with:
-          tools: pulumictl, pulumi, go, node, dotnet, python, java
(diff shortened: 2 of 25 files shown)


VenelinMartinov:
Ok, this doesn't work as the pulumi install action just skips if there is a version installed: https://github.com/pulumi/pulumi-xyz/actions/runs/9504509471/job/26197318358?pr=185

danielrbradley (approved):
Bah, thanks!

danielrbradley:
> Ok, this doesn't work as the pulumi install action just skips if there is a version installed: https://github.com/pulumi/pulumi-xyz/actions/runs/9504509471/job/26197318358?pr=185

Oh odd .. has this always been the way?

Did this break anything?

danielrbradley:
Oh I see - this is the way it's always been, but because we're just asking for any version > 3.0.0, it's happy with the pre-installed version and skips. 

Looks like our choices here would be:
- Fix a specific version, but that would require regular roll-outs from ci-mgmt
- Set the version to `dev` to always pull the lastest dev release - so we would know about issues before they get released to the public.
- Add a new option in the action for `latest` to always fetch the latest released version.

VenelinMartinov:
now works: https://github.com/pulumi/pulumi-xyz/actions/runs/9505457571/job/26200388934?pr=186

danielrbradley (approved):

t0yv0:
Is there something that prints the CLI version if I were to try to reproduce a failing test?  I'm generally anti-floating verisions but I realize the version was floating already. Perhaps having versions pinned is too much churn in the repo source history. But what I really would like to have is a reliable way to repro a failing test from CI down to local without too much worry about not having the right environment.

VenelinMartinov:
@t0yv0 yeah, the "Setup tools" job prints the CLI version: https://github.com/pulumi/pulumi-xyz/actions/runs/9505457571/job/26200388934?pr=186

PR #975, opened 2024-06-14
Add env for build_sdk and prerequisites

danielrbradley:
Explicitly generate the env for re-usable workflows.

> Any environment variables set in an env context defined at the workflow level in the caller workflow are not propagated to the called workflow.

See: https://docs.github.com/en/actions/using-workflows/reusing-workflows#limitations

This was causing GCP's rollout to fail because it uses environment variables during its unit testing.

[Example failure](https://github.com/pulumi/pulumi-gcp/actions/runs/9477202631/job/26111336756#step:14:312)


VenelinMartinov (approved):
lgtm

iwahbe (approved):

PR #977, opened 2024-06-18
Verify release action

### LINKED ISSUE #972: Post-release test
Ensure the provider version (SDKs and provider binary) we've just released is usable.

- Pick a trivial example which is fast to exercise.
- Just run preview initially to ensure the SDKs and provider are retrievable & executable without error.
- Use a matrix over platform/architecture & language (25 jobs):
  - Platform/arch:
    -  macos-x64
    - macos-arm64
    - linuxstatic-x64
    - linux-arm64
    - win-x64
  - SDK language:
    - node
    - go
    - dotnet
    - python
    - java
- Run after any release - alpha, beta or stable.

danielrbradley:
Add new job in our release workflows to verify that our released SDKs and provider are usable.

This will have to be configured into each provider as we will need a program specified in the local repo for each runtime we want to verify. Ideally this will include: nodejs, python, dotnet and Go. Without configuration, the job will not do any verification. There's an example of the available configuration as documentation in the `bridged-provider.config.yaml`.

The verify-release workflow is also runnable directly if we want to do a one-off verification. See the [pulumi/verify-provider-release](https://github.com/pulumi/verify-provider-release) action for the underlying implementation.

Fixes #972 

[Example run without configuration](https://github.com/pulumi/pulumi-xyz/actions/runs/9604446619) skips the job:

image

[Example prerelease with nodejs configuration](https://github.com/pulumi/pulumi-xyz/actions/runs/9604468261) runs on Ubuntu and Windows:

image

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/verify-release.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/verify-release.yml
@@ -29,6 +29,16 @@ env:
 jobs:
   verify-release:
     name: verify-release
+#{{- if not .Config.releaseVerification }}#
+    # We don't have any release verification configurations, so we never run this workflow.
+    # Configure your .ci-mgmt.yaml files to include the release verification configurations e.g.
+    # releaseVerification:
+    #   nodejs: path/to/nodejs/project
+    #   python: path/to/python/project
+    #   dotnet: path/to/dotnet/project
+    #   go: path/to/go/project
+    if: false
+#{{- end }}#
     strategy:
       matrix:
 #{{- if .Config.releaseVerification }}#
@@ -55,8 +65,6 @@ jobs:
           directory: #{{ .Config.releaseVerification.nodejs }}#
           provider: #{{ .Config.provider }}#
           providerVersion: ${{ inputs.providerVersion }}
-#{{- else }}#
-      - run: echo "::warning file=.ci-mgmt.yaml,title=Node.js release verification missing::Add the key releaseVerification.nodejs to .ci-mgmt.yaml pointing to the directory containing a Pulumi nodejs project to preview."
 #{{- end }}#
 #{{- if .Config.releaseVerification.python }}#
       - name: Verify python release
@@ -66,8 +74,6 @@ jobs:
           directory: #{{ .Config.releaseVerification.python }}#
           provider: #{{ .Config.provider }}#
           providerVersion: ${{ inputs.providerVersion }}
-#{{- else }}#
-      - run: echo "::warning file=.ci-mgmt.yaml,title=Python release verification missing::Add the key releaseVerification.python to .ci-mgmt.yaml pointing to the directory containing a Pulumi python project to preview."
 #{{- end }}#
 #{{- if .Config.releaseVerification.dotnet }}#
       - name: Verify dotnet release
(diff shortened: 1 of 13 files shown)


AaronFriel:
About how many CI minutes of macOS runner time would be incurred weekly from running verify release? 

I imagine we can estimate # of releases weekly * length of verification.

I ask because we have a maximum concurrency of 5 macOS runners on OSS repositories and this could result in CI brownouts for pulumi/pulumi.

Cc @mjeffryes

mjeffryes:
This looks like it adds verification after prerelease, so we'd be looking at num_provider_commits * verification_time. We closed 866 PRs in providers last week, most probably were merged and ran prerelease.  Even if verification time is < 2min, it's going to burn a fair amount. We could potentially only test windows and mac on a full release to relive some pressure there.

danielrbradley:
I'll disable verifying pre-releases on Mac for the time being - only verify them on Mac after stable releases. This should suffice but still provide fast feedback on any issues.

corymhall (approved):
🚀

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/verify-release.yml
```diff
@@ -0,0 +1,93 @@
...
+        default: false
+
+env:
+#{{ .Config.env | toYaml | indent 2 }}#
+
+jobs:
+  verify-release:
+    name: verify-release
+    strategy:
+      matrix:
```
iwahbe, line 33: It feels like we could expand this matrix to also include language. That would give us parallelism and reduce repeated code.

You might need to build a simple lookup table to map `#{{ .Config.releaseVerification.nodejs }}#` to `nodejs`, perhaps using GH action outputs.

iwahbe (commented):
What's the rollout plan here? @danielrbradley will you roll this out to all providers? To tier 1 providers?

Do we intend to roll this out to all providers eventually? If so, we will need to budget for that (not all providers have tests at all right now). If not, then we probably shouldn't warn when a test is missing:

```
    - run: echo "::warning file=.ci-mgmt.yaml,title=Python release verification missing::Add the key releaseVerification.python to .ci-mgmt.yaml pointing to the directory containing a Pulumi python project to preview."
```

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/verify-release.yml
danielrbradley: I had considered putting the runtime in the matrix, but given we're only doing a preview, this is a very fast run - most of the time is spent just installing dependencies. We could tweak the way it runs so we still report all the tests even if earlier steps fail if we like.

danielrbradley:
> What's the rollout plan here? @danielrbradley will you roll this out to all providers? To tier 1 providers?

Initially, this is just rolling out the machinery to be able to start implementing this in each provider. Tier 1 are obviously the priority, but this should really be implemented for all providers given we're not actually deploying any real cloud resources - only validating the SDK and provider setup.

If the warnings are too messy then we could tidy these up a bit into a single warning perhaps, but it would make for more messy ci-mgmt code.

iwahbe:
I'd prefer to not introduce a class of warnings unless we plan to address them (immediately), since it builds the habit of ignoring warnings. That said, we already ignore plenty of build warnings, so it's probably not that much worse here. We can always disable them later if we need to.

danielrbradley:
@iwahbe I've updated the code to completely skip the verify-release job if it's not implemented instead of the warnings

PR #981, opened 2024-06-20
Add xyz as a provider

iwahbe:
This will opt xyz into more of our automation.

### DIFF
--- a/provider-ci/providers.json
+++ b/provider-ci/providers.json
@@ -70,5 +70,6 @@
     "vault",
     "venafi",
     "vsphere",
-    "wavefront"
+    "wavefront",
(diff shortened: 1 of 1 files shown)


provider-ci/providers.json
```diff
@@ -71,4 +71,5 @@
     "venafi",
     "vsphere",
     "wavefront"
+    "xyz"
```
blampe, line 74: Missing a comma above.

blampe (approved):
Stamping to unblock.

provider-ci/providers.json
iwahbe: Thanks!

PR #983, opened 2024-06-21
Allow empty patch in upstream script

guineveresaenger:
This pull request allows us to have an empty patch file in such cases as https://github.com/pulumi/pulumi-pagerduty/pull/543, where all we need the submodule for is to make up for a lack of proper Go module versioning in an upstream provider.
This change is a minimal change to the script and setup, and we won't have to add additinal logic to ci-mgmt.yaml configs.

- **Allow empty patch for upstream apply**
- **Make test workflows**

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/scripts/upstream.sh
+++ b/provider-ci/internal/pkg/templates/bridged-provider/scripts/upstream.sh
@@ -91,7 +91,7 @@ start_rebase() {
 
   for patch in ../patches/*.patch; do
     echo "Applying $patch"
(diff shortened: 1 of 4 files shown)


iwahbe (approved):

iwahbe:
Please do a test run to ensure that `--allow-empty` works as expected. Otherwise LGTM.

guineveresaenger:
It's already applied in pulumi-vault and pulumi-pagerduty (linked above). A local run against pulumi-gcp (which has non-empty patches) shows `make upstream` works as expected. Merging.

PR #988, opened 2024-06-24
Revert "Allow empty patch in upstream script (#983)"

flostadler:
This reverts commit 13b894e0e3ea580291d037ee7dbd563fb57b67df.

This introduced a regression that caused `make upstream.rebase` to fail in AWS, GCP and other providers.

E.g. AWS:
```
Applying ../patches/0001-Add-TagsSchemaTrulyComputed-definition.patch
fatal: Resolve operation not in progress, we are not resuming
```

E.g. GCP:
```
Applying ../patches/0001-Allow-disabling-the-partner-name.patch
fatal: Resolve operation not in progress, we are not resuming.
```


thomas11 (approved):

flostadler:
CC @guineveresaenger, we had to revert this one because it blocked us from upgrading providers

PR #990, opened 2024-06-24
Use pulumi cli version from `.pulumi.version`

corymhall:
The providers have a `.pulumi.version` file which stores the current pulumi version. There is also a job `weekly-pulumi-update.yml` that updates this version.

In our CI jobs we install the pulumi cli using the `pulumi/actions` action which uses the [version installed on `ubuntu-latest`](https://github.com/actions/runner-images/blob/c5c9a7cd7b2cfd20d740f7667fdaca8a412f2b52/images/ubuntu/Ubuntu2404-Readme.md?plain=1#L92) by default. This means that we are not testing on the version that we want to.

This introduces an additional step which sets the pulumi version to download based on the value from that file.

I've tested this strategy on [this PR](https://github.com/pulumi/pulumi-aws-native/pull/1594) in `aws-native` and it seems to be working.


danielrbradley (commented):
Really love the idea of making local tooling and CI more consistent and reducing repetition!

Can we include bridged providers with this change too (or a second PR)?

Could we just integrate looking for this file at the root of the repo into https://github.com/pulumi/actions? Or passing the path to a file which has the version to use as an input to the action? Kinda following the pattern of the "Setup Node" action: https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file

corymhall:
> Could we just integrate looking for this file at the root of the repo into [pulumi/actions](https://github.com/pulumi/actions)? Or passing the path to a file which has the version to use as an input to the action? Kinda following the pattern of the "Setup Node" action: [actions/setup-node@`main`/docs/advanced-usage.md#node-version-file](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md?rgh-link-date=2024-06-24T19%3A05%3A08Z#node-version-file)

I like that idea and it makes more sense. I'll try creating a PR over there instead!

PR #993, opened 2024-06-25
Simplify provider-ci makefile

danielrbradley:
Rewrite example provider testing targets to be more linear and easy to follow.

- Move "test-workflows" to "test-providers".
- Move config files for testing into each provider repo - named .ci-mgmt.yaml - just like the real repo.
- Move temp linting repo into the bin directory which is fully ignored as it's just temp files.

This is now slightly faster for running both clean and repeated runs of `make`. 

Clean run on `make`:
- Old: `2.40s user 2.76s system 159% cpu 3.224 total`
- New: `1.39s user 1.77s system 127% cpu 2.480 total`

Repeated run of `make`:
- Old: `1.63s user 1.99s system 231% cpu 1.561 total`
- New: `0.63s user 0.95s system 243% cpu 0.649 total`


.github/workflows/pull-request.yml
```diff
@@ -41,10 +41,13 @@ jobs:
         uses: jaxxstorm/action-install-gh-release@v1.11.0
         with:
           repo: pulumi/pulumictl
+      - name: Configure git
+        # Set the default branch to silence the warnings about the default branch name changing
+        run: git config --global init.defaultBranch master
```
thomas11, line 46: Some Pulumi repos use `main`, although probably no providers (yet)?

.github/workflows/pull-request.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         with:
           repo: rhysd/actionlint
       - name: lint providers
-        run: cd provider-ci && make lint-providers
+        run: cd provider-ci && make test-providers
```
thomas11, line 26: Is this correct, running the same target for linting than in line 50 for testing?

provider-ci/Makefile
```diff
@@ -17,69 +19,42 @@ $(ACTIONLINT):
...
+
+test-providers: test-provider/aws test-provider/docker test-provider/cloudflare
+
+# 1. Delete all files except the .ci-mgmt.yaml file and run the provider-ci generate command.
+# 2. Copy the generated provider repository to a temporary git repo and run actionlint on it.
+test-provider/%: bin/provider-ci $(ACTIONLINT)
+	cd test-providers/$* && \
+		find . -type f ! -name '.ci-mgmt.yaml' -delete && \
+		../../bin/provider-ci generate --name pulumi/pulumi-$*
+	rm -rf bin/test-provider/$*/repo
```
thomas11, line 48: I don't know `make`'s variable support well but extracting `bin/test-provider/$*/repo` into a variable would help readability here.

provider-ci/Makefile
```diff
@@ -17,69 +19,42 @@ $(ACTIONLINT):
...
+.PHONY: test-providers test-provider/%
+
+test-providers: test-provider/aws test-provider/docker test-provider/cloudflare
+
+# 1. Delete all files except the .ci-mgmt.yaml file and run the provider-ci generate command.
+# 2. Copy the generated provider repository to a temporary git repo and run actionlint on it.
+test-provider/%: bin/provider-ci $(ACTIONLINT)
+	cd test-providers/$* && \
+		find . -type f ! -name '.ci-mgmt.yaml' -delete && \
+		../../bin/provider-ci generate --name pulumi/pulumi-$*
```
thomas11, line 47: Deleting whole folders would be faster but I guess it doesn't matter much here.

provider-ci/Makefile
```diff
@@ -17,69 +19,42 @@ $(ACTIONLINT):
...
+	rm -rf bin/test-provider/$*/repo
+	mkdir -p bin/test-provider/$*/repo
+	cp -r test-providers/$* bin/test-provider/$*/repo
+	cd bin/test-provider/$*/repo &&	git init
+	cd bin/test-provider/$*/repo &&	../../../../$(ACTIONLINT) -config-file ../../../../actionlint.yml
+	rm -rf bin/test-provider/$*/repo/.git
+
+# Fetch the latest .ci-mgmt.yaml from the provider repositories ready for testing.
+update-provider-configs:
```
thomas11, line 56: Who/what will be calling this target, and when?

.github/workflows/pull-request.yml
danielrbradley: This is only used for creating a temporary repo so actionlint finds the right files. Will add a comment explaining.

.github/workflows/pull-request.yml
danielrbradley: Yeah, don't know why we run this twice - the old targets were different names but both aliased to the same target 🤷‍♂️

provider-ci/Makefile
danielrbradley: Just whenever we want to update our test data - would be manual right now but could create a CI job to call this. It seems like this hasn't been done in quite a while - see the follow-up PR stacked on this one.

.github/workflows/pull-request.yml
danielrbradley: Yeah, but probably only talking about ms here as it's only the generated files rather than the whole repo - so around 20 files to delete.

.github/workflows/pull-request.yml
thomas11: Can we just remove the linting one then?

thomas11 (approved):

provider-ci/Makefile
danielrbradley: Yeah, but probably only talking about ms here as it's only the generated files rather than the whole repo - so around 20 files to delete.

.github/workflows/pull-request.yml
danielrbradley: I've done a rework of this workflow - splitting it into 3: 
1. action lint (checking this repo's own workflows)
2. native-provider-ci
3. provider-ci

I've then de-duplicated the steps within each.

PR #994, opened 2024-06-25
Update provider configs

danielrbradley:
Update the test configs to match the real repos by running `make update-provider-configs`, then re-run `make` to re-generate.

Stacked on #993


iwahbe (approved):
I don't think this is necessary, but it doesn't hurt.

danielrbradley:
> I don't think this is necessary, but it doesn't hurt.

This caught me out a little time ago when assuming the config for aws matched the real provider, but then found it was using a different configuration that then when the PR was opened. More realistic test data is probably helpful here. We could also include some fake provider configurations if there's some specific combinations of features we want to have tested.

PR #995, opened 2024-06-25
Add `pulumi-version-file` to native providers

corymhall:
This adds the new `pulumi-version-file` support to native providers (except for google which does not have a .pulumi.version file)


danielrbradley (approved):
Google Native is an odd case as we've also disabled pushing changes out to it. Perhaps we should consider just removing it from this tooling if it would allow us to simplify the config rather than keeping special cases for it.

PR #998, opened 2024-06-25
Document provider-ci config options

danielrbradley:
Document current options, their purpose, and their usage.


provider-ci/internal/pkg/templates/bridged-provider.config.yaml
```diff
@@ -1,19 +1,66 @@
+# provider is the name of the provider without the pulumi-prefix e.g. "aws"
+# This MUST be set in every provider
+#provider: xyz
+
+# major version of the current provider - used in make files and goreleaser
+# This should always be set by all providers as this is key to.
```
guineveresaenger, line 6: I think you're missing part of this sentence here.

guineveresaenger (approved):
This is incredible. I had one comment but thank you so much for this awesome work!!!

PR #1003, opened 2024-06-26
Use setup-tools everywhere

danielrbradley:
Commits are factored for review:

1. Use setup-tools anywhere we're installing go, pulumicli, pulumictl
2. Remove unused pulumictl install for Go SDK release job
3. Now we only have one use of each setup step, inline the tool versions and remove from the config.
4. Also inline `slash-command` and `create-or-update-comment` actions as they're only used once.
5. Specify tools explicitly for the verify-release job.
6. Re-generate test-providers.

### DIFF
--- a/provider-ci/internal/pkg/templates/provider/.github/actions/setup-tools/action.yml
+++ b/provider-ci/internal/pkg/templates/provider/.github/actions/setup-tools/action.yml
@@ -26,6 +26,7 @@ runs:
         cache-dependency-path: |
           provider/*.sum
           upstream/*.sum
+          sdk/*.sum
 
     - name: Install pulumictl
       if: inputs.tools == 'all' || contains(inputs.tools, 'pulumictl')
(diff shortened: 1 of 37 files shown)


provider-ci/internal/pkg/templates/provider/.github/actions/setup-tools/action.yml
```diff
@@ -20,61 +20,62 @@ runs:
...
     - name: Install Go
       if: inputs.tools == 'all' || contains(inputs.tools, 'go')
-      uses: #{{ .Config.actionVersions.setupGo }}#
+      uses: actions/setup-go@v5
       with:
         go-version: "#{{ .Config.toolVersions.go }}#"
         cache-dependency-path: |
           provider/*.sum
           upstream/*.sum
+          sdk/*.sum
```
blampe, line 29: Why not `**/*.sum`?

blampe (approved):
I do love that we're consolidating the tool setup step. Makes it easier to later drop in something like [hermit](https://cashapp.github.io/hermit/) or [mise](http://mise.jdx.dev), to keep local dev in sync with CI.

provider-ci/internal/pkg/templates/provider/.github/actions/setup-tools/action.yml
danielrbradley: Just consolidating what's there, but perhaps that could work too ... one for another day though I think!

PR #1004, opened 2024-06-26
Install tools package publishing

danielrbradley:
Re-use our own tool setups which can include additional options for caching rather than duplicating the setup logic into the publishing action.

This depends on first merging:
- https://github.com/pulumi/pulumi-package-publisher/pull/32
- https://github.com/pulumi/ci-mgmt/pull/1003

This PR assumes that `pulumi/pulumi-package-publisher` will then be published as `v0.0.19` (the next version).

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider.config.yaml
+++ b/provider-ci/internal/pkg/templates/bridged-provider.config.yaml
@@ -149,7 +149,7 @@ actionVersions:
 publish:
   # publisherAction is the version of the pulumi-package-publisher action to use.
   # This should be pinned to just the major version once v1 is released.
-  publisherAction: pulumi/pulumi-package-publisher@v0.0.18
+  publisherAction: pulumi/pulumi-package-publisher@v0.0.19
   # passed to the sdk input of pulumi-package-publisher
   # This is overridden in pulumi-local to disable python
   sdk: all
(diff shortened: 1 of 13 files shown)


danielrbradley:
> Recommend testing this using a specific commit before cutting v0.0.19, so you don't get into a situation where you need to keep cutting releases to fix problems if anything comes up.

Good throught! Here's the downstream test commit pinning to pulumi/pulumi-package-publisher@4118cd60b193847811a9420f35b7c8d4181a65a5: https://github.com/pulumi/pulumi-xyz/pull/235/commits/d384ed39992e232bd6d003e707531873cf4c81c1

Here's the test release run (v1.0.2-alpha.10) https://github.com/pulumi/pulumi-xyz/actions/runs/9685487620

danielrbradley:
Rebased and removed the now-obsolete version arguments when calling the action.

danielrbradley:
Test showed issue that we can't call local reusable workflow without a checkout. So I've also removed the checkout phase from pulumi-package-publisher to be done before calling the pulumi-package-publisher action and before calling the setup-tools action.

danielrbradley:
A further thought on this ... we seem to have a lot of breaking changes already in the `pulumi-package-publisher` action. Perhaps using a standalone action here is the wrong abstraction right now.

In addition to removing the tool setup and the repository checkout, we probably also want to remove the artifact restore because this needs to have knowledge of the inner workings of the workflow that's calling the action:
- What version of the asset upload/download action are we using (these must be in sync)
- What's the name of the asset
- What the folder structure inside the asset
- Where's the SDK located within the calling repo.

Given needing to keep all these things in sync, it would be easier for now to move the implementation for the `pulumi-package-publisher` action back inside `ci-mgmt` for now to keep them versioned in lock-step with the calling workflows rather than having to coordinate breaking changes.

PR #1006, opened 2024-06-26
Remove unused publish.goSdk config

danielrbradley:
These are never actually overridden so can be inlined to reduce complexity of possible conditons.

Single reference to goSdk was in pulumi-kafka which had no effect. PR to remove: https://github.com/pulumi/pulumi-kafka/pull/459

Stacked on top of:
- https://github.com/pulumi/ci-mgmt/pull/1003

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/prerelease.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/prerelease.yml
@@ -95,7 +95,6 @@ jobs:
         java-version: "#{{ .Config.toolVersions.java }}#"
         node-version: "#{{ .Config.toolVersions.node }}#"
         python-version: "#{{ .Config.toolVersions.python }}#"
-#{{- if .Config.publish.goSdk.usePush }}#
   publish_go_sdk:
     name: publish_go_sdk
     needs: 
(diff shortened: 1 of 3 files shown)


provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/prerelease.yml
```diff
@@ -131,18 +130,17 @@ jobs:
...
     - uses: pulumi/publish-go-sdk-action@v1
       with:
-        repository: #{{ .Config.publish.goSdk.repository }}#
-        base-ref: #{{ .Config.publish.goSdk.baseRef }}#
-        source: #{{ .Config.publish.goSdk.source }}#
-        path: #{{ .Config.publish.goSdk.path }}#
+        repository: ${{ github.repository }}
+        base-ref: ${{ github.sha }}
+        source: sdk
+        path: sdk
```
blampe, line 136: Thinking ahead, if we're planning on folding native providers into this stuff (I hope we are... I can't stand the TS steps!) -- docker-build wouldn't need this step/action since it has one module at the repo root. All of this becomes unnecessary, so you may want to at least preserve the ability to disable this step (but this is not a blocker at all).

I've shouted this from the mountain tops many times, but the _only_ reason to drop `go.mod` into a subdirectory is because you need to version that subdirectory separately. In providerland this is never what we actually want, since a provider is tightly coupled to the SDK it generates. The cost of all of these modules is a more complicated and error-prone release process, as you're well aware :)

blampe (approved):

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/prerelease.yml
danielrbradley: Yes - definitely the plan. 

It looks like we'll need special handling for p-docker-build ... though it still needs this step, but the path is `sdk/go/dockerbuild` rather than `sdk` because [that's where its SKD's go.mod lives](https://github.com/pulumi/pulumi-docker-build/blob/main/sdk/go/dockerbuild/go.mod).

We'll likely include a slightly more targeted option as we onboard the native providers.

blampe: Ah yeah I'm getting ahead of myself, I haven't actually removed that `go.mod` yet -- only in my dreams :)

Issue #967, opened 2024-06-11
Test the provider binaries we're planning to release


danielrbradley:
Currently, we build one provider binary during prerequisites which is captured as an asset and used for running our integration tests against. However, we then discard this binary and use goreleaser to build brand new binaries at the point of release which are not tested.

We should build the provider binaries for all platforms at the beginning of the workflow, capture them as asstes, use them for integration tests, then publish the _tested_ provider binaries.

This will either require the removal of goreleaser (see Azure Native for how to release without goreleaser) or pay for their premium product.

t0yv0:
Goreleaser is removed but we are still double-building and testing the wrong binaries. Could be a very good change.

blampe:
Would have prevented this p0 https://github.com/pulumi/pulumi-aws/issues/4863

Maybe a dupe of https://github.com/pulumi/ci-mgmt/issues/1033?

Issue #731, opened 2023-11-30
Improve make provider caching for bridged providers


VenelinMartinov:
Currently `make provider` reruns `upstream` and `tfgen` every time. It'd be great if it cached these steps and just re-ran them only if needed.

Longer discussion [here](https://pulumi.slack.com/archives/C02FXTZEZ6W/p1700061870452729).

PR #1016, opened 2024-06-28
Remove goreleaser

### LINKED ISSUE #816: Hook failures in goreleaser
There seems to be an attempt to save disk space by having hooks in goreleaser. Because of cross-builds it can have intense disk requirements. I would prefer a robust solution here either having each platform/arch combination have its own container like in pulumi/pulumi or removing goreleaser entirely from the pipeline like in pulumi/azure-native.

```
  ⨯ release failed after 3m44s               error=post hook failed: shell: 'env GOOS=windows GOARCH=amd64 go clean -cache': exit status 1: go: unlinkat /home/runner/.cache/go-build/80: directory not empty
Error: The process '/opt/hostedtoolcache/goreleaser-action/1.24.0/x64/goreleaser' failed with exit code 1
```

-  https://github.com/pulumi/pulumi-wavefront/issues/330

### LINKED ISSUE #1013: Release changelog is missing
It looks like we stopped producing useful changelogs for releases a couple of weeks ago. Examples:

- https://github.com/pulumi/pulumi-aws/releases/tag/v6.42.0 is missing https://github.com/pulumi/pulumi-aws/pull/4083
- https://github.com/pulumi/pulumi-gcp/releases is missing https://github.com/pulumi/pulumi-gcp/pull/2095

Most releases' changelog is just one last change (last commit). Instead, they should include all changes.

danielrbradley:
Stop using goreleaser, improve our release notes, speed up releases.

- Fixes #1013
- Closes #816
- Partially addresses #731
- Related to #967

This should also save 3 hours of waiting per release of AWS.

Build cross-platform provider binaries, in parallel, straight after prerequisites, then only calculate hashes & push to S3 and GitHub releases directly during the publish job.

image

Specifics:

- Use GitHub's own release notes generator which allows us to suppliment with our own context - specifically the schema change since the previous latest release.
- Capture and restore the `schema-embed.json` to avoid having to run `tfgen` on every provider build.
- Build the multi-platform build straight into the makefile for easier local testing.
- Remove clearing disk space during publish as we're not doing any build work there any more.

This does not include reworking tests to use the the provider binary being released - but this can be added later fairly easily.

Prerelease run: https://github.com/pulumi/pulumi-xyz/actions/runs/9749848070
Stable release run: https://github.com/pulumi/pulumi-xyz/actions/runs/9750181828
Example of published release: https://github.com/pulumi/pulumi-xyz/releases/tag/v1.0.3


provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_provider.yml
```diff
@@ -0,0 +1,64 @@
+name: "Build Provider"
+
+on:
+  workflow_call:
+    inputs:
+      version:
```
iwahbe, line 6: I'd appreciate if this had a [`description`](https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#example-of-onworkflow_callinputs). It's not obvious from the name that the version is the version of the provider to be built.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/publish.yml
```diff
@@ -57,25 +50,44 @@ jobs:
...
+      id: schema-summary
+      shell: bash
+      run: |
+        # Get latest stable release. Return only first column from result (tag).
+        LAST_VERSION=$(gh release view --repo pulumi/pulumi-#{{ .Config.provider }}# --json tagName -q .tagName)
+        {
+          echo 'summary<<EOF'
+          schema-tools compare --provider="#{{ .Config.provider }}#" --old-commit="$LAST_VERSION" --new-commit="--local-path=provider/cmd/pulumi-resource-#{{ .Config.provider }}#/schema.json"
+          echo 'EOF'
+        } >> "$GITHUB_OUTPUT"
```
iwahbe, line 75: TIL I learned that bash supported blocks.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/publish.yml
```diff
@@ -57,25 +50,44 @@ jobs:
...
-    - name: Run GoReleaser (prerelease)
-      if: inputs.isPrerelease == true
-      uses: #{{ .Config.actionVersions.goReleaser }}#
+        tag_name: v${{ inputs.version }}
+        prerelease: ${{ inputs.isPrerelease }}
+        # We keep pre-releases as drafts so they're not visible until we manually publish them.
+        draft: ${{ inputs.isPrerelease }}
+        body: ${{ steps.schema-summary.outputs.summary }}
+        generate_release_notes: true
+        files: dist/*
```
iwahbe, line 88: Does using this action give us anything over `gh release create`? If not, should we just use the CLI command to avoid the indirection of a 3rd party action?

iwahbe (approved):
This looks good to me. I'd love to see this used end to end on a provider before we roll it out to the entire fleet.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/publish.yml
danielrbradley: Mostly just easier to integrate with options and better output. Using the CLI would be a little awkward to conditionally include the `--draft` and `--prerelease` flags.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/publish.yml
danielrbradley: Ditto - shellcheck flagged this one for me!

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_provider.yml
```diff
@@ -0,0 +1,65 @@
...
+        with:
+          tool-cache: false
+          swap-storage: false
+      #{{- end }}#
+      - name: Checkout Repo
+        uses: #{{ .Config.actionVersions.checkout }}#
+        #{{- if .Config.checkoutSubmodules }}#
+        with:
+          submodules: #{{ .Config.checkoutSubmodules }}#
+      #{{- end }}#
```
thomas11, line 45: [nit] indentation

thomas11 (approved):

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/publish.yml
```diff
@@ -57,25 +50,44 @@ jobs:
...
+        merge-multiple: true
+    - name: Calculate checksums
+      working-directory: dist
+      run: shasum ./*.tar.gz > pulumi-#{{ .Config.provider }}#_${{ inputs.version }}_checksums.txt
+    - name: Get Schema Change Summary
+      id: schema-summary
+      shell: bash
+      run: |
+        # Get latest stable release. Return only first column from result (tag).
+        LAST_VERSION=$(gh release view --repo pulumi/pulumi-#{{ .Config.provider }}# --json tagName -q .tagName)
```
flostadler, line 70: This assumes we do not do any patch releases for older versions, right?

flostadler (approved):
LGTM - just one small question

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/publish.yml
danielrbradley: This will return whatever has been marked as "latest" in GitHub: https://cli.github.com/manual/gh_release_view

If we're patching older major versions, we should not mark these as latest. The tooling probably does automatically mark them as latest right now, but we have to re-set the correct latest version.

The schema diff will also be wrong for patches to old major versions too as we'd actually want to compare with the previous by comparing sequentially.

I think there's an existing issue capturing this, but I can't seem to find it right now

flostadler: Ah nice, I thought it's just getting the latest by time.

danielrbradley:
> This looks good to me. I'd love to see this used end to end on a provider before we roll it out to the entire fleet.

pulumi-kafka main run: https://github.com/pulumi/pulumi-kafka/actions/runs/9757616987
pulumi-azure main run: https://github.com/pulumi/pulumi-azure/actions/runs/9758241530

PR #1021, opened 2024-07-02
Add documentation for license check ignore option

guineveresaenger:
This field was missing in the config.yaml docs.

Real-world example: https://github.com/pulumi/pulumi-alicloud/blob/34d9dd0891282f9e0c4eb4d284ac1e560e2ea678/.ci-mgmt.yaml#L24


mjeffryes (approved):
🙇‍♂️ Thank you!

PR #1028, opened 2024-07-10
Disable respectSchemaVersion migration

danielrbradley:
This has been applied to all providers.

### DIFF
--- a/tools/sourcemigrator/index.ts
+++ b/tools/sourcemigrator/index.ts
@@ -72,32 +72,32 @@ function run(command: string): shell.ShellString {
   return checkError(shell.exec(command));
 }
 
-function respectSchemaVersion(): SourceMigration {
-  return {
-    name: "Respect Schema Version",
-    async execute(context) {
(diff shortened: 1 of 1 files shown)


tools/sourcemigrator/index.ts
```diff
@@ -72,32 +72,32 @@ function run(command: string): shell.ShellString {
...
-        if (gitStatus.includes("provider/resources.go")) {
-          run(`make tfgen build`);
-        }
-      } finally {
-        shell.popd();
-      }
-    },
-  };
-}
+// function respectSchemaVersion(): SourceMigration {
```
VenelinMartinov, line 75: should we delete it instead? It is preserved in git history

VenelinMartinov (approved):

tools/sourcemigrator/index.ts
danielrbradley: It's kinda handy as a template ... we should probably write these examples into docs at some point instead 😑

PR #1041, opened 2024-07-15
Force initialize the submodule on `make upstream`

guineveresaenger:
When `upstream` locally gets out of sync with the current default branch tip, any run of `make upstream` will exit 1 on `assert_upstream`tracked` in the `upstream.sh` script, because it will have been detected as modified. 
This happens whenever anyone pulls in upstream changes to work on a feature or bug fix. The instructions [provided](https://github.com/pulumi/ci-mgmt/blob/master/provider-ci/internal/pkg/templates/bridged-provider/upstream.sh#L96) do not work.

This pull request defaults to pass `-f` to the `init` command run via `make upstream`.

Since the [upgrade provider tool](https://github.com/pulumi/upgrade-provider/blob/main/upgrade/steps.go#L163) always force resets the submodule as well, this should be a safe change.

It is dubious why we'd need the non-force option at all, and the fact that the remedial instructions do not work is another issue as well.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/Makefile
+++ b/provider-ci/internal/pkg/templates/bridged-provider/Makefile
@@ -187,7 +187,7 @@ tfgen_build_only:
 
 upstream:
 ifneq ("$(wildcard upstream)","")
-	./upstream.sh init
+	./upstream.sh init -f
 endif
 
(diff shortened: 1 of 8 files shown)


iwahbe (approved):
Can we also fix the remedial instructions?

PR #1042, opened 2024-07-17
Only download the Pulumi CLI when the version doesn't match

iwahbe:
Currently, we re-download the Pulumi CLI whenever `provider/go.mod` changes. This is wasteful in general, but especially punishing to contributors who don't have a strong connection.

This PR adds a local check of the installed version before we download a new version of the CLI.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/Makefile
+++ b/provider-ci/internal/pkg/templates/bridged-provider/Makefile
@@ -215,8 +215,19 @@ ci-mgmt: .ci-mgmt.yaml
 		--template bridged-provider \
 		--config $<
 
+# Because some codegen depends on the version of the CLI used, we install a local CLI
+# version pinned to the same version as `provider/go.mod`.
+#
+# This logic compares the version of .pulumi/bin/pulumi already installed. If it matches
+# the desired version, we just print. Otherwise we (re)install pulumi at the desired
+# version.
 .pulumi/bin/pulumi: .pulumi/version
-	curl -fsSL https://get.pulumi.com | HOME=$(WORKING_DIR) sh -s -- --version $(shell cat .pulumi/version)
+	@if [ -x .pulumi/bin/pulumi ] && [ "v$$(cat .pulumi/version)" = "$$(.pulumi/bin/pulumi version)" ]; then \
+		echo "pulumi/bin/pulumi version: v$$(cat .pulumi/version)"; \
+	else \
+		curl -fsSL https://get.pulumi.com | \
+			HOME=$(WORKING_DIR) sh -s -- --version "$$(cat .pulumi/version)"; \
+	fi
 
 # Compute the version of Pulumi to use by inspecting the Go dependencies of the provider.
 .pulumi/version: provider/go.mod
--- a/provider-ci/test-providers/aws/Makefile
+++ b/provider-ci/test-providers/aws/Makefile
@@ -196,8 +196,19 @@ ci-mgmt: .ci-mgmt.yaml
 		--template bridged-provider \
 		--config $<
 
+# Because some codegen depends on the version of the CLI used, we install a local CLI
+# version pinned to the same version as `provider/go.mod`.
+#
+# This logic compares the version of .pulumi/bin/pulumi already installed. If it matches
+# the desired version, we just print. Otherwise we (re)install pulumi at the desired
+# version.
 .pulumi/bin/pulumi: .pulumi/version
-	curl -fsSL https://get.pulumi.com | HOME=$(WORKING_DIR) sh -s -- --version $(shell cat .pulumi/version)
+	@if [ -x .pulumi/bin/pulumi ] && [ "v$$(cat .pulumi/version)" = "$$(.pulumi/bin/pulumi version)" ]; then \
+		echo "pulumi/bin/pulumi version: v$$(cat .pulumi/version)"; \
+	else \
+		curl -fsSL https://get.pulumi.com | \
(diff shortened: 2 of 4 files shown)


provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -215,8 +215,19 @@ ci-mgmt: .ci-mgmt.yaml
...
 
+# Because some codegen depends on the version of the CLI used, we install a local CLI
+# version pinned to the same version as `provider/go.mod`.
+#
+# This logic compares the version of .pulumi/bin/pulumi already installed. If it matches
+# the desired version, we just print. Otherwise we (re)install pulumi at the desired
+# version.
 .pulumi/bin/pulumi: .pulumi/version
-	curl -fsSL https://get.pulumi.com | HOME=$(WORKING_DIR) sh -s -- --version $(shell cat .pulumi/version)
+	@if [ -x .pulumi/bin/pulumi ] && [ "v$$(cat .pulumi/version)" = "$$(.pulumi/bin/pulumi version)" ]; then \
```
t0yv0, line 225: Feels like if only GNU Make knew how to compare files by content hashes, and not by timestamp, this would not have been needed at all right? Looks like all solutions here are not without downsides. There is https://github.com/olipratt/hashdeps but it's another dependency. Perhaps if the target for `.pulumi/version: provider/go.mod` refused to overwrite `.pulumi/version` if it's already tracking the desired content, then the `atime` of `.pulumi/version` would not change needlessly and it would be another way to do this.

t0yv0 (approved):

provider-ci/internal/pkg/templates/bridged-provider/Makefile
iwahbe: > Feels like if only GNU Make knew how to compare files by content hashes, and not by timestamp, this would not have been needed at all right?

If GNU knew how to compare files by content, then this would not be needed, yeah.

> Perhaps if the target for .pulumi/version: provider/go.mod refused to overwrite .pulumi/version if it's already tracking the desired content, then the atime of .pulumi/version would not change needlessly and it would be another way to do this.

I like that idea. Unfortunately, testing shows that this doesn't work. Given the following patch:

```patch
 Makefile | 7 ++++++-
 1 file changed, 6 insertions(+), 1 deletion(-)

diff --git a/Makefile b/Makefile
index 9797f5c..77215ba 100644
--- a/Makefile
+++ b/Makefile
@@ -189,9 +189,14 @@ ci-mgmt: .ci-mgmt.yaml
 	curl -fsSL https://get.pulumi.com | HOME=$(WORKING_DIR) sh -s -- --version $(shell cat .pulumi/version)
 
 # Compute the version of Pulumi to use by inspecting the Go dependencies of the provider.
+.pulumi/version: USED=$(shell cd provider && go list -f "{{slice .Version 1}}" -m github.com/pulumi/pulumi/pkg/v3 | tee ../$@)
 .pulumi/version: provider/go.mod
 	@mkdir -p .pulumi
-	@cd provider && go list -f "{{slice .Version 1}}" -m github.com/pulumi/pulumi/pkg/v3 | tee ../$@
+	if [ -x .pulumi/bin/pulumi ] && [ -f "$@" ] && [ "$$(cat "$@")" = "${USED}" ]; then \
+		echo ".pulumi/version up to date at ${USED}"; \
+	else \
+		cd provider && go list -f "{{slice .Version 1}}" -m github.com/pulumi/pulumi/pkg/v3 | tee ../$@; \
+	fi
 
 # Start debug server for tfgen
 debug_tfgen:
```

We still see this output
```
make .pulumi/bin/pulumi                                      
if [ -x .pulumi/bin/pulumi ] && [ -f ".pulumi/version" ] && [ "$(cat ".pulumi/version")" = "3.124.0" ]; then \
        echo ".pulumi/version up to date at 3.124.0"; \
else \
        cd provider && go list -f "{{slice .Version 1}}" -m github.com/pulumi/pulumi/pkg/v3 | tee ../.pulumi/version; \
fi
.pulumi/version up to date at 3.124.0
curl -fsSL https://get.pulumi.com | HOME=/Users/ianwahbe/go/src/github.com/pulumi/pulumi-random sh -s -- --version 3.124.0
=== Upgrading Pulumi v3.124.0 to v3.124.0 ===
+ Downloading https://github.com/pulumi/pulumi/releases/download/v3.124.0/pulumi-v3.124.0-darwin-arm64.tar.gz...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  146M  100  146M    0     0  70.8M      0  0:00:02  0:00:02 --:--:--  104M
+ Extracting to /Users/ianwahbe/go/src/github.com/pulumi/pulumi-random/.pulumi/bin

warning: Pulumi has been installed to /Users/ianwahbe/go/src/github.com/pulumi/pulumi-random/.pulumi/bin, but it looks like there's a different copy
         on your $PATH at /Users/ianwahbe/go/bin. You'll need to explicitly invoke the
         version you just installed or modify your $PATH to prefer this location.

=== Pulumi is now installed! 🍹 ===
+ Get started with Pulumi: https://www.pulumi.com/docs/quickstart
```

> [!NOTE]
> Note that we see both `.pulumi/version up to date at 3.124.0` and `=== Upgrading Pulumi v3.124.0 to v3.124.0 ===`.

PR #1045, opened 2024-07-23
Provide better instructions when upstream is modified

danielrbradley:
- Recommend adding `-f` rather than custom git checkout commands.
- Hide the `format_patches` command as `check_in` is a much less confusing version of the command.
- Don't use the force option from the makefile as we don't want to wipe out any uncommitted work by accident.
- Relax validation for running 'init' and 'checkout' without the force option. Only require upstream to not be checked out on the 'pulumi/patch-checkout' branch, but don't require it to be tracked in git. This allows the unforced init to be used after pulling changes but not having yet updated the submodule.
- Assert we're not mid-rebase if force not provided (avoids accidentally loosing progress if switched to our own branch).
- Clean up old branches and rebases when using the force option with either init or checkout.

Example error (for init):
```
Error: 'upstream' submodule checked out on the 'pulumi/patch-checkout' branch.
This was likely caused by running a 'checkout' command but not running
'check_in' afterwards.

To turn the commits in the 'pulumi/patch-checkout' branch back into patches, run:
  ./upstream.sh check_in

To disgard changes in the 'pulumi/patch-checkout' branch, use the 'force' flag (-f):
  ./upstream.sh init -f
```

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/Makefile
+++ b/provider-ci/internal/pkg/templates/bridged-provider/Makefile
@@ -187,7 +187,7 @@ tfgen_build_only:
 
 upstream:
 ifneq ("$(wildcard upstream)","")
-	./upstream.sh init -f
+	./upstream.sh init
 endif
 
 #{{- if .Config.XrunUpstreamTools }}#
--- a/provider-ci/internal/pkg/templates/bridged-provider/Makefile
+++ b/provider-ci/internal/pkg/templates/bridged-provider/Makefile
@@ -238,7 +238,7 @@ ci-mgmt: .ci-mgmt.yaml
 debug_tfgen:
 	dlv  --listen=:2345 --headless=true --api-version=2  exec $(WORKING_DIR)/bin/$(TFGEN) -- schema --out provider/cmd/$(PROVIDER)
 
-.PHONY: development build build_sdks install_go_sdk install_java_sdk install_python_sdk install_sdks only_build build_dotnet build_go build_java build_nodejs build_python clean cleanup#{{ if .Config.docsCmd }}# docs#{{end}}# help install_dotnet_sdk install_nodejs_sdk install_plugins lint_provider provider provider_no_deps test tfgen upstream upstream.finalize upstream.rebase ci-mgmt test_provider debug_tfgen tfgen_build_only
+.PHONY: development build build_sdks install_go_sdk install_java_sdk install_python_sdk install_sdks only_build build_dotnet build_go build_java build_nodejs build_python clean cleanup#{{ if .Config.docsCmd }}# docs#{{end}}# help install_dotnet_sdk install_nodejs_sdk install_plugins lint_provider provider provider_no_deps test tfgen upstream ci-mgmt test_provider debug_tfgen tfgen_build_only
 
 # Provider cross-platform build & packaging
 
--- a/provider-ci/internal/pkg/templates/bridged-provider/upstream.sh
+++ b/provider-ci/internal/pkg/templates/bridged-provider/upstream.sh
@@ -12,7 +12,7 @@ NAME
   upstream.sh - Manages applying patches to the upstream submodule.
 
 SYNOPSIS
-  ${original_exec} <init|checkout|rebase|format_patches|check_in|help> [options]
+  ${original_exec} <init|checkout|rebase|check_in|help> [options]
 
 COMMANDS
   init [-f]             Initialize the upstream submodule and applies the
@@ -20,7 +20,6 @@ COMMANDS
   checkout [-f]         Create a branch in the upstream repository with the
                         patches applied as commits.
   rebase [-o] [-i]      Rebase the checked out patches.
-  format_patches        Write checkedout commits back to patches.
   check_in              Write checkedout commits back to patches, add upstream
                         and patches changes to the git staging area and exit
                         checkout mode.
@@ -51,12 +50,12 @@ EXAMPLES
   Moving the patches to a new base commit:
     ${original_exec} checkout
     ${original_exec} rebase -o <new_base_commit>
-    ${original_exec} format_patches
+    ${original_exec} check_in
 
   Interactively edit the patches:
     ${original_exec} checkout
     ${original_exec} rebase -i
-    ${original_exec} format_patches
+    ${original_exec} check_in
 EOF
 }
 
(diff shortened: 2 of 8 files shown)


provider-ci/internal/pkg/templates/bridged-provider/upstream.sh
```diff
@@ -51,12 +50,12 @@ EXAMPLES
   Moving the patches to a new base commit:
     ${original_exec} checkout
     ${original_exec} rebase -o <new_base_commit>
-    ${original_exec} format_patches
+    ${original_exec} check_in
```
iwahbe, line 53: If one command is `checkout`, then this command should be `checkin`.

provider-ci/internal/pkg/templates/bridged-provider/upstream.sh
```diff
@@ -71,35 +70,41 @@ assert_upstream_exists() {
 assert_upstream_tracked() {
   status=$(git status --porcelain upstream)
   if [[ ${status} == *"M upstream" ]]; then
+    cat <<EOF
+Error: The 'upstream' submodule is modified with untracked changes, continuing
+might loose changes in the 'upstream' submodule.
+
+Git status of 'upstream':
+${status}
+
```
iwahbe, line 79: Can we add a "to discard existing changes, run `${original_exec} init -f`".

provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -187,7 +187,7 @@ tfgen_build_only:
 
 upstream:
 ifneq ("$(wildcard upstream)","")
-	./upstream.sh init -f
+	./upstream.sh init
```
iwahbe, line 190: Using the makefile is much more convenient then `upstream.sh`, especially since `make upstream` autocompletes both words and `./upstream.sh init -f` autocompletes 0/3 words. This is adding an extra step for every manual provider update on my squad, since we generally didn't do the last update.

If you don't want `make upstream` to reset the upstream state, then lets add a new target (maybe `make upstream_init`).

iwahbe (commented):
I understand the desire to not have `make tfgen` reset upstream, but `./upstream.sh init -f` is much less ergonomic. Let's make sure we offer an equivalent `make` command.

provider-ci/internal/pkg/templates/bridged-provider/upstream.sh
danielrbradley: I think this is a case of English being a silly language - checkout is a valid word, but checkin is not. However, "check out" is a valid alternative to match with "check out", but then it doesn't match with git's use of "checkout" ... I'll add aliases to make the underscore optional in both, but gramatically this is correct.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
danielrbradley: Using the makefile should be absolutely fine still for 99% of usage - without the force.

In general, you should never need to use the force option - only if you've manually updated the upstream checkout instead of using the upgrade tool, so copying and pasting the direct script command doesn't seem too bad as a one-off.

I'm interested to understand how else people might end up with their upstream out-of-sync. Perhaps if the upgrade tool fails out during the rebase? I think there's also some issues here with the upgrade-tool not printing out errors from commands its executing.

provider-ci/internal/pkg/templates/bridged-provider/upstream.sh
danielrbradley: This is already below - see the example in the description - we either print:

```
To disgard changes in the 'pulumi/patch-checkout' branch, use the 'force' flag (-f):
  ./upstream.sh init -f
```
or

```
To continue and discard the changes to upstream, run:
  ./upstream.sh checkout -f
```

depending on the upstream checked out ref and the command which was run.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
iwahbe: You need to use force when:

1. You have a provider repo cloned to your computer, say `haspatches` at `v1.2.0`.
2. Someone else upgrades `haspatches` to `v1.3.0` and pushes the upgrade back to the `main` branch.
3. You pull in the changes from `main`. 

Because git submodules are, IMO, poorly behaved, the `git pull` won't update the submodule on disk. After (3), `git status` will show a diff on `upstream`, where it *is* pointing to `v1.2.0` but `main` points to `v1.3.0`.

Because my squad works across so many providers, this is the common case when I go to do an upgrade.

provider-ci/internal/pkg/templates/bridged-provider/upstream.sh
iwahbe: Thanks.

"check-in" is a valid word (unhelpfully), and the plural is `checkins` (also unhelpfully).

danielrbradley:
> I understand the desire to not have `make tfgen` reset upstream, but `./upstream.sh init -f` is much less ergonomic. Let's make sure we offer an equivalent `make` command.

Ok, I think I've realised what the disconnect here is. I've always operated git with the `submodule.recurse` option set to `true` which means a `git pull` in the root repo updates the upstream submodule to the updated commit. But if you don't auto-recurse, then it appears out of sync, and requires a force init.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
danielrbradley: See comment on the main PR ... but yes, I think everyone should be setting `git config --global submodule.recurse true` to get rid of this headache!

danielrbradley: But this also makes much more sense why this is a pain now! TY

danielrbradley: I think we can safely assume that if we're not on the `pulumi/patch-checkout` then we can safely blow away all submodule changes without the force option.

iwahbe: That sounds reasonable.

guineveresaenger:
Is the recommendation here that everyone change their git config? or could we 
1. check if we're on `patch-checkout`
2. if no, force reset
?

iwahbe:
> Is the recommendation here that everyone change their git config? 

No. `upstream.sh` will accommodate default `git` configuration.

guineveresaenger (approved):
I've verified the changes correctly re-set upstream without the `-f` option.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
guineveresaenger, line 1: Could you also remove the `upstream.rebase` and `upstream.finalize` mentions on L241 before merging please?

iwahbe (approved):

PR #1049, opened 2024-08-05
This prevents upstream from carrying un-tracked changes past `init`

iwahbe:
This was motivated by an hour I spent tracking down a persistent diff between CI and my laptop when there was an un-tracked `website` folder in `upstream` that wasn't present in CI.


VenelinMartinov (approved):
Looks good!

danielrbradley (approved):

PR #1051, opened 2024-08-06
Fix publish-go-sdk globbing

blampe:
The current globbing for publish-go-sdk breaks publishing for docker-build because
* `go.*` doesn't match anything (~~probably a typo of `*.go`~~ `go.mod` lives under `sdk/go/dockerbuild`)
* `go/**` also doesn't match anything because the SDK module is located under `sdk/go/dockerbuild` instead of `sdk` (so it would need `**`).

The result is an [almost-empty](https://github.com/pulumi/pulumi-docker-build/tree/sdk/go/dockerbuild/v0.0.4/sdk/go/dockerbuild) SDK branch.

This PR changes the globbing to match any `.go` files under the SDK path.

### DIFF
--- a/native-provider-ci/src/steps.ts
+++ b/native-provider-ci/src/steps.ts
@@ -1040,8 +1040,7 @@ export function PublishGoSdk(sdkModuleDir: string): Step {
       version: "${{ steps.version.outputs.version }}",
       additive: false,
       files: `\
-go.*
-go/**
+**/*.go
 !*.tar.gz`,
     },
   };
--- a/native-provider-ci/src/steps.ts
+++ b/native-provider-ci/src/steps.ts
@@ -1039,9 +1039,14 @@ export function PublishGoSdk(sdkModuleDir: string): Step {
       path: sdkModuleDir,
       version: "${{ steps.version.outputs.version }}",
       additive: false,
-      files: `\
-**/*.go
-!*.tar.gz`,
+      // TODO(https://github.com/pulumi/pulumi/issues/5403): go.mod should be scoped to only the Go SDK.
+      files:
+        sdkModuleDir == "sdk"
+          ? `\
+go.*
+go/**
+!*.tar.gz`
+          : "**",
     },
   };
 }
--- a/native-provider-ci/providers/aws-native/repo/.github/workflows/prerelease.yml
+++ b/native-provider-ci/providers/aws-native/repo/.github/workflows/prerelease.yml
@@ -567,5 +567,6 @@ jobs:
         version: ${{ steps.version.outputs.version }}
         additive: false
         files: |-
-          **/*.go
+          go.*
+          go/**
           !*.tar.gz
--- a/native-provider-ci/providers/aws-native/repo/.github/workflows/prerelease.yml
+++ b/native-provider-ci/providers/aws-native/repo/.github/workflows/prerelease.yml
@@ -567,6 +567,5 @@ jobs:
         version: ${{ steps.version.outputs.version }}
         additive: false
         files: |-
-          go.*
-          go/**
+          **/*.go
           !*.tar.gz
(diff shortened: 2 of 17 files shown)


iwahbe (commented):
I think the `go.*` is for `go.mod` and `go.sum`. I'm not sure why the `go/**` doesn't capture `go/dockerbuild`. I would lean towards moving docker to publish in the same place as all other providers, but I'm less familiar with native provider publishing.

native-provider-ci/src/steps.ts
```diff
@@ -1040,8 +1040,7 @@ export function PublishGoSdk(sdkModuleDir: string): Step {
       version: "${{ steps.version.outputs.version }}",
       additive: false,
       files: `\
-go.*
-go/**
+**/*.go
 !*.tar.gz`,
```
danielrbradley, line 1044:
```suggestion
      files: sdkModuleDir == "sdk" ? `\
go.*
go/**
!*.tar.gz` : "**",
```

danielrbradley (changes requested):
This oddity is occurring because docker-build is a special case that puts its `go.mod` in a different location to every other provider.

The globs are calculated relative to the `source` argument. For all other providers the source is `sdk`, but for docker-build it's `sdk/go/dockerbuild`.

The quick fix is to add a conditional special-case for docker-build, so if it's a different source directory, we'll just assume we want everything in that folder.

The better fix here would be to align docker-build with the other providers so we can make our ci-scripts more uniform with less special options, but that would now be a breaking change for the Go SDK, so I don't think this is possible in the short term.

native-provider-ci/src/steps.ts
blampe: @danielrbradley what about something like this?
```
**/go.mod
**/go.sum
**/*.go
```

danielrbradley: That doesn't work because it will pick up all othe other folders in `sdk` such as `sdk/nodejs` which still contain stub `go.mod` files from when they were published in-place.

blampe: @danielrbradley do you have an example? I'm wondering if that's actually a problem - I would assume they aren't even usable.

danielrbradley: https://github.com/pulumi/pulumi-azure/blob/master/sdk/nodejs/go.mod

danielrbradley: It wouldn't break anything per se, but it would include additional files that we don't need or want published.

blampe: > it would include additional files that we don't need or want published.

But that's orthogonal isn't it? Presumably we want to stop generating these fake modules?

On the flip side, it's not unreasonable to imagine publishing a separate module for and SDK with generics or something. In other words we shouldn't assume anything about where or how many modules there might be. It's our goof for generating fake modules, but we've been publishing them for ages and this action isn't the right place to take care of them.

danielrbradley: Yes, we might want to remove these files at some point, but I think they still have some effect for local testing scenarios.

> On the flip side, it's not unreasonable to imagine publishing an additional module for an SDK with generics or something

There's lots of hypotheticals, but this tooling is only being used by us internally right now, so we should focus on the only two setups we're using right now–publishing the root `sdk` module, or publishing a more specific sub-folder. For the latter, it seems reasonable that we just include all files. We only even need to specify this argument for most providers right now because of the oddities of how the go module accidentally includes all the other SDKs.

I would suggest just adding the conditional as I proposed.

blampe: Sure, I made the change and left a pointer to https://github.com/pulumi/pulumi/issues/5403.

danielrbradley (approved):

PR #1052, opened 2024-08-13
Add free-disk-space action to upgrade provider workflow

flostadler:
This adds the `free-disk-space` action to the upgrade-provider workflow to prevent running out of disk space during compilation. It uses the existing `freeDiskSpaceBeforeBuild` config flag for controlling this.

Additionally this centralizes the config for the version of the `free-disk-space` action to use.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider.config.yaml
+++ b/provider-ci/internal/pkg/templates/bridged-provider.config.yaml
@@ -69,6 +69,9 @@ freeDiskSpaceBeforeBuild: false
 # Set to true to clear disk space before running test jobs.
 freeDiskSpaceBeforeTest: false
 
+# The version of the jlumbroso/free-disk-space action to use.
+freeDiskSpaceActionVersion: 54081f138730dfa15788a46383842cd2f914a1be
+
 # Used for centrally managing tool versions.
 # This is not currently overridden by any providers: https://github.com/search?q=org%3Apulumi+path%3A.ci-mgmt.yaml+%22toolVersions%22&type=code
 toolVersions:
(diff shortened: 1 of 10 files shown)


provider-ci/internal/pkg/templates/bridged-provider.config.yaml
```diff
@@ -69,6 +69,9 @@ freeDiskSpaceBeforeBuild: false
 # Set to true to clear disk space before running test jobs.
 freeDiskSpaceBeforeTest: false
 
+# The version of the jlumbroso/free-disk-space action to use.
+freeDiskSpaceActionVersion: 54081f138730dfa15788a46383842cd2f914a1be
```
danielrbradley, line 73: Move inside `actionVersions` (see line 149)

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-provider.yml
```diff
@@ -9,6 +9,14 @@ jobs:
     name: upgrade-provider
     runs-on: #{{ .Config.runner.default }}#
     steps:
+    #{{- if .Config.freeDiskSpaceBeforeBuild }}#
```
danielrbradley, line 12: I like reusing the "..beforeBuild" option 👍

danielrbradley (commented):
Looks good, but would tidy the config location.

provider-ci/internal/pkg/templates/bridged-provider.config.yaml
flostadler: Ah yeah, didn't see that before.Thanks for pointing that out!

danielrbradley (approved):

PR #1056, opened 2024-08-16
Make the GH organization configurable

ringods:
A first small change to the GH workflow generation in an attempt to bring `ci-mgmt` slowly closer to third party package maintainers.


danielrbradley (approved):
This makes sense to me. Paging the rest of providers for review to get a few more eyes on it. Let's aim to merge tomorrow to give others a chance to look over it too.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -208,9 +208,9 @@ bin/pulumi-java-gen: .pulumi-java-gen.version
 # - Run make ci-mgmt to apply the change locally.
 #
 ci-mgmt: .ci-mgmt.yaml
-	rm .github/workflows/*.yml # Copied from update-workflows.yml
+	rm -f .github/workflows/*.yml # Copied from update-workflows.yml
```
iwahbe, line 211: Why do we need the `-f` change here?

iwahbe (approved):
The change is fine. 

> [!WARNING]
> As it stands, I don't think we maintain backwards compatibility guarantees on `ci-mgmt`. It is very much a "use at your own risk" repository.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
ringods: @iwahbe I had a run where the `rm` took place, but the workflow generation failed. An `rm` without `-f` fails on an empty folder when retrying. The `-f` prevents this.

ringods:
I will submit more changes to get thirdparty support in. The only expectation I have is that people are aware of thirdparty support and not deliberately break it or hardcode new stuff to be Pulumi only.

If there is a regression with a third-party provider, I will fast enough find issues and be able to contribute fixes for it.

README.md
```diff
@@ -76,6 +76,8 @@ To add a new provider:
    ```
 
    `ci-mgmt` will read your provider's `.ci-mgmt.yaml` and generate the standard set of CI files from templates.
+   You can override every of the [default values](./provider-ci/internal/pkg/templates/bridged-provider.config.yaml)
```
guineveresaenger, line 79:
```suggestion
   You can override every one of the [default values](./provider-ci/internal/pkg/templates/bridged-provider.config.yaml)
```

guineveresaenger (approved):

PR #1057, opened 2024-08-19
Add option for freeing disk space to sdk build and bridge upgrade

flostadler:
Some of the workflows on larger providers like aws started failing because they're running out of disk space (again.. sigh). Examples are:
- https://github.com/pulumi/pulumi-aws/issues/4379
- https://github.com/pulumi/pulumi-aws/actions/runs/10421649189/job/28864324845

This change adds a new configuration parameter (`freeDiskSpaceBeforeSdkBuild`) to free the disk space before SDK builds. Additionally it also adds the disk space cleaning to the upgrade-bridge workflow.

We should also explore using bigger runners, but this might be cost prohibitive (AWS alone had ~1.1k build hours within the last month, which would amount to ~1k$ for a larger runner).

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider.config.yaml
+++ b/provider-ci/internal/pkg/templates/bridged-provider.config.yaml
@@ -69,6 +69,9 @@ freeDiskSpaceBeforeBuild: false
 # Set to true to clear disk space before running test jobs.
 freeDiskSpaceBeforeTest: false
 
+# Set to true to clear disk space before running sdk build jobs.
+freeDiskSpaceBeforeSdkBuild: false
+
 # Used for centrally managing tool versions.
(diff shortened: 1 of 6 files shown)


provider-ci/internal/pkg/templates/bridged-provider.config.yaml
```diff
@@ -69,6 +69,9 @@ freeDiskSpaceBeforeBuild: false
 # Set to true to clear disk space before running test jobs.
 freeDiskSpaceBeforeTest: false
 
+# Set to true to clear disk space before running sdk build jobs.
+freeDiskSpaceBeforeSdkBuild: false
```
flostadler, line 73: I initially thought about making this configurable per SDK because right now we only really need it for go. But this would make the config much more complex while not really bringing any benefit.
The go SDK build takes the longest out of the other SDKs in the build matrix, so the other SDKs have to wait for go anyways.

danielrbradley (approved):

Issue #676, opened 2023-10-12
Invert tags


t0yv0:
Consider inverting build tags. 

Currently providers rely on build tags to split tests by language so that the CI test matrix gains some parallelism and finishes faster. 

An example of this can be found in:

https://github.com/pulumi/pulumi-gcp/blob/master/examples/examples_go_test.go#L2 

It reads:

```
//go:build go || all
// +build go all
```

So unless the user passes `-tags all` or `-tags go`, the code in this file is excluded from compilation.

Unfortunately this sacrifices the DX for new contributors. If you open these files in an IDE it likely will not be able to quickly run these tests until you reconfigure Go test flags. This is a bad default for developer experience.

The proposal is to switch these tags to negative tags:

```
// +build !nodejs !python !yaml !java !dotnet
```

The CI splitting work just as before, while IDEs would be able to pick these tests up without configuration.

Consider implementing the change as a quick source migration in this repo so it can auto-apply to all ci-mgmt managed providers overnight.

blampe:
> Currently providers rely on build tags to split tests by language so that the CI test matrix gains some parallelism and finishes faster.

Build tags aren't the best way to speed up CI because they're static, and sharding on language means we tend to end up with one very "heavy" (usually nodejs) shard, meanwhile the java shard does literally nothing. We're already in a situation where adding a new shard for yaml tests is non-trivial.

Instead, it would be better to assign tests randomly to some number N workers. (Currently N=5 for go/nodejs/dotnet/python/java.)
1. Collect all tests: `go test -tags all ./... -list=.`
   * Note: this is another case where we make life harder for ourselves by having multiple `go.mod` files.
2. Segment tests into N groups.
3. Execute one group of tests per worker: `go test -tags all -run "^(A|B|C)$" ./...`.

Steps 1 and 2 are fast and can be performed independently and deterministically by each worker.

This could be rolled out in a backwards-compatible way by building with `-tags all` on each shard.

blampe:
@corymhall found https://github.com/hashicorp-forge/go-test-split-action which seems to do what I described above.

PR #1058, opened 2024-08-21
Turn on test sharding for k8s and docker-build

blampe:
This turns on test sharding for the docker-build and p-k providers.

The actual logic for how we discover tests and assign them is [here](https://github.com/blampe/shard/blob/main/internal/collect.go), along with some [test cases](https://github.com/blampe/shard/tree/main/internal/testdata) if there are correctness concerns.

In terms of onboarding, if a provider wants to use this they should:
* Ensure a `make install_sdks` target exists, because each shard now needs all the SDKs.
* Remove any build tags as part of updating the CI workflow (e.g. https://github.com/pulumi/pulumi-kubernetes/pull/3175). The test command on CI no longer uses build tags to shard, and you don't want tests excluded due to missing tags.
* (Optional) Ensure tests have globally unique names. Running `go run github.com/blampe/shard@latest --total 1 --index 0` will emit a warning for any tests with name collisions. Due to the way we're invoking `go test -run (names...) packages...`, it's possible for a test to get re-run on multiple shards if its name isn't unique, but this isn't the end of the world.

Refs https://github.com/pulumi/ci-mgmt/issues/676

### DIFF
--- a/native-provider-ci/src/steps.ts
+++ b/native-provider-ci/src/steps.ts
@@ -314,7 +314,7 @@ export function InstallPythonDeps(): Step {
 export function InstallSDKDeps(): Step {
   return {
     name: "Install dependencies",
-    run: "make install_${{ matrix.language}}_sdk",
+    run: "make install_sdks",
   };
 }
 
@@ -467,21 +467,38 @@ export function DownloadProviderBinaries(provider: string, job: string): Step {
   };
 }
 
-export function DownloadSDKs(): Step {
-  return {
-    name: "Download SDK",
-    uses: action.downloadArtifact,
-    with: {
-      name: "${{ matrix.language }}-sdk.tar.gz",
-      path: "${{ github.workspace}}/sdk/",
+// Download all SDKs if sharded
+export function DownloadSDKs(sharded: boolean): Step[] {
+  if (sharded) {
+    return ["nodejs", "go", "dotnet", "python", "java"].map((language) => {
+      return {
+        name: `Download ${language} SDK`,
+        uses: action.downloadArtifact,
+        with: {
+          name: `${language}-sdk.tar.gz`,
+          path: "${{ github.workspace}}/sdk/",
+        },
+      };
+    });
+  }
+
+  return [
+    {
+      name: "Download SDK",
+      uses: action.downloadArtifact,
+      with: {
+        name: "${{ matrix.language }}-sdk.tar.gz",
+        path: "${{ github.workspace}}/sdk/",
+      },
     },
-  };
+  ];
 }
 
 export function UnzipSDKs(): Step {
   return {
     name: "UnTar SDK folder",
-    run: "tar -zxf ${{ github.workspace}}/sdk/${{ matrix.language}}.tar.gz -C ${{ github.workspace}}/sdk/${{ matrix.language}}",
+    run: 'for f in *.tar.gz; do tar -zxf "$f" -C "${f%.tar.gz}"; done',
+    "working-directory": "${{ github.workspace}}/sdk",
   };
 }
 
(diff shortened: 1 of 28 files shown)


native-provider-ci/src/steps.ts
```diff
@@ -467,21 +467,38 @@ export function DownloadProviderBinaries(provider: string, job: string): Step {
...
-  };
+  ];
 }
 
 export function UnzipSDKs(): Step {
   return {
     name: "UnTar SDK folder",
-    run: "tar -zxf ${{ github.workspace}}/sdk/${{ matrix.language}}.tar.gz -C ${{ github.workspace}}/sdk/${{ matrix.language}}",
+    run: 'for f in *.tar.gz; do tar -zxf "$f" -C "${f%.tar.gz}"; done',
+    "working-directory": "${{ github.workspace}}/sdk",
```
blampe, line 501: This relaxes the logic to unpack any SDK tarballs that happen to be present, not just a language-specific tarball.

native-provider-ci/src/steps.ts
```diff
@@ -467,21 +467,38 @@ export function DownloadProviderBinaries(provider: string, job: string): Step {
...
+      return {
+        name: `Download ${language} SDK`,
+        uses: action.downloadArtifact,
+        with: {
+          name: `${language}-sdk.tar.gz`,
+          path: "${{ github.workspace}}/sdk/",
+        },
+      };
+    });
+  }
```
blampe, line 483: Each shard now needs every SDK.

native-provider-ci/src/steps.ts
```diff
@@ -314,7 +314,7 @@ export function InstallPythonDeps(): Step {
 export function InstallSDKDeps(): Step {
   return {
     name: "Install dependencies",
-    run: "make install_${{ matrix.language}}_sdk",
+    run: "make install_sdks",
```
blampe, line 317: I _think_ this is supported by our native providers, but I could also add a conditional here to be safe.

rquitales: I don't think neither `pulumi-kubernetes` nor `pulumi-docker-build` have a `install_sdks` entrypoint currently?

native-provider-ci/src/steps.ts
```diff
@@ -513,21 +530,32 @@ export function SetNugetSource(): Step {
...
-  if (provider === "kubernetes") {
+export function RunTests(name: string, shardDir?: string): Step[] {
+  if (shardDir !== undefined) {
     const shortMode = name === "run-acceptance-tests" ? " -short" : "";
-    const testCmd = `cd tests/sdk/\${{ matrix.language }} && go test -v -count=1 -cover -timeout 2h -parallel 4${shortMode} ./...`;
-    return {
-      name: "Run tests",
-      run: testCmd,
-    };
+    const shardCmd = `echo go test -v -count=1 -cover -timeout 2h -parallel 4${shortMode} "$(go run github.com/blampe/shard@b251cf8da6a83022628496125f285095772ebe86 --total \${{ strategy.job-total }} --index \${{ strategy.job-index }})" > test-command`;
```
rquitales, line 536: Do we plan on moving`github.com/blampe/shard` to a Pulumi controlled repo?

danielrbradley: Perhaps this should be an inline script so it's versioned per-provider via ci-mgmt? If not, then we should pin it.

danielrbradley (commented):
Following up on this following the work on the generic provider template (https://github.com/pulumi/ci-mgmt/pull/1140) and to clear this off my review backlog 😅  ...

I think we should hold off on adding the sharding here and instead focus on the generic template and moving the native providers to that.

Just to mirror the comments from elsewhere:
- In order to get this live, we should create a version of the shard program in the pulumi organisation.
- How are we planning to upgrade the version of the shard tool that we depend on?
- In the implementation of the shard tool - is there any way to leverage Go's own test discovery process so we're sure the tool behaves consistently with running `go test`?
- Related: I think there's potentially an issue here around when there's modules within modules.

PR #1059, opened 2024-08-22
Test master & release workflow on PR merge

danielrbradley:
- Use a merge queue to ensure XYZ is able to be released with the latest ci-mgmt changes before the change is committed into master to be rolled out to all other providers.
- The next step here would be to auto-tag the release and verify that also completes correctly.


provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/release.yml
```diff
@@ -1,4 +1,13 @@
 # WARNING: This file is autogenerated - changes will be overwritten if not made via https://github.com/pulumi/ci-mgmt
+name: release
+on:
+  push:
+    tags:
+    - v*.*.*
+    - "!v*.*.*-**"
+#{{- if .Config.testMasterAndReleaseWorkflows }}#
+  pull_request:
+#{{ end }}#
```
iwahbe, line 10: Is this just a move?

iwahbe (approved):
This looks ok. LGTM

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/release.yml
danielrbradley: Yup, just had to change something to be able to open a PR downstream

flostadler (approved):
lgtm

PR #1060, opened 2024-08-27
Permit bridged providers to have custom workflows

t0yv0:
Custom workflows like aws-run-upstream-tests.yml should not be managed centrally.

### DIFF
--- a/.github/workflows/update-workflows.yml
+++ b/.github/workflows/update-workflows.yml
@@ -68,9 +68,11 @@ jobs:
         run: cd pulumi-${{ inputs.provider_name }} && make upstream && cd ..
         continue-on-error: true
       - name: Delete existing workflows
-        # Bridged providers completely replace workflow files from ci-mgmt currently.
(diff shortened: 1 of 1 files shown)


iwahbe (approved):
I'm going to trust you have verified that the `find` innovation is correct.

.github/workflows/update-workflows.yml
```diff
@@ -68,9 +68,11 @@ jobs:
         run: cd pulumi-${{ inputs.provider_name }} && make upstream && cd ..
         continue-on-error: true
       - name: Delete existing workflows
-        # Bridged providers completely replace workflow files from ci-mgmt currently.
+        # Bridged providers completely replace workflow files from ci-mgmt currently except for custom workflows
+        # specific to the provider, named `provider*.yml` such as `aws-run-tests.yml`.
```
flostadler, line 72: This might be a bit hard to find given that it's documented in one of the workflows.
Maybe adding a workflow allowlist to the config makes it a bit clearer/intentional what's going on.

flostadler (approved):

PR #1063, opened 2024-08-28
Ability to select SDK languages

ringods:
**Update Sep 9**: dropped the `enableAutoRelease` from this PR. Will redo this as a separate PR.

Round 2 to add support for third-party providers:

* Fix: Pass the configured `organization` to 2 more places where it is needed
* `languages` (default: all supported languages): third party providers can configure a subset of the supported languages if they don't have the publishing infrastructure ready for all the languages. Changes:
  * changes the build matrix to only the set of configured languages
  * will install only the tools needed for the given value of the language matrix. Changed this because installing the Java tools failed when no Java SDK is available.

Screenshot 2024-08-28 at 15 00 07

The changes in this PR are focussed on the `main` workflow. Testing this on `pulumiverse/pulumi-acme` already got me this far:

https://github.com/pulumiverse/pulumi-acme/actions/runs/10596199445

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider.config.yaml
+++ b/provider-ci/internal/pkg/templates/bridged-provider.config.yaml
@@ -82,9 +82,17 @@ toolVersions:
   go: "1.21.x"
   java: "11"
   gradle: "7.6"
-  node: "20.x"
+  nodejs: "20.x"
   pulumi: "dev"
   python: "3.11.8"
+
+# Control which language SDKs get built and published.
+languages:
+  - nodejs
+  - python
+  - dotnet
+  - go
+  - java
 
 # env contains an assortment of properties for different purposes.
 # Additional entries are added by individual providers for different reasons.
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.yml
@@ -19,11 +19,7 @@ jobs:
       fail-fast: true
       matrix:
         language:
-        - nodejs
-        - python
-        - dotnet
-        - go
-        - java
+#{{ .Config.languages | toYaml | indent 8 }}#
     steps:
       #{{- if .Config.freeDiskSpaceBeforeSdkBuild }}#
       # Run as first step so we don't delete things that have just been installed
@@ -48,7 +44,7 @@ jobs:
       - name: Setup tools
         uses: ./.github/actions/setup-tools
         with:
-          tools: pulumictl, pulumicli, go, node, dotnet, python, java
+          tools: pulumictl, pulumicli, ${{ matrix.language }}
       - name: Download bin
         uses: ./.github/actions/download-bin
       - name: Install plugins
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Sometimes, you want those changes NOW rather than having to wait for a GitHub PR
 This example command will generate workflows for pulumi-datadog, and place them in the specified `--out` directory.
 Adjust for your provider and filesystem.
 
-```
+```bash
 ./bin/provider-ci generate --name pulumi/pulumi-datadog --template bridged-provider --config ./providers/datadog/config.yaml --out ../../pulumi-dtadog
 ```
 
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
@@ -143,10 +143,11 @@ jobs:
     - name: Setup tools
       uses: ./.github/actions/setup-tools
       with:
-        tools: pulumictl, pulumicli, go, node, dotnet, python, java
+        tools: pulumictl, pulumicli, ${{ matrix.language }}
     - name: Download bin
       uses: ./.github/actions/download-bin
     - run: dotnet nuget add source ${{ github.workspace }}/nuget
+      if: matrix.language == 'dotnet'
     - name: Download SDK
       uses: ./.github/actions/download-sdk
       with:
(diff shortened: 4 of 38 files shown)


provider-ci/internal/pkg/templates/provider/.github/workflows/command-dispatch.yml
```diff
@@ -1,3 +1,4 @@
+#{{- if .Config.enableAutoRelease -}}#
```
danielrbradley, line 1: This seems a little odd - this workflow is about auto-commenting, not auto-releasing.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.yml
```diff
@@ -48,7 +44,7 @@ jobs:
       - name: Setup tools
         uses: ./.github/actions/setup-tools
         with:
-          tools: pulumictl, pulumicli, go, node, dotnet, python, java
+          tools: pulumictl, pulumicli, ${{ matrix.language }}
```
danielrbradley, line 47: I like this simplification - should save a good few seconds on each task.

danielrbradley (commented):
Overall approach looks good. Need clarification on if the comment-dispatch change was accidental or if we need to adjust the settings name somehow.

Out of interest - why can't external providers run the auto-release hooks? Is it possible just to make this work for them too?

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.yml
ringods: And it works for my ACME test provider. Still, I consider this the most risky change to roll out for all of the Pulumi providers. You might want to check for possible implications.

ringods:
@danielrbradley AFAICS, the auto release setup depends on the Pulumi bot setup. If I want this to work for third party providers, I want to find a solution based on the standard github `permissions:` setting within a workflow. Can't solve everything at once unfortunately, so I chose to comment out more.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.yml
danielrbradley: The downstream tests are passing which gives me pretty high confidence that this is all good.

danielrbradley: The only possible way this could break is if there's any providers injecting arbitary commands (e.g. via setupScript) which assume all these tools are installed, but I _think_ that's pretty unlikely too.

provider-ci/internal/pkg/templates/provider/.github/workflows/command-dispatch.yml
danielrbradley: @ringods what was your thoughts here?

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.yml
danielrbradley: Huh, didn't notice that subtle difference ... it will work because it just does `contains(inputs.tools, 'node')` 🙃

provider-ci/internal/pkg/templates/bridged-provider.config.yaml
```diff
@@ -86,6 +89,14 @@ toolVersions:
   pulumi: "dev"
   python: "3.11.8"
 
+# Control which language SDKs get built and published.
+languages:
+  - nodejs
```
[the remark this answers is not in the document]
ringods: @guineveresaenger good catch! While the value was `nodejs` when specifying the value for the build matrix, it is `node` when passing it to the tool installation task. Would you be OK that I unify it to `node` everywhere?

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.yml
ringods: @danielrbradley but the naming is inconsistent for this language (see my comment below):

* `nodejs` when used in `${{ matrix.language }}` in the GHA workflow
* `build_nodejs` used ast the build target in the `Makefile`
* `node` when passed to the tool installation task

provider-ci/internal/pkg/templates/bridged-provider.config.yaml
```diff
@@ -75,6 +75,9 @@ freeDiskSpaceBeforeTest: false
 # Set to true to clear disk space before running sdk build jobs.
 freeDiskSpaceBeforeSdkBuild: false
 
+# Enables the automatic release of the provider
+enableAutoRelease: true
```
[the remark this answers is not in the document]
ringods: Earlier today, I had a short review call for this PR with @danielrbradley. To prevent adding a lot more configuration options to the setup, an alternative setup would be to create a separate template folder aimed at thirdparty bridged providers, reshuffle some files between the different folders and update `generate.go` here:

https://github.com/pulumi/ci-mgmt/blob/b6d3a9ffa9e3ffb274476425aa9668a2956f5d09/provider-ci/internal/pkg/generate.go#L69-L75

How does that sound to you?

provider-ci/internal/pkg/templates/bridged-provider.config.yaml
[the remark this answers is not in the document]
ringods: Unified to `nodejs` in 649b65f

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.yml
ringods: Unified to `nodejs` in 649b65f

danielrbradley (approved):
Happy with all these changes 👍

guineveresaenger (approved):

PR #1066, opened 2024-09-05
Run setup tools when uprading the bridge

VenelinMartinov:
The upgrade-provider tool needs all the dependencies set up.

will fix https://github.com/pulumi/pulumi-aws/issues/4431


danielrbradley:
As a piece of follow up work we can then remove the tool install step from the pulumi-upgrade-provider-action and bump the version in ci-mgmt. 

... on second thoughts, it might be worth doing that in this one PR so we're not running the tool install twice - which _should_ be okay, but I could imagine causing some kind of weird problem too!

VenelinMartinov:
This need https://github.com/pulumi/upgrade-provider/pull/276
https://github.com/pulumi/pulumi-aws/actions/runs/10722756001/job/29734431334

VenelinMartinov:
tested in https://github.com/pulumi/pulumi-aws/actions/runs/10725445843

PR #1068, opened 2024-09-05
Do not delete provider custom workflows in make ci-mgmt

t0yv0:
We introduced a notion of provider-custom workflow files, such as:

    aws-upstream-tests.yml

Currently CI knows to not delete them, but `make ci-mgmt` target still does. This PR fixes this so that `make ci-mgmt`
target matches the behavior of CI workflows.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/Makefile
+++ b/provider-ci/internal/pkg/templates/bridged-provider/Makefile
@@ -208,7 +208,9 @@ bin/pulumi-java-gen: .pulumi-java-gen.version
 # - Run make ci-mgmt to apply the change locally.
 #
 ci-mgmt: .ci-mgmt.yaml
-	rm -f .github/workflows/*.yml # Copied from update-workflows.yml
+	# Copied from update-workflows.yml
+        find pulumi-#{{ .Config.provider }}/.github/workflows/*.yml -type f ! -name '#{{ .Config.provider }}*.yml' -delete
+
 	go run github.com/pulumi/ci-mgmt/provider-ci@master generate \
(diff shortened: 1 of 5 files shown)


provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -208,7 +208,7 @@ bin/pulumi-java-gen: .pulumi-java-gen.version
 # - Run make ci-mgmt to apply the change locally.
 #
 ci-mgmt: .ci-mgmt.yaml
-	rm -f .github/workflows/*.yml # Copied from update-workflows.yml
+	find pulumi-$(PACK)/.github/workflows/*.yml -type f ! -name "$(PACK)*.yml" -delete
```
flostadler, line 211: Shouldn't this exclude the `pulumi-$(PACK)/` prefix?

```suggestion
	find .github/workflows/*.yml -type f ! -name "$(PACK)*.yml" -delete
```

t0yv0: You are right, good catch

danielrbradley (commented):
We've recently added a function to provider-ci to delete old files (outside of .github) which are no longer generated. Perhaps we should remove his wildcard selection completely and make the Go program responsible for cleaning up old files.

t0yv0:
Possibly yea - can I have this in the meanwhile? We're trying to keep the AWS workflow alive for running upstream tests.

danielrbradley (approved):

flostadler (approved):

PR #1069, opened 2024-09-05
Remove gotestfmt

t0yv0:
Engineers noticing gotestfmt reporting test output in a confusing way, such as mis-attributing logs from one test to another or not reporting logs if a panic is encountered. This change removes this utility, switching to default Go test output. It also removes the `-v` flag so that the test failure output is minimally scoped to the failing tests and is easier to find. Defaults for the win.


blampe (approved):

flostadler (approved):

guineveresaenger (approved):
Let's see what this looks like!

thomas11:
@t0yv0, @guineveresaenger Late to the party but we've had the same issues in azure-native and I had good initial result with tparse. [Here's a draft PR](https://github.com/pulumi/pulumi-azure-native/pull/3404); you can look at the test results. It's essentially just regular `go test` output but with a summary at the end.

PR #1072, opened 2024-09-06
Rremove gotestfmt 2 take 2

t0yv0:
Take 2 on removing gotestfmt. Since extension workflows still use it, retain the installation stanza, but not actually use in tests.

https://github.com/pulumi/pulumi-azure-native/pull/3404 is promising but to be evaluated later I think?


corymhall:
Also for evaluation later is [gotestsum](https://github.com/gotestyourself/gotestsum). It supports the github actions format and will stream the results so you see the results of the tests as they finish. I've been using it in pulumi-cdk and it seems to work well.

https://github.com/pulumi/pulumi-cdk/blob/0cad9088ca9007cc5942c578b8e62588df82f5bc/.github/workflows/run-acceptance-tests.yml#L101

iwahbe (approved):

PR #1076, opened 2024-09-10
Use configured `languages` for tool installation

ringods:
In `setup-tools`, the language `node` was renamed to `nodejs` for consistency. We missed it in 3 spots.

It should fix the failing `verify-release` in downstream `pulumi-xyz` provider:
* https://github.com/pulumi/pulumi-xyz/actions/runs/10789722716/job/29924862046
* https://github.com/pulumi/pulumi-xyz/actions/runs/10789596609/job/29923455243


danielrbradley (approved):
TY!

PR #1077, opened 2024-09-10
Ability to disable publishing to the Pulumi CDN

ringods:
Ability to disable publishing to the Pulumi CDN (`get.pulumi.com`) for non-Pulumi managed providers.

Ran `make test-providers` which didn't render any changes, but it worked for my `pulumiverse/pulumi-acme` provider.


guineveresaenger:
I apologize for the question, but what is Pulumi CDN?

This looks reasonable to me otherwise - I would like to see a third-party provider test in the repo maybe, so we can see the applied changes alongside the PR.

ringods:
@guineveresaenger I updated the initial comment. Pulumi CDN is `get.pulumi.com` which is not accessible for 3rd party packages.

provider-ci/internal/pkg/templates/bridged-provider.config.yaml
```diff
@@ -175,6 +175,8 @@ publish:
   # passed to the sdk input of pulumi-package-publisher
   # This is overridden in pulumi-local to disable python
   sdk: all
+  # Publish the plugin binaries to the Pulumi CDN
```
danielrbradley, line 178:
```suggestion
  # Publish the plugin binaries to the Pulumi CDN (get.pulumi.com) - requires AWS credentials for S3 upload
```

danielrbradley (approved):
Happy this is a nice safe change. I think this is very specifically just an s3 upload ... it's just this s3 bucket happens to be behind a cdn. I think with the expanded comment it's fine though.

ringods:
@guineveresaenger `acme` added as test provider.

PR #1078, opened 2024-09-10
Disarm any go:embed directives before running golangci-lint for native providers

rquitales:
The pulumi-kubernetes provider embeds a minified version of a Pulumi JSON schema in the built binary. This file is not checked into source, only the non minified version is. `golangci-lint` is unable to compile the provider for linting if the file does not exist. This PR ensures that we disable any go:embed directives by doing a simple find/replace.

This is necessary as `golangci-lint` is unable to skip processing files/directories. Instead, the `exclude-dirs/files` directives will still parse those files/directories, but just suppress warnings/errors. If it is not compilable, in this case due to missing embed files, the linter will also fail. See https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors

Seehttps://github.com/pulumi/pulumi-kubernetes/pull/3355 for example passing run that modifies the workflow file directly.

### DIFF
--- a/native-provider-ci/providers/command/repo/.github/workflows/build.yml
+++ b/native-provider-ci/providers/command/repo/.github/workflows/build.yml
@@ -474,6 +474,9 @@ jobs:
       with:
         go-version: ${{ env.GOVERSION }}
         cache-dependency-path: "**/*.sum"
+    - name: Disarm go:embed directives to enable linters that compile source code
+      run: git grep -l 'go:embed' -- provider | xargs --no-run-if-empty sed -i
+        's/go:embed/ goembed/g'
     - name: golangci-lint provider pkg
       uses: golangci/golangci-lint-action@d6238b002a20823d52840fda27e2d4891c5952dc # v4.0.1
       with:
--- a/native-provider-ci/providers/command/repo/.github/workflows/run-acceptance-tests.yml
+++ b/native-provider-ci/providers/command/repo/.github/workflows/run-acceptance-tests.yml
@@ -371,6 +371,9 @@ jobs:
       with:
         go-version: ${{ env.GOVERSION }}
         cache-dependency-path: "**/*.sum"
+    - name: Disarm go:embed directives to enable linters that compile source code
+      run: git grep -l 'go:embed' -- provider | xargs --no-run-if-empty sed -i
+        's/go:embed/ goembed/g'
     - name: golangci-lint provider pkg
       uses: golangci/golangci-lint-action@d6238b002a20823d52840fda27e2d4891c5952dc # v4.0.1
       with:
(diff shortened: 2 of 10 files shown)


blampe:
@rquitales would it be easier to just check these in? They aren't that big, but we could use LFS if size is a concern. Can also add to `.gitattrbutes` to exclude from diffs to keep noise down.

rquitales:
> @rquitales would it be easier to just check these in? They aren't that big, but we could use LFS if size is a concern. Can also add to `.gitattrbutes` to exclude from diffs to keep noise down.

@blampe Is there a way to exclude from diffs to keep noise down? I'm not familiar with the configuration for that. AFAIK I can skip diffing/committing as a local config, but it'd be nice to have that as a global setting.

blampe:
> > @rquitales would it be easier to just check these in? They aren't that big, but we could use LFS if size is a concern. Can also add to `.gitattrbutes` to exclude from diffs to keep noise down.
> 
> @blampe Is there a way to exclude from diffs to keep noise down? I'm not familiar with the configuration for that. AFAIK I can skip diffing/committing as a local config, but it'd be nice to have that as a global setting.

@rquitales I was referring to [this](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github) to hide SDK changes by default, but that still commits the file. I think this would be fine, but I don't have all the background behind these files.
```
~/src/pulumi-kubernetes master*                                                                                                                                                                                                            
❯ cat .gitattributes
sdk/**/* linguist-generated=true
provider/cmd/pulumi-resource-kubernetes/schema.json linguist-generated=true
```

danielrbradley (commented):
Is there a way to change this so that so we don't need to add a special provider-specific branch? Would running something like `make ensure` work? Feel like most native providers might have that already?

Azure Native also had a similar problem but I can't remember exactly how we solved it there.

blampe:
Just ran into this. I merged https://github.com/pulumi/pulumi-kubernetes/pull/3292 but I think its tests only passed because we had the build artifacts cached after manually touching these files.

Anyway - they should be cached again so builds should continue to pass... until we touch `go.mod` or something. We should probably revisit this soon.

rquitales:
I have changed the implementation to not touch files, but remove the go:embed directives from source code prior to linting as a more general approach to this problem. This is inline with what we do in the bridged providers.

blampe (approved):
Not pretty but it works.

danielrbradley (approved):
Looks good - assuming we've manually tested this on one provider before rolling out to all of them 👍

rquitales:
This was manually tested in https://github.com/https://github.com/pulumi/pulumi-kubernetes/pull/3355, and Kubernetes is the only native provider managed by ci-mgmt that is running lint - so blast radius should be limited.

PR #1080, opened 2024-09-13
Allow skipping the Go SDK verification when invoking directly

danielrbradley:
We can't test Go SDKs for pre-releases because we don't publish the Go SDKs for them. This is already implemented when invoked via a workflow_call, but the option wasn't available when using a workflow_dispatch (manual run).

### DIFF
--- a/provider-ci/internal/pkg/templates/provider/.github/workflows/verify-release.yml
+++ b/provider-ci/internal/pkg/templates/provider/.github/workflows/verify-release.yml
@@ -11,6 +11,11 @@ on:
         description: "Enable the MacOS runner in addition to Linux and Windows. Defaults to 'false'."
         required: false
         type: boolean
+      skipGoSdk:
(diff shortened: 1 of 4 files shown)


provider-ci/internal/pkg/templates/provider/.github/workflows/verify-release.yml
```diff
@@ -11,6 +11,11 @@ on:
         description: "Enable the MacOS runner in addition to Linux and Windows. Defaults to 'false'."
         required: false
         type: boolean
+      skipGoSdk:
+        description: "Skip the Go SDK verification. Defaults to 'false'. This is used when we're not publishing a Go SDK on the default branch build."
```
thomas11, line 15: Could the description a bit more actionable, e.g.

_Skip the Go SDK verification. Defaults to 'false'. Enable this when running the workflow on a pre-release. We're not publishing a Go SDK on the default branch build._

thomas11 (approved):

flostadler (approved):

PR #1081, opened 2024-09-13
Fix test tool setup

danielrbradley:
Mirror the fix for #1079 into the nightly, main, prerelease and release jobs.

Test is the only core part of our workflow which has not been deduplicated and so we have to copy fixes into all the different workflows.

Spotted from issue https://github.com/pulumi/pulumi-azure/issues/2393


flostadler (approved):
lgtm!

PR #1084, opened 2024-09-27
Support external bridged provider by reducing Pulumi Corp targeted workflows

ringods:
This PR contains a set of changes to better support external bridged providers. Such external providers need less of the workflows used by the Pulumi Corp engineering team.

## Single set of default configuration values.

The current setup uses a default config file based on the template name: `bridged-provider` template name maps to `provider-ci/internal/pkg/templates/bridged-provider.config.json`.  The template name then points to a set of folders containing the actual template files:

https://github.com/pulumi/ci-mgmt/blob/52751c4fab81997ddc61b571ef8102e4c864d933/provider-ci/internal/pkg/generate.go#L70-L72

These template files use templating based on the configuration options to tweak the generated content for each provider.

If we want to reuse most of these templates for another template name, we must prevent duplicating the config options in a second file. After discussing this with @danielrbradley, the idea was to not have a default config file per template name, but a single default config file applicable to all templates.

## Default repository name calculation.

The CLI arguments to `provider-ci generate` has the following [currently documented](https://github.com/pulumi/ci-mgmt/blob/52751c4fab81997ddc61b571ef8102e4c864d933/provider-ci/internal/cmd/generate.go#L69):

```sh
  -n, --name string       repository name to generate (default "{config.repository}" or otherwise "pulumi/pulumi-{config.provider}")
```

However, nowhere in the code is a default of "pulumi/pulumi-{config.provider}" calculated. One of the commits in this PR resolves this and improves on it to take the `organization` config option into account.

By calculating this value based on the `provider` and `organization` config values, we can run `provider-ci generate` now without the `--name` CLI argument. In `make test-providers`, it fixes that the `acme` provider is no longer assumed to be in the `pulumi` org, but correctly configured to be in the `pulumiverse` organization as specified by its `.ci-mgmt.yml` file.

## Moved some specific into a Pulumi (Corp) folder

Some of the workflows & issue templates have been moved to a dedicated Pulumi folder. This folder is added to the default `bridged-provider` template, but not used for `external-bridged-provider`. This reduces the amount and complexity of the workflows for an external provider. An external provider doesn't have a bot running and doesn't require (for now) a few of the repository dispatch scenarios.


guineveresaenger (approved):
LGTM, thank you!

danielrbradley (approved):
🚀

PR #1085, opened 2024-09-27
Run registry docs command from correct file location

guineveresaenger:
Previous version was running `make` from `provider` folder. This is the correct location at repo root.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/Makefile
+++ b/provider-ci/internal/pkg/templates/bridged-provider/Makefile
@@ -114,7 +114,7 @@ build_python: upstream
 #{{- if .Config.registryDocs }}#
 
 build_registry_docs:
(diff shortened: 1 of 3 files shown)


t0yv0:
A few notes on how you tested this would be helpful!

iwahbe (commented):
Does this mean that `build_registry_docs` was never exercised from CI?

Should we add a note that says this is for local development?

Can you please add a comment to `build_registry_docs` that explains what it does?

guineveresaenger:
@iwahbe - since this command is behind a flag as per https://github.com/pulumi/ci-mgmt/pull/1083, it has not been run live in CI. 
@t0yv0 - [this commit](https://github.com/pulumi/pulumi-venafi/pull/433/commits/e0fd9262b81bf254ba0607d1ec63c9545f182e13) reflects the changes of this in action, but won't run with ci-mgmt as-is.

iwahbe (approved):

Issue #1091, opened 2024-10-01
AttributeError: module 'pulumi.runtime' has no attribute 'invoke_output'


t0yv0:
Remediate "AttributeError: module 'pulumi.runtime' has no attribute 'invoke_output'".

From @justinvp:

These are the relevant changes and releases:
https://github.com/pulumi/pulumi/pull/17237 - Core SDK and sdkgen change for Node.js (released in v3.133.0)
https://github.com/pulumi/pulumi/pull/17275 - Core SDK and sdkgen change for Python (released in v3.134.0)
https://github.com/pulumi/pulumi/pull/17316 - Bump Node.js sdkgen’s min version to v3.133.0 (released in v3.134.0)
https://github.com/pulumi/pulumi/pull/17353 - Fix bug in Node.js Core SDK (short-circuit output invokes if inputs contains unknowns) (released in v3.134.1)
https://github.com/pulumi/pulumi/pull/17390 - Bump Node.js and Python sdkgen’s min version to v3.134.1 (released in v3.135.0)

Summary:
v3.133.0 sdkgen is problematic for Node.js. It needs v3.135.0 sdkgen and to not be setting the explicit @pulumi/pulumi dependency so that sdkgen emits the default min version.
v3.134.0 sdkgen is problematic for Python. It needs v3.135.0 sdkgen and to not be setting the explicit pulumi dependency so that sdkgen emits the default min version.

To remediate we need to make sure to remove explicit p/p dependency, for example:

- https://github.com/pulumi/pulumi-random/blob/67ad6fefe519943a878a44181fd6d957ceb5d726/provider/resources.go#L82
and
- https://github.com/pulumi/pulumi-random/blob/67ad6fefe519943a878a44181fd6d957ceb5d726/provider/resources.go#L93

Also please remove it for .NET

## Native Providers

- [x] [aws-apigateway](https://github.com/pulumi/pulumi-aws-apigateway) (not managed)
    - https://github.com/pulumi/pulumi-aws-apigateway/pull/166
- [x] [aws-native](https://github.com/pulumi/pulumi-aws-native)
- [x] [awsx](https://github.com/pulumi/pulumi-awsx) (not managed)
    - https://github.com/pulumi/pulumi-awsx/pull/1390
- [x] [azure-native](https://github.com/pulumi/pulumi-azure-native) (not managed)
  - https://github.com/pulumi/pulumi-azure-native/pull/3611 removes the bug and the urgency
[...]

PR #1092, opened 2024-10-02
Migration to remove explicit SDK dependencies for python and JS

VenelinMartinov:
This adds a migration for providers to remove the explicit SDK dependencies for python and javascript.

this should solve https://github.com/pulumi/ci-mgmt/issues/1091 for tier 2+ providers

I've tested this on pulumi-cloudflare and pulumi-random. I've also tested it doesn't do anything the second time around.

### DIFF
--- a/tools/sourcemigrator/index.ts
+++ b/tools/sourcemigrator/index.ts
@@ -99,6 +99,33 @@ function run(command: string): shell.ShellString {
 //   };
 // }
 
+function removeExplicitSDKDependency(): SourceMigration {
+  return {
+    name: "Respect Schema Version",
+    async execute(context) {
+      const patchPath = fs.realpathSync("removeExplicitSDKdependency.patch");
+      shell.pushd(context.dir);
+      try {
+        // Apply patch
+        run(
+          `go run github.com/uber-go/gopatch@latest -p "${patchPath}" ./provider/resources.go`
+        );
+        // Format the code - twice to ensure that the code is formatted correctly
+        run(`go install mvdan.cc/gofumpt@latest`);
+        run(`gofumpt -w ./provider/resources.go`);
+        run(`gofumpt -w ./provider/resources.go`);
+        // Check if we've made changes
+        const gitStatus = run(`git status --porcelain`).stdout;
+        if (gitStatus.includes("provider/resources.go")) {
+          run(`make tfgen build_sdks`);
+        }
+      } finally {
+        shell.popd();
+      }
+    },
+  };
+}
+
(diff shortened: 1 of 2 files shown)


tools/sourcemigrator/index.ts
```diff
@@ -99,6 +99,33 @@ function run(command: string): shell.ShellString {
...
+function removeExplicitSDKDependency(): SourceMigration {
+  return {
+    name: "Respect Schema Version",
+    async execute(context) {
+      const patchPath = fs.realpathSync("removeExplicitSDKdependency.patch");
+      shell.pushd(context.dir);
+      try {
+        // Apply patch
+        run(
+          `go run github.com/uber-go/gopatch@latest -p "${patchPath}" ./provider/resources.go`
```
danielrbradley, line 111: Nit: we could just pin the version of gopatch to ensure the behaviour doesn't change in case a bad release goes out.

tools/sourcemigrator/index.ts
```diff
@@ -99,6 +99,33 @@ function run(command: string): shell.ShellString {
...
+        );
+        // Format the code - twice to ensure that the code is formatted correctly
```
flostadler, line 113: why twice? Shouldn't it do the right thing the first time around?

danielrbradley (approved):
Approach looks sound to me.

Note: This sourcemigrator will only be run for providers where ci-mgmt pushes the change out. For anyone (such as pulimiverse providers) using the pull-based update model, they will have to make this change manually. At some point, we should fold this into the `provider-ci` binary.

flostadler (approved):
LGTM

tools/sourcemigrator/index.ts
VenelinMartinov: I am not sure, I've followed the example of previous migrations: https://github.com/pulumi/ci-mgmt/blob/d1aac1f630c313324cb53267ef73ec9fb0910d9e/tools/sourcemigrator/index.ts#L89

I am assuming that some changes can lead to further formatting changes.

VenelinMartinov: @danielrbradley do you have any memory of when this happens? I am also curious.

danielrbradley: Can't remember exactly what the scenario was, but I remember it definitely needed the second run to get it right 🙃

t0yv0:
It could be useful to include an ex-post-facto check that finds schema.json, extracts the languages section and verifies that it is clean and does not have the undesirable elements.

PR #1093, opened 2024-10-02
Disable triggering a release based on issue/pr labels

ringods:
Summary of changes:

* Thirdparty providers do not have the Pulumi Bot active. Any workflow step making use of it should be skipped if no such endpoint has been defined.
* Providers need the `--repository` option set for `schema-tools` to fetch the schema in the correct Github organization

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/publish.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/publish.yml
@@ -73,7 +73,7 @@ jobs:
         {
           echo 'summary<<EOF'
           if [[ "$LAST_VERSION" != "No stable release" ]]; then
-            schema-tools compare --provider="#{{ .Config.provider }}#" --old-commit="$LAST_VERSION" --new-commit="--local-path=provider/cmd/pulumi-resource-#{{ .Config.provider }}#/schema.json"
+            schema-tools compare --provider="#{{ .Config.provider }}#" --old-commit="$LAST_VERSION" --repository="github://api.github.com/#{{ .Config.organization }}#" --new-commit="--local-path=provider/cmd/pulumi-resource-#{{ .Config.provider }}#/schema.json"
           fi
           echo 'EOF'
         } >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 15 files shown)


provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/publish.yml
```diff
@@ -73,7 +73,7 @@ jobs:
         {
           echo 'summary<<EOF'
           if [[ "$LAST_VERSION" != "No stable release" ]]; then
-            schema-tools compare --provider="#{{ .Config.provider }}#" --old-commit="$LAST_VERSION" --new-commit="--local-path=provider/cmd/pulumi-resource-#{{ .Config.provider }}#/schema.json"
+            schema-tools compare --provider="#{{ .Config.provider }}#" --old-commit="$LAST_VERSION"#{{ if ne .Config.organization "pulumi" }}# --repository="github://api.github.com/#{{ .Config.organization }}#"#{{ end }}# --new-commit="--local-path=provider/cmd/pulumi-resource-#{{ .Config.provider }}#/schema.json"
```
danielrbradley, line 76: What about always using the config organization - one less conditional where only 1 branch might actually ever be tested.

```suggestion
            schema-tools compare --provider="#{{ .Config.provider }}#" --old-commit="$LAST_VERSION" --repository="github://api.github.com/#{{ .Config.organization }}#" --new-commit="--local-path=provider/cmd/pulumi-resource-#{{ .Config.provider }}#/schema.json"
```

danielrbradley (approved):
Looks good overall, but would avoid conditionals if we can help it. Know it will cause a large diff, but think it'll be easier to maintain.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/publish.yml
ringods: @danielrbradley That's what I had in the beginning, but I did the extra work to prevent unnecessary changes to the Pulumi managed providers. I will get rid of the `if` you like that better.

ringods:
This run on `acme` provider showcasing it skips the `Tag release if labeled as needs-release` because the `RELEASE_BOT_ENDPOINT` env variable is not defined.

PR #1094, opened 2024-10-04
Permit custom Make targets

t0yv0:
The standard workflows are great, but sometimes build-time complexity is a necessary evil for temporary or experimental features. One concrete use case is simplifying computing the minimal schema in AWS:

     https://github.com/pulumi/pulumi-aws/pull/4587

This small change permits providers to define *.mk files to add some provider-specific extensions to Make.


VenelinMartinov (approved):
Looks good, assuming it passes on providers with no custom makefiles

PR #1097, opened 2024-10-07
Inline upgrade provider action

danielrbradley:
This action was originally failing due to missing dotnet if we run the disk clean-up action first. Then when trying to install dotnet using the setup-tools action, it was failing because the repo wasn't yet checked out, however, the `pulumi-upgrade-provider-action` does it's own checkout action, so it would have required removing the checkout from that action, creating a new version, then upgrading the version via ci-mgmt. At this point the `pulumi-upgrade-provider-action` would only contain the install and execution of the upgrade-provider CLI tool.

## Change summary

- Go direct to the `upgrade-provider` program instead of via the `pulumi-upgrade-provider-action` wrapper action.
- Enables us to use the shared setup-tools action for tool consistency.

## References

- Existing wrapper action: https://github.com/pulumi/pulumi-upgrade-provider-action/blob/main/action.yml
- Sample run: https://github.com/pulumi/pulumi-azure/actions/runs/11219063931/job/31184185595

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-provider.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-provider.yml
@@ -18,19 +18,28 @@ jobs:
           swap-storage: false
           dotnet: false
       #{{- end }}#
+      - name: Checkout Repo
+        uses: #{{ .Config.actionVersions.checkout }}#
+        with:
+          depth: 0 # Fetch full history for checking previous tags
+          #{{- if .Config.checkoutSubmodules }}#
+          submodules: #{{ .Config.checkoutSubmodules }}#
+          #{{- end }}#
       - name: Setup tools
         uses: ./.github/actions/setup-tools
         with:
           tools: pulumictl, pulumicli, go, nodejs, dotnet, python, java
-      - name: Call upgrade provider action
-        uses: #{{ .Config.actionVersions.upgradeProviderAction }}#
-        with:
-          kind: all
-          #{{- if .Config.javaGenVersion }}#
-          target-java-version: #{{ .Config.javaGenVersion }}#
-          #{{- end }}#
-          email: bot@pulumi.com
-          username: pulumi-bot
+      - name: Install upgrade-provider
+        run: go install github.com/pulumi/upgrade-provider@main
+        shell: bash
+      - name: "Set up git identity: name"
+        run: |
+          git config --global user.name 'bot@pulumi.com'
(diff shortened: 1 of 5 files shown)


thomas11 (approved):

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-provider.yml
```diff
@@ -18,19 +18,27 @@ jobs:
...
-        uses: #{{ .Config.actionVersions.upgradeProviderAction }}#
-        with:
-          kind: all
-          #{{- if .Config.javaGenVersion }}#
-          target-java-version: #{{ .Config.javaGenVersion }}#
-          #{{- end }}#
-          email: bot@pulumi.com
-          username: pulumi-bot
+      - name: Install upgrade-provider
+        run: go install github.com/pulumi/upgrade-provider@main
```
guineveresaenger, line 32: I am enthusiastic about running this on latest but have we looked at the guard rails for merging updates to upgrade-provider?

guineveresaenger (approved):
This is a good consolidation I think. 👍 

This would effectively sunset the `upgrade-provider-action`.
I'd like to see the following cleanup, or followup issue:

- Removal of `upgradeProviderAction` from `actionVersions` config
- Mark the action repository as no longer actively maintained. A redirect to this inline implementation perhaps? Or to `upgrade-provider` tool, with example steps in the README there.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-provider.yml
danielrbradley: This maintains the current behaviour for now which can be handy for making quick fixes and instantly rolling out everywhere. Perhaps we could move to pushing an explicit `v1` branch for rollouts once we're happy it's stable and just pinned to a major version similar to our GitHub Actions. Probably also good to bolster the testing of the upgrade-provider tool somehow.

danielrbradley:
> This is a good consolidation I think. 👍
> 
> This would effectively sunset the `upgrade-provider-action`. I'd like to see the following cleanup, or followup issue:
> 
> * Removal of `upgradeProviderAction` from `actionVersions` config
> * Mark the action repository as no longer actively maintained. A redirect to this inline implementation perhaps? Or to `upgrade-provider` tool, with example steps in the README there.

Yes, that's the direction I'm wanting to go. It's still currently used in the `upgrade-bridge` workflow and `check-upstream-upgrade` workflow. Plan will be to rework these too and then sunset the settings and repo.

Issue #963, opened 2024-06-09
Ensure that the "Check Upstream" job uses the Github Token from the enclosing repo, not a Org scoped token


iwahbe:
We have seen some "Check Upstream" jobs fail, hitting GH rate limits. It looks like they are hitting Pulumi org level rate limits, when the job could be done with a repo level token (and a repo level rate limit).

Issue #1087, opened 2024-10-01
Use built-in token when checking for upstream updates


danielrbradley:
We're only creating an issue, so don't need full bot access.

PR #1098, opened 2024-10-08
From bot token to per workflow generated token

ringods:
Fixes: #963, #1087

Moving from a central `PULUMI_BOT_TOKEN` to the Github Actions `permissions` block makes the workflows reusable for third party providers too. The generated `GITHUB_TOKEN` secret in each workflow instance will receive elevated permissions based on the `permissions` configuration block.

Docs on `permission`:
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token

Results of testing the changes on third-party providers:

* the required permissions for the `publish` workflow are set on the calling side of the nested job instead of within the nested job. I tested setting it within `publish.yml` but it seems it is not possible anymore to elevate permissions in a nested job: https://github.com/pulumiverse/pulumi-acme/actions/runs/11181469935


provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/publish.yml
```diff
@@ -93,7 +93,7 @@ jobs:
         generate_release_notes: true
         files: dist/*
       env:
-        GITHUB_TOKEN: ${{ secrets.PULUMI_BOT_TOKEN }}
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
danielrbradley, line 96: There's no permissions set on this job - it needs to push a new release.

danielrbradley (commented):
Looks good on first pass. We should pick an internal provider to roll this out to and verify all the key workflows work: upgrade-provider through to release.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/publish.yml
danielrbradley: It's from the parent workflow 👍  Let's add some more inline doc comments.

ringods:
This PR contains too much at once. So chopping it up.

Replacements: 
- #1099

Will close this PR when all functionality is submitted as separate PRs.

ringods:
Superseded by:
- #1099 
- #1102 
- #1109 
- #1110
- #1111

Issue #1053, opened 2024-08-13
making this usable outside pulumi org


omercnet:
this repo is wonderful, thanks for the hard work!

would you considering accepting a PR that adds a flag for non-pulumians? :)
this would exclude all the pulumi org specific steps and configurations (like publishing blobs to your azure/s3 in go releaser, aws auth to your account, etc.)

mjeffryes:
Thanks for the suggestion @omercnet! We definitely want to publish more resources to make it easy for anyone to maintain Pulumi providers. I'm not 100% sure we'll long-term choose the route of publishing a version of this repo that's designed for broad consumption, but in the meantime, we'd be happy to consider PRs to make some of the Pulumi specific things configurable (like thisone: https://github.com/pulumi/ci-mgmt/pull/1056) to make this usable by others!

omercnet:
That was gonna be my first PR :) 

I'll add more as you seem open to it 
Thanks!

ringods:
@omercnet I'm making progress on this one. See also my comment here: https://github.com/pulumiverse/infra/issues/40#issuecomment-2399564404

PR #1099, opened 2024-10-08
Use the standard Github generated token with elevated permissions

ringods:
Relates to: #1053, #1087

Moving from a central `PULUMI_BOT_TOKEN` to the Github Actions permissions block makes the workflows reusable for third party providers. The generated `GITHUB_TOKEN` secret in each workflow instance will receive elevated permissions based on the permissions configuration block. To limit the blast radius of a possible error, this PR focuses on the `main|master` and `(pre)release` workflows for now. Similar changes for other workflows will come in subsequent pull requests.

Besides making the workflows more reusable for third-party providers, it also improves on the situation for rate limits tied to the central `PULUMI_BOT_TOKEN`. The generated `GITHUB_TOKEN` has [much higher API rate limits](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-github_token-in-github-actions) on our Github Enterprise backed subscription for `pulumi` organization:

> The rate limit for `GITHUB_TOKEN` is 1,000 requests per hour per repository. For requests to resources that belong to a GitHub Enterprise Cloud account, the limit is 15,000 requests per hour per repository.

Already moving the `main|master` and `(pre)release` workflows to use the `GITHUB_TOKEN` reduces the usage of the `PULUMI_BOT_TOKEN`, so lowering the chance of bumping into a rate limit.

Github Docs: [Controlling permissions for `GITHUB_TOKEN`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token)

**NOTE:** The required permissions for the `publish` workflow are set on the calling side of the nested job instead of within the nested job. I tested setting it within publish.yml but it seems it is not possible anymore to elevate permissions in a nested job: https://github.com/pulumiverse/pulumi-acme/actions/runs/11181469935

The changes of this PR are validated in the following third-party packages:

| Package | Commit with changes from this PR | Green `main` run | Green `release` |
|--------|--------|--------|--------|
| [pulumiverse/pulumi-acme](https://github.com/pulumiverse/pulumi-acme) | [33e7fa5](https://github.com/pulumiverse/pulumi-acme/commit/33e7fa53d999e8eee09fd19bdb1d2046b58021ce) | [11182399934](https://github.com/pulumiverse/pulumi-acme/actions/runs/11182399934) | [11182413642](https://github.com/pulumiverse/pulumi-acme/actions/runs/11182413642) |
| [pulumiverse/pulumi-matchbox](https://github.com/pulumiverse/pulumi-matchbox) | [f370a12](https://github.com/pulumiverse/pulumi-matchbox/commit/f370a122007a617e5449fa9f6e2f97db75af9532) | [11232745061](https://github.com/pulumiverse/pulumi-matchbox/actions/runs/11232926252) | [11182413642](https://github.com/pulumiverse/pulumi-matchbox/actions/runs/11232926252) |

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
@@ -82,6 +82,8 @@ jobs:
 
   publish:
     name: publish
+    permissions:
+      contents: write
     needs:
       - prerequisites
       - build_provider
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/prerelease.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/prerelease.yml
@@ -41,6 +41,8 @@ jobs:
 
   publish:
     name: publish
+    permissions:
+      contents: write
     needs:
       - prerequisites
       - build_provider
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/publish.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/publish.yml
@@ -93,7 +93,7 @@ jobs:
         generate_release_notes: true
         files: dist/*
       env:
-        GITHUB_TOKEN: ${{ secrets.PULUMI_BOT_TOKEN }}
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
   publish_sdk:
     name: publish_sdk
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/release.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/release.yml
@@ -49,6 +49,8 @@ jobs:
 
   publish:
     name: publish
+    permissions:
+      contents: write
     needs:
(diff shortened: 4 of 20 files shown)


provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
```diff
@@ -82,6 +82,8 @@ jobs:
 
   publish:
     name: publish
+    permissions:
+      contents: write
```
danielrbradley, line 86: Nit: Do we actually need this for the main workflow as we don't actually create releases from here?

danielrbradley (approved):

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
danielrbradley: Let's keep it consistent in case there's other uses of the tokens.

danielrbradley:
Merging as this is a fairly small area of impact. Merging will release the xyz provider to prove this is all functional. If that release fails, we'll roll this back.

PR #1103, opened 2024-10-15
Migration to fixup bridge imports

blampe:
Supersedes https://github.com/pulumi/pulumi-upgrade-provider-action/pull/34.

Adds a migration to update imports to point at new paths introduced by  https://github.com/pulumi/pulumi-terraform-bridge/pull/2480 (after that's released).


iwahbe (approved):
Lets give it a shot.

danielrbradley:
Happy with this in principle.

Have you tested this on a handful of providers locally already?

We should probably have a way of tracking a migration roll-out and flagging any issues.

blampe:
> Have you tested this on a handful of providers locally already?

Tested locally as well as on https://github.com/pulumi/pulumi-gcp/pull/2593 https://github.com/pulumi/pulumi-datadog/pull/683 https://github.com/pulumi/pulumi-cloudflare/pull/955 (I also tried AWS a few times before realizing Anton had already migrated it :)).

> We should probably have a way of tracking a migration roll-out and flagging any issues.

@danielrbradley not sure if there's a better way to track, but there are currently 34 uses of the deprecated pf module which we expect to go to zero. I can kick off an "update workflows" actions and then see where things stand after those PRs have had a chance to merge.
https://github.com/search?q=org%3Apulumi+path%3Ago.mod+%22github.com%2Fpulumi%2Fpulumi-terraform-bridge%2Fpf+%22&type=code

Issue #1108, opened 2024-10-24
Restructure upgrade provider process


danielrbradley:
## Aim
Remove triggering of upgrade PR creation on the issue creation.

## Problem
Triggering of the PR creation off the issue creation requires limiting to issues created by a known user to avoid accidental triggering. This has to be a real GitHub user and not a workflow user which is then not easy to set up for third parties.

## Solution

1. Have a single workflow which is triggered on a schedule.
2. If a new version exists, ensure an issue exists for that version.
3. Attempt to automatically open a PR.

PR #1110, opened 2024-10-28
Combine upstream check & upgrade workflows

danielrbradley:
Remove the chain of dependencies where we create issues then trigger the actual upgrade from the issue creation. We previously had to filter the triggered issue opening to a specific bot user to stop other issues created by external users from triggering the workflow.

Instead, we'll check for new versions then attempt the upgrade in a single run, as well as allowing the execution of a specific version upgrade via workflow_dispatch.

Scenarios:

- If the cron triggers or a user runs the workflow without a version, it'll ensure an issue exists for the latest pending version and return the latest pending version in the `latest_version` output.
- If `latest_version` is unset then we don't need to run the upgrade as there's no pending version.
- If `latest_version` is set then we'll specifically attempt an upgrade to that version.
- If a user runs the workflow with a version input, we'll skip creating issues and just attempt the upgrade.

This builds on the work in https://github.com/pulumi/upgrade-provider/pull/282

Fixes https://github.com/pulumi/ci-mgmt/issues/1108

Example run with no new changes: https://github.com/pulumi/pulumi-azure/actions/runs/11608904415/job/32325098437


danielrbradley:
Attempting to test this on the xyz provider: https://github.com/pulumi/pulumi-xyz/actions/runs/11556612896/job/32164701827

This fails due to the upstream provider having no published stable versions.

ringods (approved):

VenelinMartinov (commented):
Looks sensible. Can we test the workflow where there is an upstream provider upgrade available somewhere? Ideally we should make sure not to miss any of the upgrade issues as they are essential for visibility, reporting etc.

To make sure i understand the change - the upstream upgrade no longer depend on the upgrade issue, is that correct? Instead the upgrade issue is now purely informational but is still opened if the workflow is run via a cron job, right?

danielrbradley:
> Looks sensible. Can we test the workflow where there is an upstream provider upgrade available somewhere? Ideally we should make sure not to miss any of the upgrade issues as they are essential for visibility, reporting etc.

Was testing this on OCI but Ian was too speedy getting the release done. It was looking mostly fine though having fixed up one bug in upgrade-provider.

> To make sure i understand the change - the upstream upgrade no longer depend on the upgrade issue, is that correct? Instead the upgrade issue is now purely informational but is still opened if the workflow is run via a cron job, right?

Yup - any time the workflow is run without an explicit version input (including when run on a cron) it will open any required issues, then attempt to upgrade to the latest version.

t0yv0 (approved):

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-provider.yml
```diff
@@ -1,11 +1,22 @@
 # WARNING: This file is autogenerated - changes will be overwritten if not made via https://github.com/pulumi/ci-mgmt
 
+name: Upgrade provider
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "The version of the upstream provider to upgrade to, without the 'v' prefix"
```
iwahbe, line 8:
```suggestion
        description: |
          The version of the upstream provider to upgrade to, without the 'v' prefix
          
          If no version is specified, it will be inferred from the upstream provider's release tags.
```

iwahbe (approved):

ringods (approved):

PR #1112, opened 2024-10-29
Fix cert-manager's SDK location

blampe:
Fixes https://github.com/pulumi/pulumi-kubernetes-cert-manager/issues/215

### DIFF
--- a/native-provider-ci/providers/kubernetes-cert-manager/config.yaml
+++ b/native-provider-ci/providers/kubernetes-cert-manager/config.yaml
@@ -1,6 +1,7 @@
 provider: kubernetes-cert-manager
 pulumiVersionFile: .pulumi.version
 major-version: 0
+sdkModuleDir: .
 env:
   AWS_REGION: us-west-2
   PULUMI_TEST_OWNER: moolumi
--- a/native-provider-ci/providers/kubernetes-cert-manager/repo/.github/workflows/prerelease.yml
+++ b/native-provider-ci/providers/kubernetes-cert-manager/repo/.github/workflows/prerelease.yml
@@ -563,11 +563,8 @@ jobs:
       with:
         repository: ${{ github.repository }}
         base-ref: ${{ github.sha }}
-        source: sdk
-        path: sdk
+        source: .
+        path: .
         version: ${{ steps.version.outputs.version }}
         additive: false
-        files: |-
-          go.*
-          go/**
-          !*.tar.gz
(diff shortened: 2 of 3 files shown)


danielrbradley (commented):
Is the SDK actually already published on the same tag as the root repo which triggers the release? At which point, how do we embed the version numbers into the Go SDK?

I.e. if the Go mod lives in the root, then pushing the tag `v1.2.3` will essentially create a release of the SDK. When this action runs and attempts to push the the Go SDK with the version number on a standalone commit, I think it'll try and push the tag `v1.2.3` to that new commit too, which will fail because it already exists.

blampe:
> I.e. if the Go mod lives in the root, then pushing the tag v1.2.3 will essentially create a release of the SDK. When this action runs and attempts to push the the Go SDK with the version number on a standalone commit, I think it'll try and push the tag v1.2.3 to that new commit too, which will fail because it already exists.

That's a good point... what do you think would be the best way to handle this? Do we not need the publish SDK step?

danielrbradley:
I think the pragmatic solution is to disable the Go publish step for this provider for now. 

This publishing method likely means the SDK is either not pinning the provider version or is pinning the wrong provider version. We should open an issue to investigate. It's possible that the only way to make this work right now is to hard code the version in the repo and bump it ahead of each release.

Long term, we should use the same design of SDK publishing for consistency but that will be a breaking change.

rshade:
@blampe this is going to affect `ingress-nginx` as well now, can I push a commit to this branch to fix those as well?

blampe:
@rshade sorry I missed this, I think Daniel's suggestion to disable the Go publishing step is probably the best path forward for now, but that will probably requiring a little bit of plumbing in ci-mgmt.

rshade:
@blampe no worries, I made a PR instead. I did try to make a go example on nginx and ran into problems:
```
go mod tidy
go: finding module for package github.com/pulumi/pulumi-kubernetes-ingress-nginx/sdk
go: simple-nginx-go imports
        github.com/pulumi/pulumi-kubernetes-ingress-nginx/sdk: module github.com/pulumi/pulumi-kubernetes-ingress-nginx/sdk@latest found (v0.0.10, replaced by ../../sdk), but does not contain package github.com/pulumi/pulumi-kubernetes-ingress-nginx/sdk
```
I am not sure that go.mod is setup right in sdk on either repo

EronWright (approved):

Issue #1118, opened 2024-11-01
Allow mixing generated and manual workflows


danielrbradley:
We currently delete all workflows before generating the new workflows. This will allow us to gradually adopt automation as we migrate things into using automation.

We only do this for the .github folder. We should allow providers to opt-in to cleaning the .github folder.

PR #1120, opened 2024-11-04
Make clean-up part of ci-mgmt program

danielrbradley:
To ensure the behaviour of the clean-up process being consistent irrespective of how provider-ci is executed, move the implementation of the current workflow deletion from the makefile and GH action into the provider-ci program.

This replicates the behaviour of skipping provider-prefixed workflows (see https://github.com/pulumi/ci-mgmt/issues/1060) and also adds a configuration option "clean-all-github-workflows" to disable the mass clean-up altogether and instead rely on the curated tracking list of deleted files in the `getDeletedFiles` method in provider-ci which also fixes https://github.com/pulumi/ci-mgmt/issues/1118


provider-ci/internal/pkg/generate.go
```diff
@@ -43,6 +43,25 @@ func GeneratePackage(opts GenerateOpts) error {
 	if err != nil {
 		return fmt.Errorf("error getting template directories: %w", err)
 	}
+	// Clean up old workflows if requested
+	if clean, found := opts.Config["clean-all-github-workflows"]; found && clean.(bool) {
+		workflows, err := os.ReadDir(filepath.Join(opts.OutDir, ".github", "workflows"))
```
t0yv0, line 48: Does this recursively traverse sub-directories? Or perhaps it does not have to?

The `find` code would recursively traverse.

t0yv0 (approved):
Go code is a better place for this bit, thank you!

provider-ci/internal/pkg/generate.go
t0yv0: I think you're right. We can fix as needed.

provider-ci/internal/pkg/templates/defaults.config.yaml
```diff
@@ -254,4 +254,7 @@ registryDocs: false
 #    # but do say mention (in README.md) that they are under Apache-2.0."
 #    - github.com/alibabacloud-go/endpoint-util/service
 #    - github.com/alibabacloud-go/tea-roa-utils/service
-#    - github.com/alibabacloud-go/tea-roa/client
\ No newline at end of file
+#    - github.com/alibabacloud-go/tea-roa/client
+
+# Delete whole of the .github directory before writing updated templates
+clean-all-github-workflows: true
```
blampe, line 260: Maybe consider something more explicit here like

```yaml
custom-workflows:
- aws-run-upstream-tests.yml
```

That makes it much more obvious what's custom vs. managed, which will come in handy if we ever want to track these down.

guineveresaenger (approved):
Let's explicitly document in `defaults.config.yaml` that "provider-specific" and "custom" workflows are
a) the same thing
b) denoted by the provider name as prefix

provider-ci/internal/pkg/templates/defaults.config.yaml
danielrbradley: I agree the explicit approach would have been nicer when this was originally added, however this would also be a breaking change requiring identifying and fixing all providers that rely on this behaviour (I think it's only AWS, but I'm not sure how to prove it).

My plan is actually to change this to default to false (and remove this code) once I can find a good way to assert that any deleted or renamed generated files are being added correctly to the `getDeletedFiles` method - so we can prove that provider-ci only manages files it's created.

PR #1123, opened 2024-11-04
Document targets design

danielrbradley:
Initial draft diagram to capture the targets, dependencies and and files critical to a provider build.


danielrbradley:
Perhaps worth splitting into 2 diagrams: now & ideal

guineveresaenger (approved):

t0yv0:
We don't like mermaid diagrams? Do not look so good but much easier to edit and GitHub supports just rendering them.

danielrbradley:
> We don't like mermaid diagrams? Do not look so good but much easier to edit and GitHub supports just rendering them. 

Don't dislike them. This was just my starting point to draw something

t0yv0:
By all means, check it in if it helps! Thank you for documenting .

PR #1124, opened 2024-11-04
Add default checkUpstreamUpgrade flag

guineveresaenger:
This flag controls whether we generate the upgrade-provider Workflow. Some providers cannot currently be upgraded automatically, either because they are archived upstream, or because their licensing conflicts with ours.

Followup for some logic that was lost in https://github.com/pulumi/ci-mgmt/pull/1110.


provider-ci/test-providers/cloudflare/.ci-mgmt.yaml
```diff
@@ -25,4 +25,5 @@ actions:
       run: |
         cd provider && go test -v -json -count=1 -cover -timeout 2h -tags=${{ matrix.language }} -parallel 4 . 2>&1 | tee /tmp/gotest.log | gotestfmt
 pulumiConvert: 1
-registryDocs: true
\ No newline at end of file
+registryDocs: true
+checkUpstreamUpgrade: false
```
VenelinMartinov, line 29: good idea adding the configuration to one of the test templates!

VenelinMartinov (approved):
lgtm!

PR #1126, opened 2024-11-05
Typed provider configuration

blampe:
This adds strong typing around our CI configuration options.

Currently our config is just a bag of values, which has led to repos specifying options which have no effect on CI (e.g. `parallel:`, `makeTemplate:` etc.).

This adds types for all of our available options and will fail if `ci-mgmt.yml` specifies an unrecognized option. For backwards compatibility we continue to allow existing no-op params in order to not break existing providers.


provider-ci/test-providers/acme/.ci-mgmt.yaml
```diff
@@ -5,7 +5,6 @@ major-version: 0
 providerDefaultBranch: main
 upstreamProviderOrg: vancluever
 publishRegistry: false
-enableAutoRelease: false
```
blampe, line 8: Only used by acme and not referenced in templates -- safe to remove.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-provider.yml
```diff
@@ -1,4 +1,3 @@
-#{{ if .Config.checkUpstreamUpgrade -}}#
```
blampe, line 1: This was generating an empty workflow which causes lint errors. Changed this to an `if:` below.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-provider.yml
```diff
@@ -22,29 +21,30 @@ env:
 jobs:
   upgrade_provider:
     name: upgrade-provider
-    runs-on: #{{ .Config.runner.default }}#
+    runs-on: #{{ .Config.Runner.Default }}#
+    if: #{{ .Config.CheckUpstreamUpgrade }}#
```
blampe, line 25: Relevant.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-provider.yml
```diff
@@ -70,6 +70,5 @@ jobs:
       - name: Attempt provider upgrade
         # Only attempt the upgrade if we have a target version
         if: steps.target_version.outputs.version != ''
-        run: upgrade-provider "${{ github.repository }}" --kind="all" --target-version="${{ steps.target_version.outputs.version }}" #{{ if .Config.javaGenVersion }}#--java-version="#{{ .Config.javaGenVersion }}#"#{{ end }}#
+        run: upgrade-provider "${{ github.repository }}" --kind="all" --target-version="${{ steps.target_version.outputs.version }}" #{{ if .Config.JavaGenVersion }}#--java-version="#{{ .Config.JavaGenVersion }}#"#{{ end }}#
         shell: bash
-#{{ end }}#
```
danielrbradley, line 75: Had this problem on my PR too - @guineveresaenger showed it can be fixed easy enough by adding a trailing `-`:

```
#{{ end -}}#
```

danielrbradley:
Is there any option to conbine the defaults into the parser too so we don't have to maintain the defaults file separately?

provider-ci/internal/cmd/generate.go
```diff
@@ -21,33 +21,26 @@ var generateCmd = &cobra.Command{
 	Use:   "generate",
 	Short: "Generate repository files.",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		localConfig, err := pkg.LoadLocalConfig(generateArgs.ConfigPath)
-		if err != nil {
-			return err
-		}
-		// Merge local config with template defaults
-		mergedConfig, err := localConfig.WithTemplateDefaults()
```
danielrbradley, line 29: Where's the merging of the defaults gone?

blampe:
> Is there any option to conbine the defaults into the parser too so we don't have to maintain the defaults file separately?

Sure! I kept it around since this is already a pretty big change but that's very doable.

provider-ci/internal/cmd/generate.go
blampe: `LoadLocalConfig` loads the default config and then overlays the local config on top of it.

provider-ci/test-providers/acme/.ci-mgmt.yaml
danielrbradley: Yup, this was added but then refactored away in the same PR it appears with just this left-over.

provider-ci/test-providers/aws/.ci-mgmt.yaml
```diff
@@ -9,7 +9,7 @@ env:
   PULUMI_MISSING_DOCS_ERROR: true
   AWS_REGION: "us-west-2"
   OIDC_ROLE_ARN: ${{ secrets.OIDC_ROLE_ARN }}
-makeTemplate: bridged
+template: bridged-provider
```
danielrbradley, line 12: Does this mean we need to roll out a change to every provider to update this field?

danielrbradley: Or is this a mis-typed field in the config that had no effect?

danielrbradley: Looks like it's set in 74 providers https://github.com/search?q=org%3Apulumi+makeTemplate+path%3A.ci-mgmt.yaml&type=code

guineveresaenger (approved):

provider-ci/test-providers/aws/.ci-mgmt.yaml
blampe: > Or is this a mis-typed field in the config that had no effect?

Yep! `makeTemplate` has no effect, so these all default to using the bridge template anyway. We'll continue to allow them to specify `makeTemplate`, and it'll continue to be a no-op.

blampe:
> Would be nice to not have the documentation duplicated between the types and the defaults at some point, but that could be a follow-up task.

Agree and created https://github.com/pulumi/ci-mgmt/issues/1146

> We'll also need a roll-out plan for fixing any bad fields in config files.

We should get build failure issues for any providers specifying some bad fields not allow-listed here, and I can fix those as needed. Otherwise cleaning these up isn't urgent (arguably not worth doing). IMO it's more important that the boilerplate repos reflect the correct usage and I think you already took care of that.

Issue #55, opened 2020-10-26
Experiment with adding a comment to the command dispatch PRs to suggest a pass / fail status


stack72:
@komalali I am going to tag you here as my person to talk through a UX with as you mentioned that it would be nice to have :D

PR #1128, opened 2024-11-05
Make upgrade workflows more accessible to third parties

ringods:
The`PULUMI_BOT_TOKEN` requirement for third parties is not very intuitive and the upgrade workflows fail without this available.

1. Add the option to use `PULUMI_PROVIDER_AUTOMATION_TOKEN` instead of the `PULUMI_BOT_TOKEN` as a more descriptive alternative.
2. Add fallback if no PAT is available to use the built-in GitHub Actions token. A `permissions` block is therefore also added to allow it to push commits, create Github issues and pull requests in the repository.

Using the built-in token has the downside that the created PR will not be automatically built because it was created by an App ([see the docs](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow) for details).

Internally, we should also switch to using `PULUMI_PROVIDER_AUTOMATION_TOKEN` instead of `PULUMI_BOT_TOKEN` as it's a better name that describes its purpose better within the ever-more crowded list of org-level secrets.

Contributes to: #1087

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-provider.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-provider.yml
@@ -29,7 +29,7 @@ permissions:
   pull-requests: write
 
 env:
-  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  GH_TOKEN: ${{ secrets.PULUMI_PROVIDER_AUTOMATION_TOKEN || secrets.PULUMI_BOT_TOKEN || secrets.GITHUB_TOKEN }}
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-bridge.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-bridge.yml
@@ -56,7 +56,7 @@ permissions:
   pull-requests: write
 
 env:
-  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  GH_TOKEN: ${{ secrets.PULUMI_PROVIDER_AUTOMATION_TOKEN || secrets.PULUMI_BOT_TOKEN || secrets.GITHUB_TOKEN }}
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         
 jobs:
--- a/provider-ci/test-providers/acme/.github/workflows/upgrade-bridge.yml
+++ b/provider-ci/test-providers/acme/.github/workflows/upgrade-bridge.yml
@@ -56,7 +56,7 @@ permissions:
   pull-requests: write
 
 env:
-  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  GH_TOKEN: ${{ secrets.PULUMI_PROVIDER_AUTOMATION_TOKEN || secrets.PULUMI_BOT_TOKEN || secrets.GITHUB_TOKEN }}
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         
 jobs:
--- a/provider-ci/test-providers/acme/.github/workflows/upgrade-provider.yml
+++ b/provider-ci/test-providers/acme/.github/workflows/upgrade-provider.yml
@@ -28,7 +28,7 @@ permissions:
   pull-requests: write
 
 env:
-  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  GH_TOKEN: ${{ secrets.PULUMI_PROVIDER_AUTOMATION_TOKEN || secrets.PULUMI_BOT_TOKEN || secrets.GITHUB_TOKEN }}
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
--- a/provider-ci/test-providers/aws/.github/workflows/upgrade-bridge.yml
+++ b/provider-ci/test-providers/aws/.github/workflows/upgrade-bridge.yml
@@ -56,7 +56,7 @@ permissions:
   pull-requests: write
 
 env:
-  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  GH_TOKEN: ${{ secrets.PULUMI_PROVIDER_AUTOMATION_TOKEN || secrets.PULUMI_BOT_TOKEN || secrets.GITHUB_TOKEN }}
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         
 jobs:
--- a/provider-ci/test-providers/aws/.github/workflows/upgrade-provider.yml
+++ b/provider-ci/test-providers/aws/.github/workflows/upgrade-provider.yml
@@ -28,7 +28,7 @@ permissions:
   pull-requests: write
 
 env:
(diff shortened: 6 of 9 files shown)


danielrbradley (commented):
Looks good at first glance. Will take this for a test drive before merging.

iwahbe:
I remember taking a crack at this earlier, and hitting permission issues with publishing. If you can get it to work, then by all means merge it.

ringods:
@iwahbe the upgrade workflows aren't publishing anything. The publishing in the release workflow is already working with `GITHUB_TOKEN` and elevated permissions, for instance:

https://github.com/pulumi/pulumi-cloudflare/blob/803fd6dd7d823d98cdf3e318739ed14080641def/.github/workflows/release.yml#L61-L65

danielrbradley:
I've set up the xyz provider to work for testing upgrade provider.

I've temporarily merged these changes into main to be able to test the upgrade bridge and provider workflows:
- Upgrade bridge: [job](https://github.com/pulumi/pulumi-xyz/actions/runs/11743762682), [PR](https://github.com/pulumi/pulumi-xyz/pull/522)

Edit: The changes for the provider upgrade were lost because the xyz provider didn't have the workflow enabled when this PR was opened.

danielrbradley:
From the [docs](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow):

> When you use the repository's GITHUB_TOKEN to perform tasks, events triggered by the GITHUB_TOKEN, with the exception of workflow_dispatch and repository_dispatch, will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs. For example, if a workflow run pushes code using the repository's GITHUB_TOKEN, a new workflow will not run even when the repository contains a workflow configured to run when push events occur.

The recommendation on this docs page is to use a PAT when pushing PRs which should trigger a workflow. Perhaps as an in-between we should use the BOT token when creating the pull request, if it's available, but fall back to the built-in token for third parties who haven't set this up yet.

Alternatively, we might be able to use a `workflow_dispatch` or `repository_dispatch` to run the acceptance tests directly, and allow the manual sentinel check to be written back to the correct PR, but this will require a little more engineering. One downside of this approach is that it won't give direct access to the job details on the PR if specific parts fail.

danielrbradley:
The `GITHUB_TOKEN` does not have scopes to create `repository_dispatch` or `workflow_dispatch` events, so we're back to requiring a PAT at some point here. 

I think we might just have to gate the step for the creation of the PR on the PAT being set up. We could allow a less pulumi-bot specific secret name to be used instead. If third parties don't want to configure the PAT, then their only other option is to periodically run the upgrade job locally.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-provider.yml
```diff
@@ -16,9 +16,15 @@ on:
...
     - cron: 0 3 * * *
 
+permissions:
+  contents: write
+  issues: write
+  pull-requests: write
+
 env:
-  GH_TOKEN: ${{ secrets.PULUMI_BOT_TOKEN }}
+  GH_TOKEN: ${{ secrets.UPGRADE_PROVIDER_TOKEN }}
```
danielrbradley, line 25: I think for each of these we need fallbacks, something like the following ... 

```suggestion
  GH_TOKEN: ${{ secrets.UPGRADE_PROVIDER_TOKEN || secrets.PULUMI_BOT_TOKEN || secrets.GITHUB_TOKEN }}
```

ringods: @danielrbradley does such a logical OR work within the interpolation?

danielrbradley: From the point of view of the org-level secret, I'm not quite convinced this is the right name yet.

Perhaps something more like `PULUMI_PROVIDER_AUTOMATED_PR_TOKEN` .. though shorter would be better.

danielrbradley: I've not tested the interpolation with ORs, but I _think_ it should work.

danielrbradley: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#example

```
env:
  MY_ENV_VAR: ${{ github.ref == 'refs/heads/main' && 'value_for_main_branch' || 'value_for_other_branches' }}
```

ringods: Suggestions:
- `PULUMI_PROVIDER_AUTO_PR_TOKEN`
- `PULUMI_PROVIDER_AUTOMATION_TOKEN`

ringods:
@danielrbradley @t0yv0 @iwahbe linking back to Daniel's comment here: https://github.com/pulumi/ci-mgmt/pull/1128#discussion_r1840184826

I would like to settle on the name of the PAT token so I can test this upfront using a Pulumiverse provider. Since I don't have admin access anymore on Pulumiverse, I have to ask approval to the Pulumiverse GH org admins when using a fine-grained PAT.

Here are my suggestions:

* `PULUMI_PROVIDER_AUTO_PR_TOKEN`
* `PULUMI_PROVIDER_AUTOMATION_TOKEN`

I personally don't mind a longer name if it explains better what it is for.

danielrbradley:
`PULUMI_PROVIDER_AUTOMATION_TOKEN` is pretty nice

t0yv0:
Folks I'm summoned to review but I don't have a good handle on what's going on here. NO objection from me if it keeps working for the Pulumi providers builds 🙏

ringods:
Verification of a succesful run:

* A PAT linked to my GH account is configured as `PULUMI_PROVIDER_AUTOMATION_TOKEN`
* The `Upgrade Provider` workflow [ran succesfully](https://github.com/pulumiverse/pulumi-cockroach/actions/runs/11976515719/job/33392362923) for the Pulumiverse Cockroach provider.
* It resulted in [PR #55](https://github.com/pulumiverse/pulumi-cockroach/pull/55)
* By using a PAT, also the `Run Acceptance Tests` workflow was [correctly started](https://github.com/pulumiverse/pulumi-cockroach/actions/runs/11976577572).

danielrbradley:
Testing with no new version to upgrade to: https://github.com/pulumi/pulumi-xyz/actions/runs/11975610349/job/33389373072

Currently, something is not working as it's never detecting the new upstream version.

iwahbe (commented):
I havn't been following very well. @ringods Are you still developing here or is this ready for review?

ringods:
@iwahbe this is ready for review. This works for me for my Pulumiverse providers, but it should still be tested on a Pulumi-managed provider.

danielrbradley:
Next test, I've opened a PR in pulumi-tf-bridge-boilerplate: https://github.com/pulumi/pulumi-tf-provider-boilerplate/pull/148

I've then started the upgrade-provider job manually from that temporary branch: https://github.com/pulumi/pulumi-tf-provider-boilerplate/actions/runs/12028327439/job/33531214829

danielrbradley (approved):
After testing via the TF bridge boilerplate this looks to be functional 👍

iwahbe (approved):
If it works (and we've tested it), it looks good to me.

PR #1129, opened 2024-11-05
Don't fail the upgrade check job if we can't open a PR

danielrbradley:
Docs for continue-on-error:
> Prevents a job from failing when a step fails. Set to true to allow a job to pass when this step fails.

Failing the job results in a P1 issue being opened. We've already opened an issue to track the pending upgrade, so we'll just exit cleanly if the PR can't be made.

This will still raise a P1 issue if the upstream check (to create the tracking issue) fails.


VenelinMartinov (changes requested):
Can we please clarify what the workflow for people who consume this is? We use failed jobs to see which provider upgrades have failed.

danielrbradley:
> Can we please clarify what the workflow for people who consume this is? We use failed jobs to see which provider upgrades have failed.

Could you expand on what you're wanting changed here?

provider-ci/test-providers/cloudflare/.github/workflows/upgrade-provider.yml
guineveresaenger, line 1: This file should not be here, we might need to revisit empty templates?

iwahbe (approved):
The change makes sense to me. I think you need to clarify (in the comments, please) how this should be used in the wild.

VenelinMartinov:
Yeah, currently the workflow for provider upgrades is:

For each open provider upgrade issue:
1. Check if it has a corresponding PR.
2. If it does, try to merge it.
3. If it doesn't, see if we had a failure in the upgrade workflow.
4. If the job failed run the upgrade manually.

I am concerned that feedback we used to have of the workflow failing is now lost and we'd have to dig in github logs to identify the failed upgrades.

I'd love for us not to make this already manual and laborious process even harder by hiding the fact that the workflow has failed. This has to be applied across 60 bridged providers, all of which have slightly different behaviours and annoyances and it has to happen every week.


Can we consider something here to make sure the process doesn't get worse? Perhaps the workflow can comment on the original PR about a failure with a link to the failed workflow?

VenelinMartinov (approved):
Thanks for fixing 🚀

guineveresaenger (approved):

iwahbe:
> Yeah, currently the workflow for provider upgrades is:
> 
> For each open provider upgrade issue:
> 
> 1. Check if it has a corresponding PR.
> 2. If it does, try to merge it.
> 3. If it doesn't, see if we had a failure in the upgrade workflow.
> 4. If the job failed run the upgrade manually.
> 
> I am concerned that the feedback we used to have of the workflow failing is now lost and we'd have to dig in github logs to identify the failed upgrades.
> 
> I'd love for us not to make this already manual and laborious process even harder by hiding the fact that the workflow has failed. This has to be applied across 60 bridged providers, all of which have slightly different behaviours and annoyances and it has to happen every week.
> 
> Can we consider something here to make sure the process doesn't get worse? Perhaps the workflow can comment on the original PR about a failure with a link to the failed workflow? If not, can we at least clarify what the expected workflow is now, and make sure everyone is aware of it?

@VenelinMartinov Just to clarify, this change is maintaining the "current process". After this change, we will still get a P1 if `upgrade-provider` fails to stand up an issue. Due to this change, we won't get a P1 if `upgrade-provider` succeeds in standing up an issue but fails in raising an associated PR (since we were already alerted by the issue). We didn't have feedback that the workflow failed before 5 days ago (when we merged https://github.com/pulumi/ci-mgmt/pull/1110). The workflow for provider upgrades remains unchanged. This PR gets us back to the same place as we were 6 days ago.

PR #1139, opened 2024-11-11
Tweak warning to make more sense to 3rd parties

danielrbradley:
As suggested during work on the TF boilerplate: https://github.com/pulumi/pulumi-tf-provider-boilerplate/pull/135#discussion_r1832763213


provider-ci/internal/pkg/templates/provider/.github/workflows/lint.yml
```diff
@@ -14,29 +14,29 @@ jobs:
...
-        git grep -l 'go:embed' -- provider | xargs sed -i 's/go:embed/ goembed/g'
-    - name: prepare upstream
-      continue-on-error: true
-      run: make upstream
-    - name: golangci-lint
-      uses: golangci/golangci-lint-action@v6
-      with:
-        version: v1.60
-        working-directory: provider
+      - name: Checkout Repo
```
guineveresaenger, line 17: These feel like spurious changes caused by editor opinion.

danielrbradley: Bah, yes

danielrbradley: Ah - only reset the changes in bridged-provider but missed the other sub-templates

guineveresaenger (changes requested):
I'd rather not flip back and forth with regard to indentation in the templates. We should standardize/enforce via lint the indentation we want. I can be convinced that the changes proposed are "more" correct but it feels like adding unrelated noise to this pull request.

provider-ci/internal/pkg/templates/provider/.github/workflows/lint.yml
danielrbradley: Fixed

danielrbradley:
> I'd rather not flip back and forth with regard to indentation in the templates. We should standardize/enforce via lint the indentation we want. I can be convinced that the changes proposed are "more" correct but it feels like adding unrelated noise to this pull request.

This is just an artifact of how our templates are currently oddly structured. Running find & replace in the editor causes an auto-format which then 'fixes' all of these indents which I then have to manually reset 😫

guineveresaenger (approved):
Thank you for cleaning this up!

Issue #1131, opened 2024-11-07
Allow `make` target customization


blampe:
Currently, we either ship a fully generated `Makefile` to providers (bridged) or we leave it entirely up to the provider to implement their own `Makefile` (native).

There are a number of problems with this:
* Manually maintained `Makefiles` can break or become stale as the underlying workflows change or evolve. For example as we look to roll out more generalized test sharding (https://github.com/pulumi/ci-mgmt/issues/676) we would need to update all native `Makefiles` to take advantage.
* Our generated `Makefiles` don't expose any entry points for customization. As a result, we've baked a number options into `ci-mgmt.yml` which in many cases only apply to one or two providers. This isn't scalable -- it's difficult to maintain and it creates a tight coupling between providers and ci-mgmt.

Instead, ci-mgmt should make some default `make` targets available to each provider while also giving providers the ability to override those targets should they need to. For example, `pulumi-eks` should be able to use standard `build_go`, `build_python`, `build_dotnet` targets but also be able to define its own `build_nodejs` since it's a component provider with unique node requirements.

I've prototyped one approach which seems viable:
* `ci-mgmt` manages a `.mk` directory for the provider. This directory includes `defaults.mk` for default targets, and `vars.mk` for default variable like `PACK=eks`.
* Default targets take the form of something like `default.build_nodejs:`. The `defaults.mk` file includes a catch-all rule like so:
    ```make
    include .mk/vars.mk

    default.build_nodejs:
       ...

    # Delegate any undefined target "foo" to its respective "default.foo" target,
    # if it exists.
    %:
        @$(MAKE) -f defaults.mk "default.$@"
    ```
* A provider's `Makefile` can then contain only the following to automatically pick up any default variables and targets. This should be completely sufficient for the boilerplate to participate in CI workflows:
    ```make
    include .mk/defaults.mk
    ```
* A provider that wished to define custom behavior for itself can then do so as needed:
    ```make
    include .mk/defaults.mk
    
    lint: # Override default lint target.
        ...
    ```

t0yv0:
We use a simple form of this in pulumi-aws. 

https://github.com/pulumi/pulumi-aws/blob/master/.mk/minimal_schema.mk

It's very finnicky though to get the customized targets run at the right time either "before or after" the normal targets. 

Perhaps if the default makefile had before_build_nodejs and after_build_nodejs etc targets then these could be used to add behavior in .mk overrides at the precise phase.

mjeffryes:
[...]

Issue #1133, opened 2024-11-07
provider/license template assumes a `make upstream` target


blampe:
https://github.com/pulumi/ci-mgmt/blob/fbc688d36a0d5f4b15970560c7e4e8f780f49c47/provider-ci/internal/pkg/templates/provider/.github/workflows/license.yml#L25

`upstream` is specific to bridged providers.

We should move this action to the bridged template, or fix it to not require building upstream.

PR #1140, opened 2024-11-12
[WIP] Generic provider template

blampe:
This is a _very_ WIP fork of our bridged workflows which is intended to be used by native, component, or other miscellaneous providers not currently managed by ci-mgmt.

There are two primary design principles which deviate from the bridged template:
1. `make` targets define the entire contract between a provider and `ci-mgmt`. We do not assume anything about the provider, e.g. it could be implemented in a language other than Go.
2. The provider is responsible for defining its own overrides and customizations, and `ci-mgmt` is responsible for exposing the appropriate hooks (https://github.com/pulumi/ci-mgmt/issues/1131).

For example, if the provider needs some additional tooling then it is responsible for ensuring the tool is installed. Historically these sorts of customizations have been handled as one-offs in `ci-mgmt`, but we want to push this down to the `make` target or into the test itself. Not only does this make `ci-mgmt` easier to maintain and refactor, but it also forces us to write tests that are more easily reproducible in local environments. The SSH key handling between the [docker](https://github.com/pulumi/pulumi-docker/blob/f1e7e25d9404ca908d4a57a5fc13258673d9a4b3/.github/workflows/run-acceptance-tests.yml#L205-L208) and [docker-build](https://github.com/pulumi/pulumi-docker-build/blob/4c030122685eaebbef6ac58494f1443f75640cd5/examples/main_test.go#L16-L21) providers is illustrative.

I've been testing this in EKS and haven't gotten tests passing reliably yet. I was planning on consolidating the test into a shared action once I got that far, but also feel free to do that sooner than later!

EKS was already sharding tests so I've opted to keep that behavior here -- a provider just needs to implement a `test_shard` target which takes a `TESTS` and `TAG` env var. I've tweaked it somewhat so the sharding is totally arbitrary, which in practice just means we need to install all the SDKs instead of one at a time.

Other differences:
* We don't install gotestfmt. (Ref https://github.com/pulumi/ci-mgmt/pull/1072) 
* We don't have setupScript, extraTests, preTest, etc.

TODO:
- [x] Make e2e test path configurable (currently hard-coded to examples).
- [x] Make shard determination configurable and local (currently a workflow step).
- [x] Consolidate test action.


provider-ci/internal/pkg/templates/generic/.github/workflows/run-acceptance-tests.yml
```diff
@@ -0,0 +1,221 @@
...
+        # Write an explicit status check called "Sentinel" which will only pass if this code really runs.
+        # This should always be a required check for PRs.
+        context: 'Sentinel'
+        description: 'All required checks passed'
+        state: 'success'
+        # Write to the PR commit SHA if it's available as we don't want the merge commit sha,
+        # otherwise use the current SHA for any other type of build.
+        sha: ${{ github.event.pull_request.head.sha || github.sha }}
+
+  # TODO: Extract into shared action.
```
danielrbradley, line 105: 👍

provider-ci/internal/pkg/templates/generic/.github/workflows/license.yml
```diff
@@ -0,0 +1,48 @@
+# WARNING: This file is autogenerated - changes will be overwritten if not made via https://github.com/pulumi/ci-mgmt
+
+name: license_check
```
danielrbradley, line 3: I think this should already be included from the `provider` template .. unless there's something subtle we're tweaking here I didn't spot?

provider-ci/internal/pkg/generate.go
```diff
@@ -93,6 +93,8 @@ func getTemplateDirs(templateName string) ([]string, error) {
 	case "external-bridged-provider":
 		// Render more specific templates last to allow them to override more general templates.
 		return []string{"dev-container", "provider", "bridged-provider"}, nil
+	case "generic":
+		return []string{"provider", "pulumi-provider", "generic"}, nil
```
danielrbradley, line 97: When we use this template for the native boilerplate, we'll probably add `external-generic` which will exclude the `pulumi-provider` sub-template.

danielrbradley (commented):
Direction & progress looks good. It would probably be worth writing down a design for the required makefile targets and artifacts while doing this work - perhaps in a README somewhere in this repo - which can serve as a design doc too.

provider-ci/internal/pkg/generate.go
blampe: Or maybe "third-party"?

provider-ci/internal/pkg/templates/generic/.github/workflows/license.yml
blampe: Ah yeah, I duplicated it here because of https://github.com/pulumi/ci-mgmt/issues/1133

provider-ci/internal/pkg/templates/generic/.github/workflows/build_sdk.yml
```diff
@@ -0,0 +1,70 @@
...
+            .pulumi/examples-cache
+          key: ${{ runner.os }}-${{ hashFiles('provider/go.sum') }}
+      - name: Setup tools
+        uses: ./.github/actions/setup-tools
+        with:
+          tools: pulumictl, pulumicli, ${{ matrix.language }}
+      - name: Download bin
+        uses: ./.github/actions/download-bin
+      - name: Install plugins
+        run: make install_plugins
```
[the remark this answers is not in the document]
blampe: Yeah, I was also wondering about this because the CLI should automatically download these plugins anyway. Related - these plugins eat a ton of space! https://github.com/pulumi/pulumi/issues/17718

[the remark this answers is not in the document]
danielrbradley: I think for bridged providers, if they're opted into using the converter plugin, then it explicitly disables auto-download because we need to have the provider version pinned, so failing fast is good. For generic providers, perhaps we can just let the provider figure out what it needs and how it will install them internally within the makefile.

provider-ci/internal/pkg/generate.go
danielrbradley: Just above is the `external-bridged-provider` template which is like `bridged-provider` but with bits removed for third parties. Whatever name we use should aim to keep these consistent.

provider-ci/internal/pkg/templates/README.md
```diff
@@ -0,0 +1,137 @@
...
+languages, but this is restrictive and a poor developer experience in general
+(https://github.com/pulumi/ci-mgmt/issues/676).
+
+* `make install_sdks`: Install SDKs for all available languages.
+* `make shard`: This target takes two environment variables -- `$SHARD_TOTAL`
+  and `$SHARD_INDEX` -- and is responsible for determining tests to run for
+  this shard. It will probably mutate the environment in some way, for example
+  by appending to `$GITHUB_ENV`, in order to inform the `test_shard` target.
+* `make test_shard`: This target is responsible for executing tests identified
+  in the `shard` target.
```
danielrbradley, line 130: I'm a little confused why we have two targets here. It feels like we shouldn't need `make shard` at all .. and we don't want to have to tell the provider author to know how to interact with github env.

Perhaps we should be just using `make test` or `make integration_test` with variables for `SHARD_TOTAL` and `SHARD_INDEX` e.g. `make integration_test SHARD_TOTAL=5 SHARD_INDEX=3`.

It would also be nice to link from here to an example of how to do sharding with `go test`.

danielrbradley (commented):
Love the design in the README. That's super helpful 💜 

Sharding is interesting, but doesn't feel it's quite there yet.

Relying on phony targets vs files is worth a quick think, but the phony targets work fine if in doubt.

danielrbradley:
Done a bit more experimenting around the test sharding and am pretty sure we can do this with a single target with optional make variables.

Here's how we can adapt our current bridge makefile to support sharding (or runing specific tests locally):

```
TEST_RUN_ARG =
ifneq "$(origin TEST_RUN)" "undefined"
  TEST_RUN_ARG = -run "${TEST_RUN}"
else ifneq ("$(origin TEST_SHARD_TOTAL)|$(origin TEST_SHARD_INDEX)", "undefined|undefined")
	TEST_RUN_ARG = -run "$$(cd examples && go run github.com/blampe/shard@9d1f3b21786e18caa1989e19502595143985d61b --total "$(TEST_SHARD_TOTAL)" --index "$(TEST_SHARD_INDEX)" --output env)"
endif

test: export PATH := $(WORKING_DIR)/bin:$(PATH)
test:
	cd examples && go test -v -tags=all -parallel $(TESTPARALLELISM) -timeout 2h $(TEST_RUN_ARG)
.PHONY: test

TEST_PROVIDER_RUN_ARG =
ifneq "$(origin TEST_RUN)" "undefined"
  TEST_PROVIDER_RUN_ARG = -run "${TEST_RUN}"
else ifneq ("$(origin TEST_SHARD_TOTAL)|$(origin TEST_SHARD_INDEX)", "undefined|undefined")
	TEST_PROVIDER_RUN_ARG = -run "$$(cd provider && go run github.com/blampe/shard@9d1f3b21786e18caa1989e19502595143985d61b --total "$(TEST_SHARD_TOTAL)" --index "$(TEST_SHARD_INDEX)" --output env)"
endif
```

Using `-n` to just print commands to demonstrate usage:

```console
make -n test TEST_RUN="Things"
cd examples && go test -v -tags=all -parallel 10 -timeout 2h -run "Things"

make -n test_provider TEST_RUN="Things"
cd provider && go test -v -short ./... -parallel 10 -run "Things"

make -n test TEST_SHARD_INDEX=2 TEST_SHARD_TOTAL=5
cd examples && go test -v -tags=all -parallel 10 -timeout 2h -run "$(cd examples && go run github.com/blampe/shard@9d1f3b21786e18caa1989e19502595143985d61b --total "5" --index "2" --output env)"

make -n test_provider TEST_SHARD_INDEX=2 TEST_SHARD_TOTAL=5
cd provider && go test -v -short ./... -parallel 10 -run "$(cd provider && go run github.com/blampe/shard@9d1f3b21786e18caa1989e19502595143985d61b --total "5" --index "2" --output env)"
```

danielrbradley:
Next thoughts on the sharding ... I think we should roll it out as a smallish change to bridged providers first to solidify the design.

1. In order to get this live, we should create a version of the shard program in the pulumi organisation.
2. How are we planning to upgrade the version of the shard tool that we depend on? Can we hook this into Renovate?
3. In the implementation of the shard tool - is there any way to leverage Go's own test discovery process so we're sure the tool behaves consistently with running go test?
4. Related to (3): I think there's potentially an issue here around when there's modules within modules (e.g. with the shim or upstream directories).

mjeffryes:
We've roughly landed all the remaining providers with some hacks on the bridged template so I don't think we need this PR anymore

PR #1144, opened 2024-11-14
Remove defaultable provider-ci arguments

danielrbradley:
I noticed while testing https://github.com/pulumi/ci-mgmt/pull/1143 that [the generated PR](https://github.com/pulumi/pulumi-tf-provider-boilerplate/pull/138/files) ignored the repo's configured template and changed it.

- Use values from the config instead of hard coding them
- `--name` defaults to `{config.repository}` or `{config.organization}/pulumi-{config.provider}`
- `--template` defaults to `{config.template}` or otherwise "bridged-provider"
- `--config` defaults to `.ci-mgmt.yaml` so only specify in the update-workflows job where it's checked out into a different directory
- `--out` defaults to `.` so only specify in the update-workflows job where it's checked out into a different directory

### DIFF
--- a/.github/workflows/update-workflows.yml
+++ b/.github/workflows/update-workflows.yml
@@ -78,8 +78,6 @@ jobs:
         if: inputs.bridged
         run: |
           cd ci-mgmt/provider-ci && go run ./... generate \
-          	--name pulumi/pulumi-${{ inputs.provider_name }} \
-                --template bridged-provider \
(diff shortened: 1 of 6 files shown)


VenelinMartinov (approved):

flostadler (approved):
lgtm

iwahbe (approved):
We should probably document somewhere that the interface to `github.com/pulumi/ci-mgmt/provider-ci` is considered internal, and may change freely between commits.

PR #1150, opened 2024-11-21
Move migration support to provider-ci

danielrbradley:
Move away from the typescript scripts which are only run during push but not from the `make ci-mgmt` target. This ensures migrations are run in both the pull and push styles consistently.

- Migrate existing migrations.
- Remove old typescript code.
- Remove old step in CI.
- Skip running migrations for local test-providers as they have no code in them and will fail most of the time.


danielrbradley:
Any suggestions on why embed works perfectly for me locally but seems to be failing in CI most welcome!

iwahbe (approved):
This is better.

PR #1151, opened 2024-11-21
Makefile performance improvements

danielrbradley:
Allow targets to be up-to-date by using the "sentinel file" pattern where each unit of work is tracked via a file in the `.make` directory to indicate when it was last successfully completed. Replace dependencies on phony targets with these sentinel files.

Other improvements:
- Add `make schema` target as a generic replacement for `make tfgen`.
- Split out a generate phase from all the build sdk targets to allow skipping running the SDK builds locally.
- Add `make generate` target for generating all things which need to be committed without having to wait on build steps.
- Add `make prepare_local_workspace` which currently encompases installing plugins and preparing upstream.
- Group `.PHONY` declarations near the definition of the phony target rather than in one very long line. This allows us to see if we're missing declarations for specific targets more easily.
- Fix the `.pulumi/version` getting out of sync with the installed files.
- Fix the `clean` target to include the `bin` and `.make` directories (remove the duplicate `cleanup` target alongside).
- Replace the broken `help` target which hasn't been functional in many years AFAIK.
- Move nuget setup for install_dotnet_sdk into the makefile from CI.
- Use the new standard targets (schema & provider) instead of the internal targets (tfgen_build_only, tfgen_no_deps, provider_no_deps)
- Save and restore makefile progress via artifacts (prerequisites.make and build_[language].make). This maintains file timestamps between jobs.
 
Commits are structued to address one target at a time for review and history.

Preview of new help text:

```
Usage: make [target]

Main Targets
  build (default)     Build the provider and all SDKs and install for testing
  generate            Generate all SDKs, documentation and schema
  provider            Build the local provider binary
  lint_provider<.fix> Run the linter on the provider (& optionally fix)
  test_provider       Run the provider tests
  test                Run the example tests (must run 'build' first)
  clean               Clean up generated files

More Precise Targets
  schema        Generate the schema
  generate_sdks Generate all SDKs
  build_sdks    Build all SDKs
  install_sdks  Install all SDKs
  provider_dist Build and package the provider for all platforms

Tool Targets
  ci-mgmt     Re-generate CI configuration from .ci-mgmt.yaml
  debug_tfgen Start a debug server for tfgen

Internal Targets (automatically run as dependencies of other targets)
  prepare_local_workspace  Prepare for building
  install_plugins          Install plugin dependencies
  upstream                 Initialize the upstream submodule, if present

Language-Specific Targets
  generate_[language]    Generate the SDK files ready for committing
  build_[language]       Build the SDK to check correctness
[...]

PR #1152, opened 2024-11-22
Run build_registry_docs as part of SDK build

guineveresaenger:
https://github.com/pulumi/pulumi-snowflake/issues/752 revealed that we weren't running registry docs build on pull requests.
This addresses that issue. Index docs will now be built on pull requests in CI.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/prerequisites.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/prerequisites.yml
@@ -64,6 +64,10 @@ jobs:
 #{{- end }}#
     - name: Generate schema
       run: make schema
+    #{{- if .Config.RegistryDocs }}#
+    - name: Build registry docs
+      run: make build_registry_docs
+    #{{- end }}#
     - name: Build provider binary
       run: make provider
     - name: Unit-test provider code
--- a/provider-ci/test-providers/cloudflare/.github/workflows/prerequisites.yml
+++ b/provider-ci/test-providers/cloudflare/.github/workflows/prerequisites.yml
@@ -66,6 +66,8 @@ jobs:
       run: make prepare_local_workspace
     - name: Generate schema
       run: make schema
+    - name: Build registry docs
+      run: make build_registry_docs
     - name: Build provider binary
       run: make provider
     - name: Unit-test provider code
--- a/provider-ci/test-providers/docker/.github/workflows/prerequisites.yml
+++ b/provider-ci/test-providers/docker/.github/workflows/prerequisites.yml
@@ -79,6 +79,8 @@ jobs:
       run: make prepare_local_workspace
     - name: Generate schema
       run: make schema
+    - name: Build registry docs
+      run: make build_registry_docs
     - name: Build provider binary
       run: make provider
     - name: Unit-test provider code


danielrbradley (commented):
I'm not sure this is the right place to put this command.

1. It looks like we're running this on every parallel job in the build_sdk matrix, but it's doing exactly the same work in each branch.
2. It doesn't look like we're doing anything with the result of the build - unless it's uploaded as part of each language's SDK, but my understanding from the `build_registry_docs` makefile target is it's just outputting the `docs/` directory.
3. If we're just trying to assert that the files are up to date, I think this needs to go _before_ the "Check worktree clean" step.

Alternative options:
1. Could we build this as part of the schema generation step? It's not exactly clear from the makefile what input files this target depends on running first.
2. Perhaps we can model this dependency in the makefile rather than in CI so that it also runs automatically for anyone working on the provider locally?

Related to this .. I'm currently working on a larger piece of work to connect all the makefile targets together - so all dependencies are modelled even for local builds:
- https://github.com/pulumi/ci-mgmt/pull/1151

If it's a makefile change, I can make it as part of that PR instead.

iwahbe:
I would implement this via the Makefile:

```patch
-build_sdks: build_nodejs build_python build_dotnet build_go build_java build_registry_docs
+-build_sdks: build_nodejs build_python build_dotnet build_go build_java
```
```patch
-tfgen: install_plugins upstream tfgen_no_deps
+tfgen: install_plugins upstream tfgen_no_deps build_registry_docs\
```

It feels like generating docs is more in the `make tfgen` category (with the schema) then as an SDK artifact.

guineveresaenger:
Unfortunately @iwahbe's suggestion won't work - `make registry_docs` depends on `make tfgen` (having the provider binary, with edit rules).

> It doesn't look like we're doing anything with the result of the build 
That's right, it's a plain "is this still valid" check

> If we're just trying to assert that the files are up to date
Not quite - my approach was perhaps a bit simplistic - I want to run this as a separate check that would fail if there were docs changes incompatible with docs replaces that have a hard fail baked into them. 

It sounds like two things should happen here:

1. Perhaps run this as part of prerequisites so we don't run it four times 😅 
2. show the dependency on `make tfgen` in the Makefile.

iwahbe:
> Unfortunately @iwahbe's suggestion won't work - `make registry_docs` depends on `make tfgen` (having the provider binary, with edit rules).

This works for me (tested in random from a cleaned repo):

```patch
 Makefile | 6 +++---
 1 file changed, 3 insertions(+), 3 deletions(-)

diff --git a/Makefile b/Makefile
index 455c4723..225d1d7a 100644
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ development: install_plugins provider build_sdks install_sdks
 
 build: install_plugins provider build_sdks install_sdks
 
-build_sdks: build_nodejs build_python build_dotnet build_go build_java build_registry_docs
+build_sdks: build_nodejs build_python build_dotnet build_go build_java
 
 install_go_sdk:
 
@@ -99,7 +99,7 @@ build_python: upstream
 		../venv/bin/python -m build .
 
 # Run the bridge's registry-docs command to generated the content of the installation docs/ folder at provider repo root
-build_registry_docs:
+build_registry_docs: upstream tfgen_build_only
 	$(WORKING_DIR)/bin/$(TFGEN) registry-docs --out $(WORKING_DIR)/docs
 
 clean:
@@ -155,7 +155,7 @@ test_provider:
 	@echo ""
 	cd provider && go test -v -short ./... -parallel $(TESTPARALLELISM)
 
-tfgen: install_plugins upstream tfgen_no_deps
+tfgen: install_plugins upstream tfgen_no_deps build_registry_docs
 
 tfgen_no_deps: export PULUMI_HOME := $(WORKING_DIR)/.pulumi
 tfgen_no_deps: export PATH := $(WORKING_DIR)/.pulumi/bin:$(PATH)
```

It incurs minimal overhead of `make tfgen` (only the actual docs generation is a new target) and builds correctly from a clean repo.

guineveresaenger:
I'd been holding off on this due to https://github.com/pulumi/ci-mgmt/pull/1151 - with those changes, we have `build_registry_docs` as a separate Make target dependent on `make schema(tfgen)` and we can run it as-is in the prerequisites workflow.

danielrbradley (approved):

PR #1157, opened 2024-11-26
Fix asset rebuilds & dependencies

danielrbradley:
Follow-up to #1151

## Fix tfgen makefile dependencies

- Rebuild the binary if any go source files change (use a wildcard here rather than limiting to resource.go).
- Rebuild the binary if upstream changes in any way.

## Use make --touch for restoring progress 

When we restore the make progress as an asset it had the timestamps of when the files were created in the original job, as expected. However, the source files had been freshly checked out. Therefore, any file that has already been built in a previous job and restored in a subsequent job was older than the freshly checked out files and was rebuilt, causing longer runtimes.

- Any job that depends on the `prerequisites` job (and restores its artifacts) should touch the `provider` and `schema` targets.
- Any job that depends on the `build_sdk` job matrix should touch the `provider`, `schema` and `build_[language]` targets.

### DIFF
--- a/provider-ci/test-providers/acme/.github/workflows/build_provider.yml
+++ b/provider-ci/test-providers/acme/.github/workflows/build_provider.yml
@@ -48,10 +48,7 @@ jobs:
           merge-multiple: true
           path: provider/cmd/pulumi-resource-acme/schema-embed.json
       - name: Restore makefile progress
-        uses: actions/download-artifact@v4
-        with:
-          name: prerequisites.make
-          path: .make
+        run: make --touch provider schema
       - name: Build & package provider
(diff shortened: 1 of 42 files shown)


provider-ci/test-providers/acme/.github/workflows/build_provider.yml
```diff
@@ -48,10 +48,7 @@ jobs:
           merge-multiple: true
           path: provider/cmd/pulumi-resource-acme/schema-embed.json
       - name: Restore makefile progress
-        uses: actions/download-artifact@v4
-        with:
-          name: prerequisites.make
-          path: .make
```
blampe, line 54: Fewer dependent GitHub actions is always a good thing!

blampe (approved):
Very clean.

Issue #1156, opened 2024-11-26
Manage action versions with Renovate


blampe:
ci-mgmt is responsible for managing workflows for providers, including their action versions.

We don't want to also manage those action versions with Renovate on a per-provider basis, because it will just fight with ci-mgmt.

Instead, we should use Renovate to keep ci-mgmt's workflows using the correct action versions. Then ci-mgmt can ship those to providers, and Renovate won't need to touch provider workflows by default.

PR #1158, opened 2024-11-26
Start pinning action versions

blampe:
This starts pinning our action versions to the latest tag within the current refs we're using.

Our default action versions are now consolidated under `.github/workflows/default-action-versions.yml` so we can keep them consistent across native and bridged providers. More importantly, this will allow Renovate to start maintaining (and pinning) these actions for us automatically.

The net result of this will be that `ci-mgmt` can continue to act as the source of truth for workflows, and Renovate won't need to compete with it when it bumps dependencies.

By using more precise pins (e.g. `v1.1.0` vs. `v1`) this should avoid some of the ambiguity @t0yv0 ran into ([internal Slack](https://pulumi.slack.com/archives/C02FXTZEZ6W/p1732639814645349?thread_ts=1732561931.052099&cid=C02FXTZEZ6W)), but we'll need to confirm this.

Native providers were behind in a couple places and will be upgraded as part of this:
* `pulumi/pulumi-upgrade-provider-action@v0.0.5` → `0.0.12`
* `google-github-actions/auth@v0.9.0` → `2.1.7`

Refs https://github.com/pulumi/ci-mgmt/issues/1156


provider-ci/internal/pkg/config.go
```diff
@@ -345,6 +347,52 @@ type publish struct {
 func loadDefaultConfig() (Config, error) {
 	var config Config
 
+	// Walk up our directory tree until we can find GitHub workflows.
+	var wf workflow
+	cwd, _ := os.Getwd()
+	for attempt := 0; attempt < 10; attempt++ {
+		out, err := os.ReadFile(filepath.Join(cwd, ".github/workflows/default-action-versions.yml"))
```
danielrbradley, line 354: How does this work when run from a provider via `make ci-mgmt`? Would we not need to embed this file in the binary so we can `go run ..` it?

blampe: Good point, I forgot about that use case! I think you're right, we can embed this. I'll do that now.

t0yv0 (approved):

danielrbradley (approved):
Happy with the pinning and the approach. As with most things in this repo, we should really have more documentation somewhere (like the readme) on how these parts (renovate and provider-ci) are connected and how to perform updates manually if renovate does something wrong. 

Not blocking on this, but would love a version of this PR description somewhere in the repo!

PR #1176, opened 2024-11-27
Remove unused sshPrivateKey config

blampe:
Only the Docker provider uses the `webfactory/ssh-agent` action, and that's configured via [`preTest`](https://github.com/pulumi/pulumi-docker/blob/master/.ci-mgmt.yaml#L35-L39) instead of the special-purpose `sshPrivateKey` option.

Nothing else [appears](https://github.com/search?q=org%3Apulumi%20path%3A.ci-mgmt.yml%20sshPrivateKey&type=code) to use this, so we can remove it.

### DIFF
--- a/provider-ci/internal/pkg/config.go
+++ b/provider-ci/internal/pkg/config.go
@@ -205,12 +205,6 @@ type Config struct {
 	// https://github.com/search?q=org%3Apulumi+path%3A.ci-mgmt.yaml+%22docker%3A%22&type=code
 	Docker bool `yaml:"docker"`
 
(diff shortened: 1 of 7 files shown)


blampe:
@danielrbradley yeah that all sounds good. We can do this SSH setup in the docker tests directly, that way it won't need this hook or any other pretest stuff.

danielrbradley (approved):
Great - happy with this now we've got a route to removing the preTest hook via https://github.com/pulumi/pulumi-docker/pull/1300

PR #1186, opened 2024-11-28
Remove resync-build workflow

ringods:
The `resync-build` workflow copies a set of generated files from `ci-mgmt` into the respective provider repo. 

https://github.com/pulumi/ci-mgmt/blob/5c4c4b9781e6baeba99ed540065bcc4c2be763c9/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/resync-build.yml#L34

These files are no longer available in `ci-mgmt`, which makes this workflow deprecated. Cleaning up as a result.


iwahbe (approved):
Thanks for cleaning this up.

PR #1193, opened 2024-12-03
Fix checking out upstream without patches

danielrbradley:
Allow the wildcard listing operations to return zero results instead of failing with:

```
error: can't open patch '../patches/*.patch': No such file or directory

Failed to apply *.patch.
```


provider-ci/internal/pkg/templates/bridged-provider/upstream.sh
```diff
@@ -138,11 +138,14 @@ apply_patches() {
   # Iterating over the patches folder in sorted order,
   # apply the patch using a 3-way merge strategy. This mirrors the default behavior of 'git merge'
   cd upstream
+  # Allow directory to be empty
+  shopt -s nullglob
```
VenelinMartinov, line 142: 🥲

VenelinMartinov (approved):

danielrbradley:
Pulumi plugin install failed while testing in xyz due to rate limit it.

PR #1196, opened 2024-12-03
Reuse provider build for testing

### LINKED ISSUE #967: Test the provider binaries we're planning to release

### LINKED ISSUE #1033: Test provider binaries for publishing
Use the provider binary from the `build_provider` workflow when running tests rather than building a throw-away binary in pre-requisites.

It might be helpful to call `go generate` as part of the provider build to avoid requiring a dependency on the `prerequisites` job. We only need the version calculated before we can build the provider binary as the schema is already committed. We can verify that the committed schema is up to date in a standalone job.

t0yv0:
Fixes #967 
Fixes #1033

Unlike the ticket spec in #967 , only linux-amd64 binary is tested by the full set of integration tests. We might consider opening a follow up to smoke-test every single one of the platform builds, or else keep #967 open. 

As to the ticket spec in #1033 the dependency from build_provider to prerequisites is real because prerequisites builds the schema and the schema is needed in the binaries. This needs to stay.

Test run: https://github.com/pulumi/pulumi-azuread/pull/1589

Additional exploration: in prerequisites we run `make test_provider` which seems to try running unit tests in the `./provider` directory. Unfortunately some tests in `./provider` directory are also integration tests, e.g. in pulumi-aws and pulumi-azuread, and the convention of passing `go test -short` to only run the unit tests does not seem to be observed universally (pulumi-azuread breaks for example). For this reason removing `make provider` from prerequisites is not yet possible as it breaks our builds. This is likely out of scope of this PR.


provider-ci/internal/pkg/templates/bridged-provider/.github/actions/download-bin/action.yml
```diff
@@ -1,16 +1,23 @@
-name: Download binary assets
-description: Downloads the provider and tfgen binaries to `bin/`.
+name: Download the provider binary
```
t0yv0, line 1: This action is repurposed to only download the provider binary (not tfgen). It does that by unpacking the actual production build for linux-amd64 and not the custom "for-test" build.

provider-ci/internal/pkg/templates/bridged-provider/.github/actions/download-tfgen/action.yml
```diff
@@ -0,0 +1,17 @@
+name: Download the tfgen binary
```
t0yv0, line 1: This action is split out from download-bin to download the tfgen binary only.

provider-ci/internal/pkg/templates/bridged-provider/.github/actions/upload-bin/action.yml
```diff
@@ -1,15 +0,0 @@
-name: Upload bin assets
```
t0yv0, line 1: This action is removed and inlined in favor of a simple UploadArtifact invocation for the tfgen binary.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_sdk.yml
```diff
@@ -49,8 +49,8 @@ jobs:
           tools: pulumictl, pulumicli, ${{ matrix.language }}
       - name: Prepare local workspace
         run: make prepare_local_workspace
-      - name: Download bin
-        uses: ./.github/actions/download-bin
+      - name: Download tfgen
```
t0yv0, line 52: This only needs the tfgen binary.

blampe (approved):

PR #1202, opened 2024-12-05
Sign bridged windows binaries

blampe:
This separates the provider build/packaging steps so we can add signatures before uploading.

Signing is done using [Jsign](https://ebourg.github.io/jsign/), which has the advantage of being able run on Ubuntu.

Confirmed the signature is OK on a Windows VM:

image

Refs https://github.com/pulumi/home/issues/3766


danielrbradley (approved):
Wow that was suprisingly painless! Great to get this in!

Before this rolls out, we should ensure that the windows release is definitely usable. I think the xyz provider has windows post-release testing, but we should double check that is run and passes when we merge this.

blampe:
Validated here https://github.com/pulumi/pulumi-xyz/actions/runs/12242813485/job/34151349361

ringods:
@blampe @danielrbradley this now breaks the third party providers I rolled out `ci-mgmt` to:

https://github.com/pulumiverse/pulumi-acme/actions/runs/12248873082/job/34169377386
https://github.com/pulumiverse/pulumi-scaleway/actions/runs/12248856577/job/34169268696
https://github.com/pulumiverse/pulumi-cockroach/actions/runs/12248837708/job/34169220478

PR #1206, opened 2024-12-10
Only sign the windows binary when the signing secrets are configured

ringods:
The addition of signing Windows binaries unconditionally broke the setup for non-Pulumi managed providers:

https://github.com/pulumi/ci-mgmt/pull/1202#issuecomment-2530623379

This PR checks if the signing secrets have been configured and skips the signing step when not configured, as is currently the case for the providers I manage.

This is tested for the `acme` provider:

* [commit](https://github.com/pulumiverse/pulumi-acme/pull/95/commits/16ad22e76f06a89290c19b2a453ba7e019acb708)
* [corresponding run which skips "Sign Windows provider"](https://github.com/pulumiverse/pulumi-acme/actions/runs/12251907770/job/34177631020)


ringods:
With my change, the `xyz` provider is still signed:

https://github.com/pulumi/pulumi-xyz/actions/runs/12252090260/job/34178075178?pr=640

danielrbradley (approved):

PR #1210, opened 2024-12-11
Spin up cluster for cert-manager

blampe:
Fixes https://github.com/pulumi/pulumi-kubernetes-cert-manager/issues/251

### DIFF
--- a/native-provider-ci/src/steps.ts
+++ b/native-provider-ci/src/steps.ts
@@ -1270,7 +1270,10 @@ export function FreeDiskSpace(runner: string): Step {
 }
 
 export function CreateKindCluster(provider: string, name: string): Step {
-  if (provider === "kubernetes" && name === "run-acceptance-tests") {
+  if (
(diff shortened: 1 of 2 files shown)


native-provider-ci/src/steps.ts
```diff
@@ -1270,7 +1270,10 @@ export function FreeDiskSpace(runner: string): Step {
 }
 
 export function CreateKindCluster(provider: string, name: string): Step {
-  if (provider === "kubernetes" && name === "run-acceptance-tests") {
+  if (
+    (provider === "kubernetes" || provider == "kubernetes-cert-manager") &&
+    name === "run-acceptance-tests"
+  ) {
```
blampe, line 1276: In general we try to avoid special-casing like this, but we already have this exception for k8s so we might as well use it. Separately we'll be trying to pull these native providers into the same template we use for bridged ones.

rshade (approved):
LGTM

PR #1212, opened 2024-12-11
Always emit verbose test output

blampe:
This might be intentional for the sake of easier debugging of failures, but we're going to start moving some tests around for the sake of sharding and it will be very important to be able to see the exact tests being run. Otherwise we run the risk of accidentally dropping something.

Let's at least turn this on temporarily.


danielrbradley (approved):
Yes, this looks like a very sensible change.

Output from the downstream test: https://github.com/pulumi/pulumi-xyz/actions/runs/12283672594/job/34277829995#step:13:189

PR #1213, opened 2024-12-11
Remove integrationTestProvider option

blampe:
This is currently [unused](https://github.com/search?q=org%3Apulumi%20path%3A.ci-mgmt.yaml%20integrationTestProvider&type=code), and we don't want to encourage two separate places for integration tests. If a provider includes e2e tests in the `provider` directory, the recommendation is to consolidate those under `examples`.


danielrbradley (commented):
Happy to delete this if we really don't need it, but I'd question the direction on requiring all e2e tests to be in the `examples` module.

We've moved some e2e tests into the provider module in [Azure Native](https://github.com/pulumi/pulumi-azure-native/blob/b059fcc6c2030a8d534b41e50a01ad8231d605eb/provider/pkg/provider/provider_e2e_test.go#L50-L59) and [Azure Classic](https://github.com/pulumi/pulumi-azure/blob/b0ffdc75036657a5904a4b37c31bc26052d8538e/provider/provider_yaml_test.go#L100-L110) because it allows the tests to be much more productive for repro'ing issues. This is because the test is in the same module as the provider code and so when attaching the debugger and using the provider debug attach option, it lets you step right through from the test into the provider code.

That said, to make this work well, we do need to integrate these tests better into our CI flow as they're just just run along with the provider unit tests right now which can slow down the whole pipeline.

blampe:
> We've moved some e2e tests into the provider module in [Azure Native](https://github.com/pulumi/pulumi-azure-native/blob/b059fcc6c2030a8d534b41e50a01ad8231d605eb/provider/pkg/provider/provider_e2e_test.go#L50-L59) and [Azure Classic](https://github.com/pulumi/pulumi-azure/blob/b0ffdc75036657a5904a4b37c31bc26052d8538e/provider/provider_yaml_test.go#L100-L110) because it allows the tests to be much more productive for repro'ing issues. This is because the test is in the same module as the provider code and so when attaching the debugger and using the provider debug attach option, it lets you step right through from the test into the provider code.

Yeah, this is understandable. Was talking about this with Florian this morning [here](https://github.com/pulumi/pulumi-aws/pull/4910#issuecomment-2539684647).

The integration tests should _absolutely_ be in the same module as the provider code, and it's our fault for over-complicating things by creating more than one module. The best and simplest fix is to consolidate everything under a root module, which is straightforward -- move `provider/go.mod` to the root, copy any dependencies from `examples/go.mod` into it, fix `replace` statements, and rewrite imports. This doesn't impact SDKs since they already have their own module.

In other words this
```
.
├── examples
│   ├── go.mod
│   └── go.sum
├── provider
│   ├── go.mod
│   └── go.sum
└── test
```

becomes 

```
.
├── examples
├── go.mod
├── go.sum
├── provider
└── test
```
and now it _really_ doesn't matter where your integration tests live.

Issue #1034, opened 2024-07-11
Extract test shared workflow


danielrbradley:
The test job is currently duplicated in multiple workflows. The test process should be extracting into a single reusable workflow which can be called from each other top-level workflow.

Issue #1211, opened 2024-12-11
Remove special-casing for TestPulumiExamples


blampe:
`TestPulumiExamples` is an integration test specific to azure-native. It runs tests against the p/examples, which is very cool, but ci-mgmt doesn't need to know anything about it.

PR #1214, opened 2024-12-11
Consolidate test workflow

blampe:
* Updates `test-provider/aws/.ci-mgmt.yaml` to what's currently checked-in.
* Consolidates test jobs into a shared workflow using `run-acceptance-tests` as the starting point.

`extraTests` are included in the shared workflow mostly for simplicity. This will break aws because those tests will specify `needs:` that are no longer valid. We can fix that manually or just remove these extra tests (https://github.com/pulumi/pulumi-aws/pull/4910, https://github.com/pulumi/pulumi-aws/pull/4909).

No special treatment is given to TestPulumiExamples. We're not using this outside of azure-native, and it should really just be treated like any other integration test https://github.com/pulumi/ci-mgmt/issues/1211. If we don't want these tests to run in certain workflows then we should expose some info to let them `t.Skip` themselves.

Fixes https://github.com/pulumi/ci-mgmt/issues/1034


danielrbradley (commented):
Very excited to see this go in!

I think it would be worth opening tentative PRs to a few key providers to get feedback before merging this. AWS, Azure and GCP at least. It would be good to prepare the fixes for AWS before merging too so that we're not scrambling to fix things after this merges.

Issue #1215, opened 2024-12-12
Windows signing breaks if `freeDiskSpaceBeforeBuild: true` is set


blampe:
The signing step is currently using the runner's baked-in `az` CLI to login, and the free-disk-space action is cleaning it up.

PR #1216, opened 2024-12-12
Fix windows signing with freeDiskSpaceBeforeBuild

blampe:
Fixes https://github.com/pulumi/ci-mgmt/issues/1215.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_provider.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/build_provider.yml
@@ -31,7 +31,6 @@ jobs:
           - os: windows
             arch: amd64
     steps:
-      #{{- if .Config.FreeDiskSpaceBeforeBuild }}#
       # Run as first step so we don't delete things that have just been installed
       - name: Free Disk Space (Ubuntu)
         uses: #{{ .Config.ActionVersions.FreeDiskSpace }}#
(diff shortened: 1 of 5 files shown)


danielrbradley:
I would suggest, instead of using the login action, we should just not delete the AZ CLI that we need. This would give us the flexibility to move the signing to the makefile without needing to revert this.

The AZ CLI removal is part of the `large-packages` option in the action: https://github.com/jlumbroso/free-disk-space

The current disk space savings in AWS are:
```
=> Android library: Saved 7.6GiB
=> Haskell runtime: Saved 5.4GiB
=> Large misc. packages: Saved 6.3GiB
=> Docker images: Saved 2.5GiB
=> Overall 22GiB
```

This would therefore reduce to approximately 16GiB savings instead, which I think should suffice. Alternatively, we could just disable large packages on the build job only.

Of the 2 minutes spent saving disk space, 1m 44s is spent on the large packages, so disabling this option would also save a significant amount of time in each run. Which might make it feasible to turn back on for the xyz package too - to increase our test coverage.

blampe:
> I would suggest, instead of using the login action, we should just not delete the AZ CLI that we need. This would give us the flexibility to move the signing to the makefile without needing to revert this.

Yeah, I initially went that route but then figured 6GB was a bit too much to trade off. I'll go ahead and hook it up that way, the login action needs an additional secret anyway.

t0yv0 (approved):

Issue #190, opened 2022-05-20
Save build time by adding an option to break acceptance tests down further


jkodroff:
For our larger and older providers (e.g. pulumi-azure), tests are often nodeje-heavy because Pulumi originally only supported nodejs. This leads to unbalanced build times for our acceptance tests. In pulumi-azure, nodejs takes about 25 minutes and the other languages take about 5. We always need to wait for the slowest group of acceptance tests to finish before failing a build.

At present, we tie the `tags` in `go test` to the same matrix as the languages we build for. We can add an option to `config.yaml` per-provider to break these down even further, e.g. `node-group1, node-group2, python, go, csharp`. In order to avoid breaking changes, we'll default to the current per-language breakdown.

If we made this change and divided the nodejs tests approximately evenly between `node-group1` and `node-group2`, we should save about 12.5 minute of build time, which is nothing to sneeze at.

We should attempt this _first_ on a large provider with unbalanced testing times, then backport to this repo.

mjeffryes:
We now support arbitrary test sharding

Issue #1207, opened 2024-12-10
Manage secrets with ESC


blampe:
CI workflows often consume over a dozen secrets via environment variables -- some of which were set years ago and aren't available in 1pass. This makes it very difficult (and sometimes impossible) to reproduce behavior locally.

Instead we should manage these secrets via ESC so developers can easily unseal the same environment used by CI.

Something like https://github.com/pulumi/esc/issues/92 would be nice but isn't strictly necessary.

This would probably look like:
* And org-wide "pulumi" environment that all providers can share.
* Per-provider "pulumi-xyz" environments with provider-level overrides.

Migration could be gradual, with GitHub secrets moving into the org-wide environment as needed.

blampe:
Take together with https://github.com/pulumi/ci-mgmt/issues/1262 and https://github.com/pulumi/ci-mgmt/issues/1263 this would allow developers to get a local environment virtually identical to the one used in CI.

PR #1217, opened 2024-12-12
Allow bridged providers to opt-in to sharding

blampe:
Extracts the sharding part of https://github.com/pulumi/ci-mgmt/pull/1140 on
top of the test consolidation in https://github.com/pulumi/ci-mgmt/pull/1214.

Testing this in https://github.com/pulumi/pulumi-aws/pull/4912.

The template's test workflow gets messy because sharded tests need to setup all
SDKs instead of individual ones.

Sharding logic has been forked to http://github.com/pulumi/shard so anyone can contribute to it while I'm out. (It pains me that it's public, but GitHub requires forks to stay public for some reason.)

Fixes https://github.com/pulumi/ci-mgmt/issues/190


blampe:

- #1217 ◀

provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -379,6 +379,15 @@ provider_dist-darwin-arm64: bin/$(PROVIDER)-v$(VERSION_GENERIC)-darwin-arm64.tar
 provider_dist-windows-amd64: bin/$(PROVIDER)-v$(VERSION_GENERIC)-windows-amd64.tar.gz
 provider_dist: provider_dist-linux-amd64 provider_dist-linux-arm64 provider_dist-darwin-amd64 provider_dist-darwin-arm64 provider_dist-windows-amd64
 .PHONY: provider_dist-linux-amd64 provider_dist-linux-arm64 provider_dist-darwin-amd64 provider_dist-darwin-arm64 provider_dist-windows-amd64 provider_dist
-
+#{{ if .Config.Shards -}}#
+# shard computes tests to run and modifies the CI runner's environment.
+shard:
```
danielrbradley, line 384: We can't use $(GITHUB_ENV) to modify the environment as that won't be runnable locally. I really think we want to leverage the single target that's integrated via the existing test target which might be a little more fiddly to write but makes it easier for the caller to repro exactly what CI did.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -379,6 +379,15 @@ provider_dist-darwin-arm64: bin/$(PROVIDER)-v$(VERSION_GENERIC)-darwin-arm64.tar
...
+	@(cd #{{ .Config.E2ETestDirectory }}# && go run github.com/blampe/shard@latest --total $(SHARD_TOTAL) --index $(SHARD_INDEX) --output env) >> "$(GITHUB_ENV)"
```
danielrbradley, line 385: We need to move this out of your personal repo before we ship this so we can fix any issues while you're out.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/test.yml
```diff
@@ -39,29 +39,50 @@ jobs:
...
+      #{{- range $_, $language := .Config.Languages }}#
+    - name: Download #{{ $language }}# SDK
+      uses: ./.github/actions/download-sdk
+      with:
+        language: #{{ $language }}#
+    - name: Restore makefile progress
+      run: make --touch provider schema build_#{{ $language }}#
+    - name: Install dependencies
+      run: make install_#{{ $language }}#_sdk
+      #{{- end }}#
```
danielrbradley, line 78: I think we can make this less repetitive in the generated code by doing the touch and install outside the loop and leveraging the phony grouping targets.

```suggestion
      #{{- range $_, $language := .Config.Languages }}#
    - name: Download #{{ $language }}# SDK
      uses: ./.github/actions/download-sdk
      with:
        language: #{{ $language }}#
      #{{- end }}#
    - name: Restore makefile progress
      run: make --touch provider schema build_sdks
    - name: Install dependencies
      run: make install_sdks
```

provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -258,7 +258,7 @@ bin/$(PROVIDER): .make/schema
 
 test: export PATH := $(WORKING_DIR)/bin:$(PATH)
 test:
-	cd examples && go test -v -tags=all -parallel $(TESTPARALLELISM) -timeout 2h
+	cd #{{ .Config.E2ETestDirectory }}# && go test -v -tags=all -parallel $(TESTPARALLELISM) -timeout 2h
```
danielrbradley, line 261: Alternative design - proxy all test runs through your sharded test runner which could then apply the sharding in a single step, if needed:

```suggestion
	cd #{{ .Config.E2ETestDirectory }}# && go run github.com/pulumi/go-test-shard@v1 --total "$(SHARD_TOTAL)" --index "$(SHARD_INDEX)" -- -v -tags=all -parallel $(TESTPARALLELISM) -timeout 2h
```

The exact same change could be applied to the `provider_test` target too and could be moved from the prerequisites to the test matrix too.

danielrbradley (commented):
The CI side of things looks good, but I would strongly advocate for not adding multiple CI-specific targets in the makefile. Added some alternative design ideas inline.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/test.yml
danielrbradley: Additionally, we could allow the "Download SDK" action to accept multiple languages (e.g. comma separated or something) to make the call sight cleaner and more understandable.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
blampe: > The exact same change could be applied to the provider_test target too and could be moved from the prerequisites to the test matrix too.

I've been thinking of `provider_test` as essentially unit tests which make sense to keep in `prerequisites` as a lightweight sanity check before the heavy tests start.

See my comment [here](https://github.com/pulumi/ci-mgmt/pull/1213#issuecomment-2540167749) about integration tests in `provider`. Basically our poor module structure is pushing us towards some awkward conventions, and there are a ton of benefits to fixing the underlying issue instead of catering to those conventions.

_By default_ we can expect "slow" tests to live in one directory, but this doesn't have to be a requirement as long as the provider can eventually customize targets. With a root-level `go.mod` it becomes trivial to run integration tests across any number of directories -- in other words it's possible to specify `SHARD_PATHS="./examples ./provider"` when they belong to the same module.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
blampe: Totally agree with everything you're saying but would prefer to make those optimizations in a followup. It's nice that we manage the Makefile for these bridged providers because we're not locked in to anything here.

It is _possible_ to repro things locally by looking up the shard information
```
Run make test_shard
env:
    SHARD_TESTS: ^(?:TestACMUpgrade|TestAccLogGroup|TestAccRoute53|TestAccServerlessRaw|TestAccWebserverGo|TestEcrLifecyclePolicyUpgrade|TestIMDSAuth|TestLogGroupUpgrade|TestRdsInstanceUpgrade|TestRegress3421Update|TestRegress4128|TestSecretManagerPy)\$
    SHARD_PATHS: ./.
```

and you can generate that information yourself if you really want to (`GITHUB_ENV=env SHARD_TOTAL=10 SHARD_INDEX=2 make shard && cat env`).

So you can in theory run `SHARD_TESTS=... SHARD_PATHS=... make test_shard` locally. But in practice there are so many other environment variables I don't think this will actually be feasible until we've done something like https://github.com/pulumi/ci-mgmt/issues/1207.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/test.yml
blampe: Totally forgot about `install_sdks`, good call!

t0yv0:
Talking with @rquitales today we certainly would need this to onboard eks to the bridged template because without sharding EKS test suite is too slow. I'm looking at rebasing the PR and trying this out again.

t0yv0:
Closing in favor of https://github.com/pulumi/ci-mgmt/pull/1293 which is this rebased and simplified.

PR #1220, opened 2024-12-12
Verify actual published Python SDK version

### LINKED ISSUE #1219: Pass python SDK version into release verification
For pre-releases of SDKs, the Python package version is transformed by codegen to be compatible with PyPi's version format (which is not semver v2). When we attempt to fetch the PyPi package based on the provider version it will never be found for python prereleases.

Therefore, we need to extract the generated python package version and pass it into the verify-provider-release `packageVersion` field.

Related issues:
- https://github.com/pulumi/verify-provider-release/issues/74
- https://github.com/pulumi/verify-provider-release/issues/39

danielrbradley:
Fixes #1219

Extract SDK version after publishing the Python SDK. The SDK is downloaded internally by the pulumi-package-publisher action.

The xyz provider does not have a python example so we will need to manually test this against another provider.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/publish.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/publish.yml
@@ -104,6 +104,8 @@ jobs:
     name: publish_sdk
     needs: publish
     runs-on: #{{ .Config.Runner.Default }}#
+    outputs:
+      python_version: ${{ steps.python_version.outputs.version }}
     steps:
     - name: Checkout Repo
       uses: #{{ .Config.ActionVersions.Checkout }}#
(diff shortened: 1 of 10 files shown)


iwahbe (approved):
Looks good to me. Let's make sure we test it before merging.

danielrbradley:
Pushing to pulumi-random: https://github.com/pulumi/ci-mgmt/actions/runs/12296659267

danielrbradley:
## Testing fix of using packageVersion instead of providerVersion

Pushing to pulumi-random: https://github.com/pulumi/ci-mgmt/actions/runs/12297668850

danielrbradley:
This is now fully passing for pulumi-random 🎉

PR #1222, opened 2024-12-12
Don't populate the Go cache from the license check job

blampe:
GitHub cache artifacts are immutable, so the first job to write to the cache "wins" in the sense that all subsequent jobs will start with whatever that job happened to cache.

We currently use `setup-go` in the license check action, but this job is very quick (typically a couple minutes) and doesn't really build any artifacts or download modules. Subsequent jobs then need to re-download and re-compile all the provider's modules.

This PR disables caching for the license check job. This gives the `prerequisites` job an opportunity to more fully populate the cache.

Ideally we would only disable _writing_ to the cache, but there isn't an official way to do that yet (https://github.com/actions/setup-go/issues/316) and manually restoring the cache is fragile.

### DIFF
--- a/provider-ci/internal/pkg/templates/provider/.github/actions/setup-tools/action.yml
+++ b/provider-ci/internal/pkg/templates/provider/.github/actions/setup-tools/action.yml
@@ -14,6 +14,12 @@ inputs:
         dotnet
         java
     default: all
+  cache-go:
+    description: |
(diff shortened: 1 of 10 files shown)


danielrbradley (approved):

t0yv0:
Cool! I think it's possible to have separate named caches in the repo, but I've not looked in a bit.

t0yv0 (approved):

PR #1224, opened 2024-12-12
Upload unit test coverage for bridged providers by default

blampe:
Turn on codecov reporting for provider unit tests.

### DIFF
--- a/provider-ci/internal/pkg/config.go
+++ b/provider-ci/internal/pkg/config.go
@@ -317,6 +317,7 @@ type actionVersions struct {
 	UpgradeProviderAction   string `yaml:"upgradeProviderAction"`
 	FreeDiskSpace           string `yaml:"freeDiskSpace"`
 	ProviderVersionAction   string `yaml:"providerVersionAction"`
+	Codecov                 string `yaml:"codeCov"`
 }
 
 type toolVersions struct {
@@ -344,8 +345,6 @@ type publish struct {
 
 func loadDefaultConfig() (Config, error) {
 	var config Config
-
-	var wf workflow
 
 	// Parse our actions file while preserving comments.
 	var doc yaml.Node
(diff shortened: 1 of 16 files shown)


provider-ci/internal/pkg/config.go
```diff
@@ -394,42 +393,15 @@ func loadDefaultConfig() (Config, error) {
...
-			case "pulumi/pulumi-upgrade-provider-action":
-				config.ActionVersions.UpgradeProviderAction = s.Uses
-			case "jlumbroso/free-disk-space":
-				config.ActionVersions.FreeDiskSpace = s.Uses
-			case "pulumi/provider-version-action":
-				config.ActionVersions.ProviderVersionAction = s.Uses
-			}
-		}
-	}
-
```
blampe, line 432: Unused / leftover.

This is some hairy logic to preserve YAML comments. It would probably be simpler to forgo `config.ActionVersions` and just let Renovate manage the template files directly, but we can do that later.

provider-ci/test-providers/docker/Makefile
```diff
@@ -239,7 +239,7 @@ test:
 .PHONY: test
 
 test_provider:
-	cd provider && go test -v -short ./... -parallel $(TESTPARALLELISM)
+	cd provider && go test -v -short -coverprofile="coverage.txt" -coverpkg=./... ./... -parallel $(TESTPARALLELISM)
```
t0yv0, line 242: Do we need to spell out packages from ./upstream into coverpkgs or it's happening automatically? e.g. how are we covering the Terraform provider code itself.

t0yv0 (approved):

provider-ci/test-providers/docker/Makefile
blampe: We can capture coverage for multiple modules like so:
```
-coverpkg="./...,github.com/hashicorp/terraform-provider-..."
```
This captures all of our local modules and any TF provider imports.

I don't know how this will render in Codecov (I'm guessing it won't show any TF information) but it can be helpful locally.

PR #1227, opened 2024-12-13
Sign windows binaries via makefile

danielrbradley:
Move the logic for signing windows binaries into the makefile from the CI scripts. This allows testing with signed binaries locally.

- Fail with error if running in CI so we don't release unsigned binaries. Just warn locally.
- Add inline documentation explaining how to fix the missing configuration.
- Allow third party providers to skip signing by having none of the configuration set. Perhaps we could make this more explicit in the future, but that would potentially require changing the config on a bunch of providers - one side or the other.


t0yv0 (approved):
Ability to run outside of GHA is always an improvement. A bit heavy for Make in my taste but I'm amenable to where the team wants to take it.

danielrbradley:
Was a little on the fence about putting into into a bash script to keep the makefile cleaner, but we've not got precidence for that so this felt like the more acceptable change for now. Though, perhaps we should look at that as a follow-up?

provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -358,6 +365,35 @@ bin/%/$(PROVIDER) bin/%/$(PROVIDER).exe:
...
+		else \
+			mv $@ $@.unsigned; \
+			az login --service-principal \
+				--username "${AZURE_SIGNING_CLIENT_ID}" \
+				--password "${AZURE_SIGNING_CLIENT_SECRET}" \
+				--tenant "${AZURE_SIGNING_TENANT_ID}" \
+				--output none; \
+			ACCESS_TOKEN=$$(az account get-access-token --resource "https://vault.azure.net" | jq -r .accessToken); \
+			wget https://github.com/ebourg/jsign/releases/download/6.0/jsign-6.0.jar --output-document=bin/jsign-6.0.jar; \
+			java -jar bin/jsign-6.0.jar \
```
ringods, line 386: @danielrbradley if the intent is to test more easily locally, I'm tempted to make the download of the signing tool a separate make target so it isn't downloading it over and over again.

ringods (approved):

provider-ci/internal/pkg/templates/bridged-provider/Makefile
danielrbradley: Yes, could pull that out

PR #1228, opened 2024-12-16
Update pulumi package publisher to v0.0.21

iwahbe:
Update the version of `pulumi-package-publisher` used to v0.0.21.

It looks like `.Config.Publisher.PublisherAction` hasn't worked in the last 6 months: https://github.com/pulumi/ci-mgmt/pull/1007. Given that it hasn't worked (presumably by accident) and that we have heard no complaints, I'm going to remove the functionality in a follow-up PR.

Fixes https://github.com/pulumi/pulumi-package-publisher/issues/44.

PR #1007, opened 2024-06-26
Extract reusable publish workflow

danielrbradley:
- Combine release, prerelease and main into a single workflow with minimal options.
- This enables verifying all releases except Go on main. We can't verify Go because we don't release Go because that would clutter the git history with tags.

## Prerelease -> Release differences

Here's a diff between `prerelease.yml` and `release.yml` before the change. The observed changes are:
1. Extra `IS_PRERELEASE: true` env var
2. Optional extra `create_docs_build` job in `release.yml`.
3. Customised goreleaser args
4. `enableMacosRunner` is disabled for pre-release.
5. Workflow name and tags filters.

```diff
4d3
<   IS_PRERELEASE: true
22a22,40
>   #{{ if .Config.publishRegistry -}}#
>   create_docs_build:
>     name: create_docs_build
>     needs: publish_go_sdk
>     runs-on: #{{ .Config.runner.default }}#
>     steps:
>       - name: Dispatch Metadata build
>         uses: peter-evans/repository-dispatch@v3
>         with:
>           token: ${{ secrets.PULUMI_BOT_TOKEN }}
>           repository: pulumi/registry
>           event-type: resource-provider
>           client-payload: |-
>             {
>               "project": "${{ github.repository }}",
>               "project-shortname": "#{{ .Config.provider }}#",
>               "ref": "${{ github.ref_name }}"
>             }
>   #{{ end -}}#
79,80c97
<         args: -p #{{ .Config.parallel }}# -f .goreleaser.prerelease.yml --rm-dist --skip-validate --timeout
<           #{{ .Config.timeout }}#m0s
---
>         args: -p #{{ .Config.parallel }}# release --rm-dist --timeout #{{ .Config.timeout }}#m0s
138a156,176
> 
>   clean_up_release_labels:
>     name: Clean up release labels
>     #{{ if .Config.publishRegistry -}}#
>     needs: create_docs_build
>     #{{ else }}#
>     needs: tag_sdk
>     #{{- end }}#
>     runs-on: #{{ .Config.runner.default }}#
>     steps:
>     - name: Checkout Repo
>       uses: actions/checkout@v4
>     - name: Clean up release labels
>       uses: pulumi/action-release-by-pr-label@main
>       with:
>         command: "clean-up-release-labels"
>         repo: ${{ github.repository }}
>         commit: ${{ github.sha }}
>       env:
>         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
> 
255,256c293
[...]

PR #1229, opened 2024-12-16
Remove `.Config.Publish.PublisherAction` in favor of renovate

iwahbe:
Follow-up on https://github.com/pulumi/ci-mgmt/pull/1228.

This PR removes the `.Config.Publish.PublisherAction` option. It looks like .Config.Publisher.PublisherAction hasn't worked in the last 6 months: #1007. Given that it hasn't worked (presumably by accident) and that we have heard no complaints, it seems safe to remove the option and reduce the configuration complexity in `ci-mgmt`.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/publish.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/publish.yml
@@ -121,13 +121,13 @@ jobs:
         tools: pulumictl, pulumicli, #{{ range $index, $element := .Config.Languages }}##{{if $index}}#, #{{end}}##{{ $element }}##{{end}}#
     - name: Publish SDKs
       if: inputs.skipJavaSdk == false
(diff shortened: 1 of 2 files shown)


ringods:
@iwahbe should the publisher action be listed here?

https://github.com/pulumi/ci-mgmt/blob/master/provider-ci/internal/pkg/action-versions.yml

It seems this file is created on purpose for Renovate to detect & match new versions.

iwahbe:
> @iwahbe should the publisher action be listed here?
> 
> https://github.com/pulumi/ci-mgmt/blob/master/provider-ci/internal/pkg/action-versions.yml
> 
> It seems this file is created on purpose for Renovate to detect & match new versions.

I don't think so, but I'm not 100% sure. That file was created by Renovate, so my assumption is that any necessary maintenance will also be done by Renovate.

t0yv0 (approved):

PR #1233, opened 2024-12-17
Set permissions on the calling side.

ringods:
The `permissions` block in a reusable workflow doesn't grant these permissions. These are more used as an early indication of the permissions required by the reusable workflow. The correct permissions still need to be set on the calling side.

See `Reusable workflows` in [this blog article](https://www.kenmuse.com/blog/github-actions-workflow-permissions/).

Follow up to #1214.

Without setting the `permissions` on the calling side, the reusable workflow will not have the correct permissions.

The Pulumiverse runs failed without the required `permissions` block:
https://github.com/pulumiverse/pulumi-acme/actions/runs/12365788538
After adding the `permissions` block on the calling side:
https://github.com/pulumiverse/pulumi-acme/actions/runs/12369028130

https://github.com/pulumiverse/pulumi-acme/pull/98

For Pulumiverse, this fails because the organization wide workflow settings are configured to `Read repository contents and packages permissions`.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
@@ -122,6 +122,9 @@ jobs:
       - prerequisites
       - build_provider
       - build_sdk
+    permissions:
+      contents: read
+      id-token: write
     secrets: inherit
     with:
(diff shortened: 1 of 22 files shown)


ringods:
Contributes to #1087

iwahbe (approved):
This looks good to me.

PR #1235, opened 2024-12-17
Always spin up KinD cluster for pulumi-kubernetes-* providers in test step

rquitales:
We need to create the KinD cluster in all workflows that run tests, not just the `run-acceptance` workflow. For the pulumi-kubernetes provider, we don't need to create a KinD cluster as we create a GKE cluster for testing instead.

Closes: https://github.com/pulumi/pulumi-kubernetes-cert-manager/issues/254


blampe (approved):
Reasonable to unbreak things.

Just so you're aware, we actually just consolidated the test workflow for bridged providers in https://github.com/pulumi/ci-mgmt/pull/1214, which is nice because it prevents bugs like this.

In the future I'll be looking into moving this setup out of CI and into the tests.

PR #1238, opened 2024-12-19
Consolidate go modules for bridged providers

blampe:
This introduces a migration to automatically consolidate `./{provider,examples,tests}/go.mod` under a single `./go.mod`. The SDK module isn't modified so users are not impacted.

Example: https://github.com/pulumi/pulumi-aws/pull/4919

This gives us significantly more flexibility with regard to how we choose to organize our unit and integration tests. In particular it becomes straightforward to run integration tests under `./examples` and `./providers` simultaneously because they now belong to the same module. (We currently run one and then wait for it to complete before starting the other, but this sacrifices parallelism while waiting for long polls to complete.)

This is essentially an alternative to the `integrationTestProvider` option with a number of add benefits, including better performance and an improved overall developer experience. 

https://github.com/pulumi/ci-mgmt/pull/1039 (removing arbitrary injection hooks) will be further [unblocked](https://github.com/search?q=org%3Apulumi+path%3A.ci-mgmt.yaml&type=code) after this:
* Pre-test hooks for things like credentials ([example](https://github.com/pulumi/pulumi-aws/blob/f9da959d00c6f781a7b2e245ab77ea45579090eb/.ci-mgmt.yaml#L55-L63)) can be handled with some logic in [TestMain](https://pkg.go.dev/testing#hdr-Main).
* The "Run provider tests" step ([example](https://github.com/pulumi/pulumi-meraki/blob/0b2c4ba8043cd4fdb1f01a11e460844fb751bfd8/.ci-mgmt.yaml#L22-L25)) can be removed because executing `go test -tags=${{ matrix.language }} ...` from the repository root will now capture integration tests under `./examples` as well as `./provider`.

This only benefits bridged providers, but we are very close to being able to put native providers on the same (or very similar) workflows as bridged providers -- at which point the same migration can be applied.

Related:
* https://github.com/pulumi/ci-mgmt/pull/1213#issuecomment-2540167749
* https://github.com/pulumi/pulumi-aws/pull/4910#issuecomment-2539684647
* https://github.com/pulumi/ci-mgmt/pull/1226
* https://github.com/pulumi/ci-mgmt/issues/936


blampe:
@ringods wanted to put this on your radar. This would move us towards a more idiomatic layout for these projects but I don't want it to be a surprise for pulumiverse.

ringods (changes requested):
@blampe may I ask for one addition to this PR?

Users working with VSCode to manage a provider (me included), add a `go.work` and `go.work.sum` file at the root of the repo currently. This is required for VSCode to have proper comletion in the different folders which currently have a `go.mod` file.

In this PR, you consolidate these `go.mod` files into a single one at the root. Can you either add `go.work` and `go.work.sum` to [getDeletedFiles](https://github.com/pulumi/ci-mgmt/blob/1243b9e2dff8ae4d4391591bcec61a38daeac0af/provider-ci/internal/pkg/generate.go#L110) or make it part of your migration?

provider-ci/internal/pkg/migrations/consolidate_modules.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+		_, err = run("go", "mod", "edit", fmt.Sprintf("-replace=%s=%s", old, r.New.Path))
+		if err != nil {
+			return fmt.Errorf("replacing %q: %w", old, err)
+		}
+	}
+
+	// Remove examples/tests modules. We'll recover their requirements with a
+	// `tidy` at the end. It's OK if these don't exist.
+	_, _ = run("git", "rm", "examples/go.mod")
```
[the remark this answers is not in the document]
blampe: @ringods yes, `./sdk` is unchanged and will remain in your `go.work`.

ringods (approved):

provider-ci/Makefile
```diff
@@ -30,6 +30,9 @@ lint:
 format:
 	go fmt ./...
 
+unit:
```
t0yv0, line 33: Is this necessary? The tests this will run are not exactly unit tests.

t0yv0 (approved):

provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -332,7 +332,7 @@ ci-mgmt: .ci-mgmt.yaml
 	fi
 
 # Compute the version of Pulumi to use by inspecting the Go dependencies of the provider.
-.pulumi/version: provider/go.mod
+.pulumi/version: $(wildcard go.mod provider/go.mod) # go.mod can exist in either of these places.
```
danielrbradley, line 335: Would we need the wildcard once migrated? Or is this just for non-bridged providers who share this file?

danielrbradley (approved):
Just noticed this on my notifications backlog. Happy with this approach.

Issue #1101, opened 2024-10-12
[Epic] Standardize on provider-ci for all providers


mjeffryes:
## Overview

Only bridged providers benefit from the consistency and automatic rollout of workflows generated from provider-ci today. This creates a lot of work when we have to make a change to all providers. Native provider ci also helps automate changes for some non-bridged providers but has also varied in structure to provider-ci. We'd like to drive greater consistency for across the whole fleet for things like package publish to gain the benefits of investment here across the whole fleet. 

## Key KPIs

- Number of implementations required to roll out a publishing change to all providers.
- Number of duplicate implementations where behaviour is the same.

## Key Stakeholders

- Provider team
- Community provider authors

## Key Deliverables

- New provider-ci templates for EKS, AWSx and API Gateway
- Integrate provider-ci into boilerplates
- Migrate native-provider-ci into provider-ci
- Minimise duplication & fragmentation

### References 📔

- [ ] [Project View](https://github.com/orgs/pulumi/projects/227/views/1) 
- [ ] PR/FAQ 
- [ ] Design Doc 
- [ ] UX Designs 
- [ ] Decision Log

t0yv0:
Why does the title say Standardize on CI-mgmt for all providers and the overview says "Bridged providers currently display unusable or misleading results in pulumi preview." ?

mjeffryes:
Probably because I failed a copy-paste! thanks for the catch!

danielrbradley:
Looks like this probably supersedes:
- https://github.com/pulumi/ci-mgmt/issues/936

Most of that issue is complete at this point. The only dangling issues are:
- https://github.com/pulumi/ci-mgmt/pull/1039
- https://github.com/pulumi/ci-mgmt/pull/1037

blampe:
This week we pulled most of the differences in the generic template into the bridged one https://github.com/pulumi/ci-mgmt/pull/1140. In particular sharding and some workflow consolidation that makes subsequent changes easier.
* https://github.com/pulumi/ci-mgmt/pull/1214 
* https://github.com/pulumi/ci-mgmt/pull/1217 

We've also cleaned up a bunch of instances where providers had be using custom workflows:
* https://github.com/pulumi/ci-mgmt/pull/1176
* https://github.com/pulumi/pulumi-aws/pull/4910
* https://github.com/pulumi/pulumi-aws/pull/4917
* https://github.com/pulumi/pulumi-aws/pull/4918

[...]

PR #1239, opened 2024-12-19
AWS Native: Add AWS credentials before Test Provider Library step

mikhailshilkov:
In https://github.com/pulumi/pulumi-aws-native/pull/1919 I want to run an actual end-2-end test with `pulumitest` in AWS Native, which requires AWS credentials. In understand that a larger consolidation may be coming with https://github.com/pulumi/ci-mgmt/issues/1101, but this change would allow me to land my test ahead of that. Please let me know if there is a better way of achieving the same outcome.


flostadler:
I think this approach is fine as a stop-gap until native providers are on `provider-ci` as well! Code changes look good to me.

flostadler (approved):

PR #1243, opened 2024-12-20
Adding in sdk fixes, like #1112

rshade:


rshade:
I need that config to tweak kind, for nginx. I am not sure that is the problem though.

rshade:
@blampe do you know what I am doing wrong here?

blampe:
@rshade run `make` locally in the root of the repo, that will re-generate all of the checked-in fixtures. It helps us lint things and see what will eventually get pushed to repos.

rshade:
`with` needing a blank `config` is something I am going to need to test on cert-manager, before we apply this.

rshade:
it looks to be fine here: https://github.com/pulumi/pulumi-kubernetes-cert-manager/pull/279

native-provider-ci/src/steps.ts
```diff
@@ -1325,15 +1325,19 @@ export function CreateKindCluster(provider: string, name: string): Step {
...
       node_image: "kindest/node:v1.29.2",
+      config: "",
     },
   };
 
   switch (provider) {
     case "kubernetes":
       return name === "run-acceptance-tests" ? step : {};
     case "kubernetes-cert-manager":
+      return step;
```
EronWright, line 1336: You don't want it to fallthrough to `step.with.config = "kind.config.yml";` ?

EronWright (approved):
I have the impression that we seek to standardize the providers' CI, so if the version package is typically `pkg/provider/version` then maybe we should update the non-standard providers rather than indulge them.

native-provider-ci/src/goreleaser.ts
```diff
@@ -114,14 +114,26 @@ export class PulumiGoreleaserPreConfig implements GoreleaserConfig {
...
-      );
+      if (opts["major-version"] > 1) {
+        ldflags.push(
+          `-X github.com/pulumi/pulumi-${opts.provider}/provider/v${opts["major-version"]}/pkg/version.Version={{.Tag}}`
+        );
+      } else {
+        ldflags.push(
+          `-X github.com/pulumi/pulumi-${opts.provider}/provider/pkg/version.Version={{.Tag}}`
+        );
+      }
```
danielrbradley, line 136: Let's revert this.

native-provider-ci/src/workflows.ts
```diff
@@ -859,8 +859,20 @@ export class PublishGoSdkJob implements NormalJob {
...
+      case "kubernetes-cert-manager":
+        this.if = false;
+        break;
+      case "kubernetes-coredns":
+        this.if = false;
+        break;
+      case "kubernetes-ingress-nginx":
+        this.if = false;
+        break;
+    }
```
danielrbradley, line 875: Let's revert this too

danielrbradley (changes requested):
As per discussion on call. Let's revert the 2 commented parts, re-generate and then we should have a much smaller change to merge

native-provider-ci/src/steps.ts
```diff
@@ -1319,20 +1319,25 @@ export function CreateKindCluster(provider: string, name: string): Step {
...
-  const step = {
+  const step: Step = {
     name: "Setup KinD cluster",
     uses: action.createKindCluster,
     with: {
       cluster_name: "kind-integration-tests-${{ matrix.language }}",
       node_image: "kindest/node:v1.29.2",
+      ...(provider === "kubernetes-ingress-nginx"
+        ? { config: "kind.config.yml" }
+        : {}),
```
EronWright, line 1330: I'm confused why there'd be a "Setup KinD cluster" step that didn't have a config. What's the behavior in that case? What's special about nginx provider?  Asking that you capture the knowledge.

PR #1248, opened 2024-12-23
Automatically accept SDK changes for Renovate PRs

blampe:
We have a problem right now where automated dependency upgrades can change our generated code, and this causes tests to fail because we expect no changes in CI.

The "right" way to do this is with a hermetic `make renovate` step which can re-generate this code during the Renovate job, as @t0yv0 has started doing for EKS (e.g. https://github.com/pulumi/pulumi-eks/pull/1552). But we don't have hermetic builds so we don't yet have an easy way to get this to work across all providers.

This PR implements a suggestion from Daniel to help address the long tail. Essentially right before we fail a build for containing SDK changes, we will commit those changes and push them back to the PR -- _but only if it came from Renovate._ The build will still fail, but tests will be re-tried against the updated SDK.

Renovate has already been configured to "ignore" updates from pulumi-bot (https://github.com/pulumi/renovate-config/commit/82931a62d50a9b7a628032bdefbf3c6f64a4e895), so if tests pass it will still automatically squash and merge.

Here's an example where I manually modified the SDK and the PR was automatically corrected:

image

Refs https://github.com/pulumi/ci-mgmt/issues/1101
Refs https://github.com/pulumi/ci-mgmt/issues/936

### DIFF
--- a/native-provider-ci/providers/aws-native/repo/.github/workflows/build.yml
+++ b/native-provider-ci/providers/aws-native/repo/.github/workflows/build.yml
@@ -108,6 +108,7 @@ jobs:
     - name: Build Provider
       run: make provider
     - name: Check worktree clean
+      id: worktreeClean
       uses: pulumi/git-status-check-action@54000b91124a8dd9fd6a872cb41f5dd246a46e7c # v1.1.1
       with:
         allowed-changes: |-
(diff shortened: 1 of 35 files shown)


t0yv0 (approved):
Let's try this.

blampe:
Verified 
* https://github.com/pulumi/pulumi-random/pull/1514
* https://github.com/pulumi/pulumi-docker-build/pull/380

PR #1252, opened 2024-12-27
Handle sdk-gen during prerequisites

blampe:
Some native providers, particularly go-providers like p-docker-build and p-command, generate their SDKs during the prerequisites step. This is a roadblock for Renovate https://github.com/pulumi/pulumi-command/actions/runs/12519817033/job/34924483839?pr=609

This tweaks our native workflows to allow committing SDK changes during the prerequisites step.

This ends up touching release, prerelease, and other workflows as a result. I added an additional check -- `github.event_name == 'pull_request'` -- to safeguard against any of this running during releases etc.

I also added a check to confirm `sdk` was touched before we try to commit it, to avoid situations where `prerequisites` might touch something else like `schema.json`.

I had a couple changes locally which I forgot to push to https://github.com/pulumi/ci-mgmt/pull/1248 which also bring native behavior more in line with bridged: checking out the repo with `persist-credentials: false` (without this tests won't re-run after we push) and `fail-fast: false` for Renovate PRs.

### DIFF
--- a/native-provider-ci/providers/aws-native/repo/.github/workflows/build.yml
+++ b/native-provider-ci/providers/aws-native/repo/.github/workflows/build.yml
@@ -50,7 +50,6 @@ jobs:
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       with:
         lfs: true
-        persist-credentials: false
     - id: version
       name: Set Provider Version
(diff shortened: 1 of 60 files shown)


t0yv0:
To confirm, is this better than brute-forcing these providers to conform to the expected step scheme and generate the SDKs where expected?

t0yv0 (approved):
Let's try it.

PR #1255, opened 2024-12-30
Fix errors surfaced from Make

t0yv0:
Currently both CI and local executions surface errors such as:

```
set -e; \
if [[ "darwin-arm64" = "windows-amd64" && false != "true" ]]; then \
	if [[ "|***|***|***|***|" == *"||"* ]]; then \
		echo "Can't sign windows binaries as required configuration not set: AZURE_SIGNING_CLIENT_ID, AZURE_SIGNING_CLIENT_SECRET, AZURE_SIGNING_TENANT_ID, AZURE_SIGNING_KEY_VAULT_URI"; \
		echo "To rebuild with signing delete the unsigned bin/darwin-arm64/pulumi-resource-aws and rebuild with the fixed configuration"; \
		if [[ true == "true" ]]; then exit 1; fi; \
	else \
		mv bin/darwin-arm64/pulumi-resource-aws bin/darwin-arm64/pulumi-resource-aws.unsigned; \
		az login --service-principal \
			--username "***" \
			--password "***" \
			--tenant "***" \
			--output none; \
		ACCESS_TOKEN=$(az account get-access-token --resource "https://vault.azure.net" | jq -r .accessToken); \
		java -jar bin/jsign-6.0.jar \
			--storetype AZUREKEYVAULT \
			--keystore "PulumiCodeSigning" \
			--url "***" \
			--storepass "${ACCESS_TOKEN}" \
			bin/darwin-arm64/pulumi-resource-aws.unsigned; \
		mv bin/darwin-arm64/pulumi-resource-aws.unsigned bin/darwin-arm64/pulumi-resource-aws; \
		az logout; \
	fi; \
fi
/bin/sh: 2: [[: not found
```

The fix turns `${CI}` into `"${CI}"` and `${SKIP_SIGNING}` into `"${SKIP_SIGNING}"`. The additional layer prevents Make-level substitution from generating a syntactically invalid shell script.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/Makefile
+++ b/provider-ci/internal/pkg/templates/bridged-provider/Makefile
@@ -45,7 +45,7 @@ LDFLAGS=$(LDFLAGS_PROJ_VERSION) $(LDFLAGS_UPSTREAM_VERSION) $(LDFLAGS_EXTRAS) $(
 
 # Create a `.make` directory for tracking targets which don't generate a single file output. This should be ignored by git.
 # For targets which either don't generate a single file output, or the output file is committed, we use a "sentinel"
-# file within `.make/` to track the staleness of the target and only rebuild when needed. 
+# file within `.make/` to track the staleness of the target and only rebuild when needed.
 # For each phony target, we create an internal target with the same name, but prefixed with `.make/` where the work is performed.
 # At the end of each internal target we run `@touch $@` to update the file which is the name of the target.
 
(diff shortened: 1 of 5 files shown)


guineveresaenger (approved):
Oh no. 😂  That's so unfortunate, and so Bash.

LGTM.

iwahbe (approved):
Ah... Bash + Make = 😒

danielrbradley (approved):
Bah, good catch. Shellcheck would have caught this, but it can't parse scripts inside of makefiles 😞

PR #1256, opened 2024-12-30
Adding in support for kind.config.yml which has labels for nginx

rshade:


rshade:
@blampe do you know what I am doing wrong here?

PR #1259, opened 2025-01-02
Add pulumistack provider

blampe:
https://github.com/pulumi/pulumi-pulumistack

This thing has never been published but Joe needs it for some reason. Let's at least try to keep its workflows under control.


rshade (approved):
I can't see pulumi-stack, so I am assuming the url is right. But looks good to me

Issue #1191, opened 2024-12-03
Onboard pulumi-awsx to ci-mgmt


mjeffryes:
Use the generic ci-mgmt template to manage AWSX

t0yv0:
Instead of using the non-existing generic template we are opting to use a variation of a bridged template to onboard AWSX, POC here: https://github.com/pulumi/pulumi-awsx/pull/1471 - it looks like it's working well enough. 

The complete checklist would be:

- [ ] onboard
- [ ] add release verification
- [ ] perform an alpha release
- [ ] check the rel

PR #1268, opened 2025-01-04
[WIP] Alternative approach for native providers

blampe:
This is an alternative to https://github.com/pulumi/ci-mgmt/pull/1140, on top of https://github.com/pulumi/ci-mgmt/pull/1217.

Instead of forking the bridged template -- which will inevitably cause the kind of drift we want to avoid -- this adds a new parameter to the bridged template to disable bridge-specific things like `make upgrade`, etc.

I don't know if this is even feasible. I do like that it ships a `Makefile` to the native provider. A couple targets are no-op'd, some CI workflows are omitted, and instead of invoking tfgen we invoke the provider's gen binary. This means the manual part of onboarding native providers will probably look more like getting the codegen binaries to be invoked more like tfgen.

There are still references to tfgen/upstream/etc. but this is cosmetic. I think this is worth exploring more as it might make onboarding easier.


t0yv0:
I would like to try this for awsx to fix https://github.com/pulumi/ci-mgmt/issues/1191 to see what the issues would be.

t0yv0:
We've essentially mainlined this piecemeal with @rquitales and used to onboard AWSX, with aws-apigateway and EKS coming next.

PR #1269, opened 2025-01-06
Fix upstream not refreshing after pull

danielrbradley:
When running `git pull`, the upstream HEAD will not be updated if recursive submodule checkout is not enabled (as is the default).

1. Add `./upstream.sh file_target` command which just prints the upstream HEAD file path, if it exists (mimicking the existing wildcard statement).
2. Before printing, mark the file as updated if it doesn't match the new desired commit which will then trigger the upstream init to run and update the submodule.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/upstream.sh
+++ b/provider-ci/internal/pkg/templates/bridged-provider/upstream.sh
@@ -23,6 +23,7 @@ COMMANDS
   check_in              Write checkedout commits back to patches, add upstream
                         and patches changes to the git staging area and exit
                         checkout mode.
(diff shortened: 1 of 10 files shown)


provider-ci/internal/pkg/templates/bridged-provider/upstream.sh
```diff
@@ -334,6 +335,21 @@ re-initializing using updated patches and updated upstream base.
 EOF
 }
 
+file_target() {
```
iwahbe, line 338: Could you add a doc comment that explains what `file_target` does?

iwahbe (approved):

Issue #1265, opened 2025-01-04
Add release verification to aws


mjeffryes:
We should hook up some examples the to release verification workflow to smoke test releases. Additional context and examples in parent.

corymhall:
closed by https://github.com/pulumi/pulumi-aws/pull/5034

PR #1274, opened 2025-01-08
Run preTest actions during release verification workflow

corymhall:
The ci-mgmt config has a configuration property for `preTest` jobs. These are run during the `test` workflow before the tests themselves are run. Since the verify release workflow runs the same tests we should also run these `preTest` steps. For example, in `pulumi-aws` we have `preTest` jobs to get AWS credentials which are required to run the tests in the verify release workflow.

re https://github.com/pulumi/pulumi-aws/pull/5049, re #1265

### DIFF
--- a/provider-ci/internal/pkg/templates/provider/.github/workflows/verify-release.yml
+++ b/provider-ci/internal/pkg/templates/provider/.github/workflows/verify-release.yml
@@ -80,6 +80,9 @@ jobs:
         with:
           tools: pulumicli, #{{ range $index, $element := .Config.Languages }}##{{if $index}}#, #{{end}}##{{ $element }}##{{end}}#
 #{{- if .Config.ReleaseVerification }}#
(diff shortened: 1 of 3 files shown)


iwahbe (approved):
This is stretching the definition of `.Config.Actions.PreTest`, but since release verification is basically just running an example test, I think this is ok.

danielrbradley:
@corymhall we might need to revert this as it's caused a P1 in azure – master is broken. Pre-test currently has the assumption that it's only run in the test job, which has some addition setup including gotestfmt.

- https://github.com/pulumi/pulumi-azure/issues/2895

I would agree this is a slightly odd place to call preTest as release verification isn't running any tests - just downloading and executing the released SDKs and provider binary.

> Since the verify release workflow runs the same tests

I don't think this is true. We don't run any Go tests in the verify job.

PR #1275, opened 2025-01-08
Reusing bridged provider makefile for native providers

t0yv0:
Continuing with the experiment of reusing bridged provider makefile for native providers started in https://github.com/pulumi/ci-mgmt/pull/1268 when using pulumi-awsx as a test case. 

I managed to get `make schema` working by adjusting the conventions in the repo to the bridged conventions.

The next hiccup is that AWSX uses yarn to create provider binaries. This necessitates a small extension point to overload build_provider_cmd. Experimenting with this.


t0yv0:
This PR served its purpose for POC, and has now been merged piecemeal.

PR #1276, opened 2025-01-10
Revert "Run preTest actions during release verification workflow"

danielrbradley:
Reverts pulumi/ci-mgmt#1274

This caused a P1 in the pulumi-azure provider:
- https://github.com/pulumi/pulumi-azure/issues/2895

We can revisit an alternative way of setting up required credentials for AWS. See: https://github.com/pulumi/ci-mgmt/pull/1274#issuecomment-2582085346


danielrbradley:
All usages have been fixed instead

PR #1278, opened 2025-01-10
Generalize bridged template for reuse with NoUpstream providers

t0yv0:
This is https://github.com/pulumi/ci-mgmt/pull/1268 without the sharding. This should be safe as it is not affecting any providers per se yet, just building up a capability.


provider-ci/internal/pkg/config.go
```diff
@@ -265,6 +265,10 @@ type Config struct {
 	// MakeTemplate has no effect but is set by 78 providers.
 	// https://github.com/search?q=org%3Apulumi+path%3A.ci-mgmt.yaml+%22makeTemplate%3A%22&type=code
 	MakeTemplate string `yaml:"makeTemplate"`
+
+	// NoUpstream is a temporary hack to disable bridge-specific workflow steps
+	// as part of the work to cnsolidate these with native providers.
```
flostadler, line 270:
```suggestion
	// as part of the work to consolidate these with native providers.
```

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-bridge.yml
```diff
@@ -1,3 +1,4 @@
+#{{ if not .Config.NoUpstream -}}#
```
flostadler, line 1: What does GitHub do if it encounters an empty yaml file in the workflows directory? Does that work?

provider-ci/internal/pkg/config.go
```diff
@@ -265,6 +265,10 @@ type Config struct {
 	// MakeTemplate has no effect but is set by 78 providers.
 	// https://github.com/search?q=org%3Apulumi+path%3A.ci-mgmt.yaml+%22makeTemplate%3A%22&type=code
 	MakeTemplate string `yaml:"makeTemplate"`
+
+	// NoUpstream is a temporary hack to disable bridge-specific workflow steps
```
corymhall, line 269: What is temporary about it? What do we plan on doing long term?

provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -287,7 +287,11 @@ tfgen_build_only: bin/$(TFGEN)
 bin/$(TFGEN): provider/*.go provider/go.* .make/upstream
 	(cd provider && go build $(PULUMI_PROVIDER_BUILD_PARALLELISM) -o $(WORKING_DIR)/bin/$(TFGEN) -ldflags "$(LDFLAGS_PROJ_VERSION) $(LDFLAGS_EXTRAS)" $(PROJECT)/$(PROVIDER_PATH)/cmd/$(TFGEN))
 .PHONY: tfgen schema tfgen_no_deps tfgen_build_only
-
+#{{ if .Config.NoUpstream }}#
```
flostadler, line 290: Shouldn't this test for `.Config.NoUpstream` being set to false?

provider-ci/internal/pkg/templates/bridged-provider/upstream.sh
```diff
@@ -1,3 +1,4 @@
+#{{ if not .Config.NoUpstream -}}#
```
flostadler, line 1: Is there a way how we could exclude those files from being generated all together if `.Config.NoUpstream` is set to true?
I find it a bit odd that we drop empty scripts into the repos

provider-ci/internal/pkg/config.go
t0yv0: Yeah I'm taking this comment from Bryce's original PR. It might have been a part of a larger plan.

provider-ci/internal/pkg/templates/bridged-provider/upstream.sh
t0yv0: Yeah, true I'll have a look.

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-bridge.yml
t0yv0: Looks like the empty files just don't get generated (see https://github.com/pulumi/pulumi-awsx/pull/1471/files )

rquitales: GitHub actions will skip/ignore the empty yaml file

provider-ci/internal/pkg/templates/bridged-provider/upstream.sh
t0yv0: Looks like they are excluded in fact.

provider-ci/internal/pkg/templates/bridged-provider/Makefile
rquitales: No. I believe it should be `NoUpsteam == true`, as this branch is just creating a dummy upstream folder instead of actually populating the upstream submodule folder.

t0yv0: So code is correct yes? here's how it renders for AWS (https://github.com/pulumi/ci-mgmt/blob/22b93b47bda086268f861b83fc9e9aabab5bc1d1/provider-ci/test-providers/aws/Makefile#L266) - no changes, which is good. The new form renders for eks.

https://github.com/pulumi/ci-mgmt/blob/22b93b47bda086268f861b83fc9e9aabab5bc1d1/provider-ci/test-providers/eks/Makefile#L255

rquitales: @t0yv0 Yes, the code is correct. My message was sent around the same time as your reply 😄

provider-ci/internal/pkg/templates/bridged-provider/upstream.sh
flostadler: That's cool!

rquitales (approved):

PR #1284, opened 2025-01-13
Automatically install pulumictl

t0yv0:
Bridged provider Makefile used to assume that `pulumictl` is available in PATH. With this change it will now automatically install it if this is not the case. This is easier to use and also will help `make renovate` targets succeed without having `pulumictl` installed.

The version of the tool is controlled so that it is in sync across the Makefile and CI.

### DIFF
--- a/provider-ci/internal/pkg/generate.go
+++ b/provider-ci/internal/pkg/generate.go
@@ -51,7 +51,7 @@ func GeneratePackage(opts GenerateOpts) error {
 	}
 
 	if opts.Config.ToolVersions.PulumiCTL == "" {
-		opts.Config.ToolVersions.PulumiCTL = "v0.0.46"
+		opts.Config.ToolVersions.PulumiCTL = defaultPulumiCTLVersion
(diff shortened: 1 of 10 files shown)


provider-ci/internal/pkg/generate.go
```diff
@@ -50,6 +50,10 @@ func GeneratePackage(opts GenerateOpts) error {
 		opts.Config.CheckUpstreamUpgrade = false
 	}
 
+	if opts.Config.ToolVersions.PulumiCTL == "" {
+		opts.Config.ToolVersions.PulumiCTL = "v0.0.46"
```
rquitales, line 54: Can we make this default value a Go constant somewhere more visible?

rquitales (approved):

provider-ci/internal/pkg/generate.go
t0yv0: Yeah let me take a look where the other default versions are specd

PR #1285, opened 2025-01-13
Update yarn command in bridge Makefile

TheJokersThief:
This might have changed in a recent version of `yarn`, but I ran into an issue where `--cwd` was required to be the first argument after the `yarn` command to work, otherwise it would throw an error:

```
Usage Error: The --cwd option is ambiguous when used anywhere else than the very first 
parameter provided in the command line, before even the command path
```

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/Makefile
+++ b/provider-ci/internal/pkg/templates/bridged-provider/Makefile
@@ -231,7 +231,7 @@ install_go_sdk:
 install_java_sdk:
 install_nodejs_sdk: .make/install_nodejs_sdk
 .make/install_nodejs_sdk: .make/build_nodejs
-	yarn link --cwd $(WORKING_DIR)/sdk/nodejs/bin
+	yarn --cwd link $(WORKING_DIR)/sdk/nodejs/bin
 	@touch $@
 install_python_sdk:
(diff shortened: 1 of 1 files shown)


provider-ci/internal/pkg/templates/bridged-provider/Makefile
```diff
@@ -231,7 +231,7 @@ install_go_sdk:
 install_java_sdk:
 install_nodejs_sdk: .make/install_nodejs_sdk
 .make/install_nodejs_sdk: .make/build_nodejs
-	yarn link --cwd $(WORKING_DIR)/sdk/nodejs/bin
+	yarn --cwd link $(WORKING_DIR)/sdk/nodejs/bin
```
t0yv0, line 234: It looks like this will be interpreted as `--cwd link` that is current working directory set to `link`. What errors are you having from the default invocation?

TheJokersThief: I'm getting the error I mentioned in my PR description 

```
Usage Error: The --cwd option is ambiguous when used anywhere else than the very first 
parameter provided in the command line, before even the command path
```

I've since realised this is because you expect yarn 1.x whereas modern installs are using 4.x. Gonna close this PR :)

t0yv0: That makes sense yes! yarn 1.x comes with GitHub runners by default still so we have not yet migrated to yarn 2.x.

PR #1296, opened 2025-01-16
Fix binary signing

t0yv0:
Fixes binary signing for Node binaries and darwin-arm64.


danielrbradley (approved):
Looks good. No harm in having the extra step even though unused. We could feature-flag this to only nodejs based providers in the future.

t0yv0:
It seeems to be very quick (likely uses GHA cache or some such), perhaps no harm in leaving in.

PR #1299, opened 2025-01-16
Fix release verification to cover Mac OS appropriately

t0yv0:
This makes sure that release verification runs jobs on MacOS as intended.

Example release in https://github.com/pulumi/pulumi-xyz/pull/756 - see https://github.com/pulumi/pulumi-xyz/actions/runs/12817331557

Looks like there were a few bugs in the code:

- different variable naming
- treatment of boolean vs string
- github.event.inputs vs inputs - according to GitHub docs these treat booleans differently (!), but it has to be inputs.* I believe to be available in both workflow_call context and workflow_dispatch context

### DIFF
--- a/provider-ci/internal/pkg/templates/provider/.github/workflows/verify-release.yml
+++ b/provider-ci/internal/pkg/templates/provider/.github/workflows/verify-release.yml
@@ -26,7 +26,7 @@ on:
         description: "The version of the provider to verify"
         required: true
         type: string
-      enableMacosRunner:
+      enableMacRunner:
         description: "Enable the macos-latest runner in addition to ubuntu-latest and windows-latest. Defaults to 'false'."
         required: false
         type: boolean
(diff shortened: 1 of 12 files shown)


provider-ci/internal/pkg/templates/provider/.github/workflows/verify-release.yml
```diff
@@ -64,7 +64,7 @@ jobs:
         # Expression expands to ["ubuntu-latest","windows-latest"] or ["ubuntu-latest","windows-latest","macos-latest"]
         # GitHub expressions don't have 'if' statements, so we use a ternary operator to conditionally include the MacOS runner suffix.
         # See the docs for a similar example to this: https://docs.github.com/en/actions/learn-github-actions/expressions#fromjson
-        runner: ${{ fromJSON(format('["ubuntu-latest","windows-latest"{0}]', github.event.inputs.enableMacRunner == 'true' && ',"macos-latest"' || '')) }}
+        runner: ${{ fromJSON(format('["ubuntu-latest","windows-latest"{0}]', inputs.enableMacRunner && ',"macos-latest"' || '')) }} #
```
rquitales, line 67: If only we had a more readable way to express this.

rquitales: nit: trailing "#"

rquitales (approved):
Good catch!

provider-ci/internal/pkg/templates/provider/.github/workflows/verify-release.yml
t0yv0: There could be a more readable way but testing is hard, so i just tested this one and would like to roll with it.

PR #1312, opened 2025-01-22
Split out bridge-independent CI code

danielrbradley:
Restructure existing template so we can build out more flexible variations of workflows for native providers without lots of conflig switches for excluding ceratain files or blocks

- Separate the CI from the `bridge-provider` template folder into `provider-ci` template folder so it can be used without the bridge makefile.
- Keep only `update-bridge` and `upgrade-provider` workflows in the bridge-specific template directory as they only apply to bridged providers.


provider-ci/internal/pkg/templates/provider-ci/.github/workflows/upgrade-bridge.yml
mjeffryes, line 1: This isn't actually general though right? it's specific to bridged providers

danielrbradley: Yes, that one should move back

PR #1316, opened 2025-01-24
Allow customizing test folder directory and adding kubectl installation

rquitales:
This PR contains 2 changes that will not be squashed when merged to maintain their history.

Changes:

- Enable specifying if `kubectl` needs to be installed in a workflow
- Add a configuration option to specify where integration tests are located. Defaults to `examples` if unspecified

### DIFF
--- a/provider-ci/internal/pkg/config.go
+++ b/provider-ci/internal/pkg/config.go
@@ -301,6 +301,9 @@ type Config struct {
 	// This function is called by the make renovate target after the Renovate bot has finished updating
 	// project dependencies in a PR. It is responsible for rebuilding any generated files as needed.
 	RenovateCmd string `yaml:"renovateCmd"`
+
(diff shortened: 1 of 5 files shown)


provider-ci/internal/pkg/config.go
```diff
@@ -301,6 +301,13 @@ type Config struct {
 	// This function is called by the make renovate target after the Renovate bot has finished updating
 	// project dependencies in a PR. It is responsible for rebuilding any generated files as needed.
 	RenovateCmd string `yaml:"renovateCmd"`
+
+	// Kubectl determines if we need to install the kubectl CLI. Used in pulumi-eks currently.
+	Kubectl bool `yaml:"kubectl"`
```
t0yv0, line 306: Maybe call this "InstallKubectl" ? Does it need version instead of true/false, or which version does it install?

t0yv0 (approved):

PR #1318, opened 2025-01-25
Add windows signing for native provider releases

rquitales:
This PR enables Windows singing for native provider releases made with GoReleaser. These changes require the `make sign-goreleaser-exe` target to exist. This is done in the individual provider repos as the Makefiles are not managed centrally by ci-mgmt.

For testing purposes, these changes are also copied to their respective native provider repos with a prerelease tag cut to validate that the binaries are signed.

Confirmed native providers with signed prerelease windows builds:

- [kubernetes](https://github.com/pulumi/pulumi-kubernetes/actions/runs/12996104839)
- [aws-native](https://github.com/pulumi/pulumi-aws-native/actions/runs/12996105418)
- [command](https://github.com/pulumi/pulumi-command/actions/runs/12996105991)
- [docker-build](https://github.com/pulumi/pulumi-docker-build/actions/runs/12996106508)
- [kubernetes-cert-manager](https://github.com/pulumi/pulumi-kubernetes-cert-manager/actions/runs/12997632596)
- [kubernetes-coredns](https://github.com/pulumi/pulumi-kubernetes-coredns/actions/runs/12997647427)
- [kubernetes-ingress-nginx](https://github.com/pulumi/pulumi-kubernetes-ingress-nginx/actions/runs/12997658107)
- [pulumistack](https://github.com/pulumi/pulumi-pulumistack/actions/runs/12996108738)
- [google-native](https://github.com/pulumi/pulumi-google-native/actions/runs/12996109430)


native-provider-ci/src/goreleaser.ts
```diff
@@ -162,15 +163,28 @@ export class PulumiGoreleaserPreConfig implements GoreleaserConfig {
...
         env: ["CGO_ENABLED=0", "GO111MODULE=on"],
-        goos: ["darwin", "windows", "linux"],
+        goos: ["darwin", "linux"],
         goarch: ["amd64", "arm64"],
         ignore: ignores,
         main: `./cmd/pulumi-resource-${opts.provider}/`,
         ldflags: ldflags,
         binary: `pulumi-resource-${opts.provider}`,
       },
+      {
```
rquitales, line 176: Building for Windows needs to be separated out to allow calling for a post build hook.

danielrbradley (commented):
It's quite confusing that this is named goreleaser - we don't use goreleaser anywhere any more. Edit: oh do we still goreleaser but only for native CI?

It's also a bit unfortunate that we're exposing another custom make target that gets called in a specific way. Can we not make this more transparent and just a configuration option when calling the normal build targets?

danielrbradley:
In fact, can we just copy the approach taken already in all bridged providers and also avoid drift in the approaches?

t0yv0 (approved):
I'm OK with that if tested e2e. IF we're planning to rip out goreleaser sounds like we forgot to scope that in? So deal with it in another unit of work?

PR #1332, opened 2025-01-30
Set gcloud auth permissions on each publish step 

guineveresaenger:
Followup to https://github.com/pulumi/ci-mgmt/pull/1331 which was supposed to fix https://github.com/pulumi/pulumi-gcp/issues/2919 but instead pushed the permissions problem up into the next parent Workflow.
Upshot is that the "invalid Workflow" error happens in a different file: https://github.com/pulumi/pulumi-gcp/actions/runs/13042448162/workflow#L101 

I believe this is all of the places where we need to tell Actions to inherit slightly elevated permissions.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
@@ -87,6 +87,9 @@ jobs:
     name: publish
     permissions:
       contents: write
+      #{{- if .Config.GCP }}#
+      id-token: write
+      #{{- end }}#
     needs:
       - prerequisites
       - build_provider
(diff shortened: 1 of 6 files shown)


provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
```diff
@@ -87,6 +87,9 @@ jobs:
     name: publish
     permissions:
       contents: write
+      #{{- if .Config.GCP }}#
+      id-token: write
+      #{{- end }}#
```
danielrbradley, line 92: Perhaps we could just enable this for every provider for the sake of consistency and less code path forks?

danielrbradley (approved):
Looks fine to me 👍

provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/main.yml
guineveresaenger: For permissions a tighter scope is perhaps more advised? But I'll leave it up to your team! <3

PR #1333, opened 2025-01-30
Unconditional id-token permissions

guineveresaenger:
If we limit permissions via template conditionals, it means that our test proivder xyz cannot validate that code path, leading us to guess in the dark at valid workflow configurations.
This pull request implements the suggestion from here: https://github.com/pulumi/ci-mgmt/pull/1332#discussion_r1935180350.
- **Set contents: write and id-token: write unconditionally, so we can validate the workflows**
- **test providers**


t0yv0 (approved):

danielrbradley (approved):
👍

PR #1335, opened 2025-01-31
Do not pollute workdir during pulumictl auto-installation

t0yv0:
When pulumictl is not discovered in PATH it is auto-installed by the Makefile. 

Unfortunately the previous method used a GOPATH override to guide `go install` as to where install pulumictl. As a side effect of this, the auto-installation did not use the standard Go package cache and polluted the working dir with another copy of Go package cache.

After this change, auto installation happens against the standard Go package cache and the result is then copied into bin.

Users should prefer to have pulumictl in the environment than relying on the auto-installation.

Tested manually on pulumi-awsx.

### DIFF
--- a/provider-ci/internal/pkg/defaults.go
+++ b/provider-ci/internal/pkg/defaults.go
@@ -1,5 +1,5 @@
 package pkg
 
 const (
-	defaultPulumiCTLVersion = "v0.0.46"
+	defaultPulumiCTLVersion = "v0.0.48"
 )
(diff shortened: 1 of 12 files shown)


provider-ci/internal/pkg/defaults.go
```diff
@@ -1,5 +1,5 @@
 package pkg
 
 const (
-	defaultPulumiCTLVersion = "v0.0.46"
+	defaultPulumiCTLVersion = "v0.0.48"
```
t0yv0, line 4: The 0.0.46 version had replaces in go.mod that caused go install to complain. The latest version has no such problem.

VenelinMartinov (approved):

ringods (approved):

corymhall:
@t0yv0 is this ready to be merged?

mjeffryes:
I think @danielrbradley beat you to landing this in https://github.com/pulumi/ci-mgmt/pull/1355

PR #1341, opened 2025-02-03
Extract templates function to own file with docs

danielrbradley:
1. Make it easier to find the key template composition code.
2. Add more documentation around the concepts and content.

### DIFF
--- a/provider-ci/internal/pkg/templates.go
+++ b/provider-ci/internal/pkg/templates.go
@@ -1,0 +1,41 @@
+package pkg
+
+import (
+	"fmt"
+)
+
+// Template directories
+const (
+	// configuration to run the repository in a dev container
+	devContainer = "dev-container"
+	// git attributes, go config, CI reusable workflows
+	provider = "provider"
+	// CoC, upgrade-provider config, issue templates, command dispatch workflows
+	pulumiProvider = "pulumi-provider"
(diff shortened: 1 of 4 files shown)


provider-ci/internal/pkg/templates.go
```diff
@@ -0,0 +1,41 @@
...
+	devContainer = "dev-container"
+	// git attributes, go config, CI reusable workflows
+	provider = "provider"
+	// CoC, upgrade-provider config, issue templates, command dispatch workflows
+	pulumiProvider = "pulumi-provider"
+	// ci-mgmt pull-based updates workflow
+	externalProvider = "external-provider"
+	// CI and Makefile for bridged providers (internal & external)
+	bridgedProvider = "bridged-provider"
+)
```
iwahbe, line 19: While we are here, can we make this a proper enum? Preferably with a encapsulating struct:

```go
type TemplateDir struct{name string}
var (
	// configuration to run the repository in a dev container
	DevContainer = TemplateDir{"dev-container"}
	...
)
```

At minimum, with a new type:

```go
type TemplateDir string
const (
	// configuration to run the repository in a dev container
	DevContainer TemplateDir = "dev-container"
	...
)
```

iwahbe (approved):
Approval conditional on a new type being added for template directories.

PR #1345, opened 2025-02-03
Allow auto-merging upstream provider upgrades

VenelinMartinov:
This PR adds a config `autoMergeProviderUpgrades` which will mark upstream provider upgrades to be auto-merged. I've also moved the upgrade-provider call to use the upgrade-provider action like we do with bridge upgrades.

Related to https://github.com/pulumi/ci-mgmt/issues/1344

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-provider.yml
+++ b/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-provider.yml
@@ -80,14 +80,20 @@ jobs:
         # upstream_version will be empty if the provider is up-to-date
         run: echo "version=${{ github.event.inputs.version || steps.upstream_version.outputs.latest_version }}" >> "$GITHUB_OUTPUT"
         shell: bash
-      - name: Attempt provider upgrade
+      - name: Call upgrade provider action
         id: upgrade_provider
-        # Only attempt the upgrade if we have a target version
         if: steps.target_version.outputs.version != ''
-        # Don't mark the build as failed if we can't auto-open a PR as we've already opened the upgrade issue for tracking
         continue-on-error: true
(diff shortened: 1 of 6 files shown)


provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-provider.yml
```diff
@@ -80,14 +80,21 @@ jobs:
...
-        # Don't mark the build as failed if we can't auto-open a PR as we've already opened the upgrade issue for tracking
         continue-on-error: true
-        run: upgrade-provider "${{ github.repository }}" --kind="all" --target-version="${{ steps.target_version.outputs.version }}" #{{ if .Config.JavaGenVersion }}#--java-version="#{{ .Config.JavaGenVersion }}#"#{{ end }}#
-        shell: bash
+        uses: #{{ .Config.ActionVersions.UpgradeProviderAction }}#
+        with:
+          kind: all
+          email: bot@pulumi.com
+          username: pulumi-bot
+          # TODO
```
guineveresaenger, line 92: This looks like it should be removed.

guineveresaenger (approved):
Dangling TODO, otherwise LGTM.

My main question here is whether switching using the upgrade provider action loses us some outputs in case a run fails, and also consider that it introduces versioning, meaning we'd need to release the action every time we push a critical fix/update to upgrade-provider. (I think that's Good, Actually - just something to think about.)

VenelinMartinov:
upgrade-provider-action uses upgrade-provider@main, so we shouldn't have any differences here:
https://github.com/pulumi/pulumi-upgrade-provider-action/blob/699de04d401c93c74824805b8d3e2936369f97db/action.yml#L95C60-L96C16

upgrade-provider-action is already used for the bridge upgrades.

PR #1346, opened 2025-02-03
Organise template directories

danielrbradley:
1. Flatten the dev-container config into the general provider internal template. This adds the dev-container support to the "generic" providers.
2. Rename internal template directory names for clarity so they're only a single word:
   - Rename "provider" to "base"
   - Rename "pulumi-provider" to "internal"
   - Rename "external-provider" to "external"
   - Rename "bridged-provider" to "bridged"


provider-ci/test-providers/eks/devbox.lock
```diff
@@ -0,0 +1 @@
+{}
```
thomas11, line 1: should that be committed?

thomas11 (approved):

provider-ci/test-providers/eks/devbox.lock
danielrbradley: Yup ... from https://github.com/pulumi/ci-mgmt/issues/721

> - add an empty devbox.lock file to the bridged provider template so that the devcontainer will build.

danielrbradley:
Replaced by https://github.com/pulumi/ci-mgmt/pull/1360

PR #1358, opened 2025-02-04
Generalise management of prerequisites artifacts

danielrbradley:
Avoid needing to hard code the name of the gen/tfgen binary .. if we even have one for the project. This is a step towards removing the "temporary hack" of the `NoUpstream` config option.

- Capture whole `bin` directory instead of just the codegen binary.
- Create upload action to mirror the download action.
- Rename action to reflect coverage of all assets from the job.
- Automatically capture and restore all executable permissions.
- This avoids having to hard-code the name of the generation binary (if we even have one).


provider-ci/internal/pkg/templates/provider-ci/.github/actions/upload-prerequisites/action.yml
```diff
@@ -0,0 +1,23 @@
+name: Upload SDK asset
+description: Upload the SDK for a specific language as an asset for the workflow.
+
+runs:
+  using: "composite"
+  steps:
+    - name: Capture executable permissions
+      shell: bash
+      run: find bin -type f -perm +111 > bin/executables.txt
```
rquitales, line 9: This doesn't work for GNU find:
```
-perm +mode
              This is no longer supported (and has been deprecated since
              2005).  Use -perm /mode instead.
```
Ref: https://man7.org/linux/man-pages/man1/find.1.html

danielrbradley: Good find. Have updated.

provider-ci/internal/pkg/templates/provider-ci/.github/actions/download-prerequisites/action.yml
```diff
@@ -0,0 +1,28 @@
...
+  steps:
+    - name: Download the prerequisites bin
+      uses: #{{ .Config.ActionVersions.DownloadArtifact }}#
+      with:
+        name: prerequisites-bin
+        path: bin
+
+    - name: Restore executable permissions
+      shell: bash
+      run: chomod +x $(< bin/executables.txt)
```
rquitales, line 15:
```suggestion
      run: chmod +x $(< bin/executables.txt)
```

danielrbradley: 😅

rquitales (approved):

PR #1360, opened 2025-02-05
Simplify internal template naming & organisation

danielrbradley:
Each provider template is a combination of:
1. `base` - files common to all providers (this contains the majority of all templates)
2. `internal` or `external` - which add customisations for providers owned by Pulumi or providers owned by 3rd parties.
3. `bridged` - for bridged providers–not generic providers such as eks or pulumiservice
4. `internal-bridged` - adds a single special customisation for non-third party bridged providers.

This has no impact on the generated files except for adding the missing devContainer support for "generic" providers. 

I'd recommend use the separated commits for reviewing each logical step.

## Steps

1. Combine "devContainer" and "provider" into "base"
   - This adds the devContainer support to "generic" providers.

2. Flatten "provider-ci" into "base"
   - "provider-ci" is use in all templates so can just live in base to avoid adding additional splitting of folders.

3. Rename "pulumi-provider" to "internal"
   -This pairs better with "external" which is used for third party providers.

4. Rename "external-provider" to "external"
   - Everything's a provider within "provider-ci" so no need for the extension.

5. Rename "bridged-provider" to "bridged"
   - Drop the "provider" suffix.

6. Add "internal-bridged" template
   - This is files which are only for bridged providers, owned by Pulumi. This was requested to not be included for external providers because they configure upgrades more manually.
   - This removes the empty file being generated for "generic" providers and allows us to remove the NoUpstream conditional hack.

7. Move Makefile to "base"
   - This is currently used for all templates, not only bridged providers.

8. Remove "bridged" folder from "generic"
   - The generic template no longer renders any of the files from the bridged template as they're all wrapped in conditionals which are never true for generic provider configurations.
   - This allows us to remove the NoUpstream conditionals from the two files in the bridged template.

9. Remove empty .upgrade.config.yml for generic providers
   - This was intended to be excluded by a global conditional round the template, but was actually generated as an empty file instead which was not the desired outcome.


thomas11 (approved):
This is great, looks much cleaner!

PR #1362, opened 2025-02-05
Fix noupstream in makefile

danielrbradley:
Removes the NoUpstream option hack from #1278 

This replaces the last uses of the option within the makefile. This adds one, more targeted option for the `GenName` which can be used to change the `tfgen` binary name to `gen` for "generic" providers.

We can replace the conditional around the upstream target with a check to see if upstream.sh exists before executing it.

The `LDFLAGS_UPSTREAM_VERSION` conditional had no effect and so can just be removed.

> Somewhat confusingly, `ProviderVersion` is only set [for 3 providers](https://github.com/search?q=org%3Apulumi+path%3A.ci-mgmt.yaml+%22providerVersion%3A%22&type=code) (aws, gcp & azure) in order to set the version within the upstream codebase. Therefore, generic providers would never set this option and we don't need to add the extra guard.


rquitales (approved):

danielrbradley:
To investigate:

```
make: *** No rule to make target 'is', needed by '.make/upstream'.  Stop.
```

danielrbradley:
Something was going weird on linux when it worked fine on mac, so I've removing the magic bash command and just moved the skipping if upstream isn't configured into the script itself, which is now just always included for all providers, rather than only bridged providers. 

The concept of upstream isn't unique to bridged providers as native providers also have similar concepts for specification sources etc, and there's nothing bridge-related in the script itself - it's just managing updating a git submodule and optionally applying locally stored patches.

thomas11 (approved):

PR #1379, opened 2025-02-11
Allow missing resource docs for Tier2+ providers

VenelinMartinov:
This PR changes all Tier2+ providers to allow missing resource docs when running a provider upgrade. This is a common issue which happens and we have to manually work around.

Fixes https://github.com/pulumi/upgrade-provider/issues/303

Tier 1 providers are exempt as the config is explicitly false there:
https://github.com/pulumi/pulumi-gcp/pull/2971
https://github.com/pulumi/pulumi-aws/pull/5193
https://github.com/pulumi/pulumi-azure/pull/3003
https://github.com/pulumi/pulumi-azuread/pull/1877

### DIFF
--- a/provider-ci/test-providers/aws/.github/workflows/upgrade-provider.yml
+++ b/provider-ci/test-providers/aws/.github/workflows/upgrade-provider.yml
@@ -86,7 +86,7 @@ jobs:
           username: pulumi-bot
           automerge: false
           target-version: ${{ steps.target_version.outputs.version }}
-          allow-missing-docs: true
+          allow-missing-docs: false
       - name: Comment on upgrade issue if automated PR failed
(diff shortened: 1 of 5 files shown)


provider-ci/test-providers/aws/.github/workflows/upgrade-provider.yml
```diff
@@ -86,7 +86,7 @@ jobs:
           username: pulumi-bot
           automerge: false
           target-version: ${{ steps.target_version.outputs.version }}
-          allow-missing-docs: false
+          allow-missing-docs: true
```
t0yv0, line 89: Still kind of confusing :) Perhaps stick false in the "for-test" ci-mgmt.yml?

PR #1381, opened 2025-02-12
Scaffolding a new parameterized-go template

t0yv0:
Downsizing generic template to fit https://github.com/pulumi/pulumi-terraform-module-provider

One key difference is there's no SDKs. There is no schema generation either. 

Minor differences: go.mod at top-level, slightly different repository name (-provider).

Approach:

- Makefile rewritten by taking the bridged makefile and simplifying it
- Per Daniel's suggestion shared sections of Makefile moved into .mk include files to not duplicate code here
- ModulePath config added to guide the code on where go.mod is (provider/go.mod vs go.mod)
- NoSchema config added to be able to remove all code related to schema computation or SDK handling

Verified by trying this code in pulumi-terraform-module-provider and getting through PR checks, pre-release and release.


provider-ci/internal/pkg/templates/parameterized-go/Makefile
```diff
@@ -0,0 +1,262 @@
...
+# the desired version, we just print. Otherwise we (re)install pulumi at the desired
+# version.
+.pulumi/bin/pulumi: .pulumi/version
+	@if [ -x .pulumi/bin/pulumi ] && [ "v$$(cat .pulumi/version)" = "$$(.pulumi/bin/pulumi version)" ]; then \
+		echo "pulumi/bin/pulumi version: v$$(cat .pulumi/version)"; \
+		touch $@; \
+	else \
+		curl -fsSL https://get.pulumi.com | \
+			HOME=$(WORKING_DIR) sh -s -- --version "$$(cat .pulumi/version)"; \
+	fi
```
danielrbradley, line 167: This would be nice as a shared makefile include perhaps in `scripts/pulumi-cli.mk` (we can't use the `.mk` folder because of the existing `include $(wildcard .mk/*.mk)`).

```suggestion
include scripts/pulumi-cli.mk
```

provider-ci/internal/pkg/templates/parameterized-go/Makefile
```diff
@@ -0,0 +1,262 @@
...
+provider: bin/$(PROVIDER)
+
+# `make provider_no_deps` builds the provider binary directly, without ensuring that
+# `cmd/pulumi-resource-#{{ .Config.Provider }}#/schema.json` is valid and up to date.
+# To create a release ready binary, you should use `make provider`.
+provider_no_deps:
+	$(call build_provider_cmd,$(shell go env GOOS),$(shell go env GOARCH),$(WORKING_DIR)/bin/$(PROVIDER))
+bin/$(PROVIDER):
+	$(call build_provider_cmd,$(shell go env GOOS),$(shell go env GOARCH),$(WORKING_DIR)/bin/$(PROVIDER))
+.PHONY: provider provider_no_deps
```
danielrbradley, line 126: 1. Drop the `build_provider_cmd` stuff as that's only needed when re-purposing the makefile for typescript builds.
2. Drop the legacy `provider_no_deps` - it's not needed.

provider-ci/internal/pkg/templates/parameterized-go/Makefile
```diff
@@ -0,0 +1,262 @@
...
+	@echo "  provider_dist Build and package the provider for all platforms"
+	@echo ""
+	@echo "Tool Targets"
+	@echo "  ci-mgmt     Re-generate CI configuration from .ci-mgmt.yaml"
+	@echo ""
+	@echo "Internal Targets (automatically run as dependencies of other targets)"
+	@echo "  prepare_local_workspace  Prepare for building"
+	@echo "  install_plugins          Install plugin dependencies"
+	@echo ""
+.PHONY: help
```
danielrbradley, line 77: Perhaps extract:

```suggestion
include scripts/help.mk
```

provider-ci/internal/pkg/templates/parameterized-go/Makefile
```diff
@@ -0,0 +1,262 @@
...
+
+# Check version doesn't start with a "v" - this is a common mistake
+ifeq ($(shell echo $(PROVIDER_VERSION) | cut -c1),v)
+$(error PROVIDER_VERSION should not start with a "v")
+endif
+
+# Strips debug information from the provider binary to reduce its size and speed up builds
+LDFLAGS_STRIP_SYMBOLS=-s -w
+LDFLAGS_PROJ_VERSION=-X $(GO_MODULE)/pkg/version.Version=$(PROVIDER_VERSION)#{{if .Config.ProviderVersion}}# -X #{{ .Config.ProviderVersion }}#=$(PROVIDER_VERSION)#{{end}}#
+LDFLAGS_EXTRAS=#{{- range .Config.ExtraLDFlags }}# #{{ . }}# #{{- end }}#
```
danielrbradley, line 33: Delete this line - it's only used for newrelic

```suggestion
```

provider-ci/internal/pkg/templates/parameterized-go/Makefile
```diff
@@ -0,0 +1,262 @@
...
+#{{- end }}#
+PROVIDER := pulumi-resource-$(PACK)
+TESTPARALLELISM := 10
+GOTESTARGS := ""
+WORKING_DIR := $(shell pwd)
+#{{- if .Config.GoBuildParallelism }}#
+PULUMI_PROVIDER_BUILD_PARALLELISM ?= -p #{{ .Config.GoBuildParallelism }}#
+#{{- else }}#
+PULUMI_PROVIDER_BUILD_PARALLELISM ?=
+#{{- end }}#
```
danielrbradley, line 19: I'd delete this whole option personally .. not required for smaller providers.

provider-ci/internal/pkg/templates/parameterized-go/Makefile
```diff
@@ -0,0 +1,262 @@
...
+			java -jar bin/jsign-6.0.jar \
+				--storetype AZUREKEYVAULT \
+				--keystore "PulumiCodeSigning" \
+				--url "${AZURE_SIGNING_KEY_VAULT_URI}" \
+				--storepass "$${ACCESS_TOKEN}" \
+				$@.unsigned; \
+			mv $@.unsigned $@; \
+			az logout; \
+		fi; \
+	fi
```
danielrbradley, line 223: This would be a nice candidate for extracting to a `.sh` file that both makefiles can call.

```suggestion
	scripts/sign-windows-binary.sh
```

provider-ci/internal/pkg/templates/parameterized-go/Makefile
```diff
@@ -0,0 +1,262 @@
...
+	@# $@ is the current target e.g. bin/pulumi-resource-xyz-v1.2.3-linux-amd64.tar.gz
+	tar --gzip -cf $@ README.md LICENSE -C $$(dirname $<) .
+
+provider_dist-linux-amd64: bin/$(PROVIDER)-v$(PROVIDER_VERSION)-linux-amd64.tar.gz
+provider_dist-linux-arm64: bin/$(PROVIDER)-v$(PROVIDER_VERSION)-linux-arm64.tar.gz
+provider_dist-darwin-amd64: bin/$(PROVIDER)-v$(PROVIDER_VERSION)-darwin-amd64.tar.gz
+provider_dist-darwin-arm64: bin/$(PROVIDER)-v$(PROVIDER_VERSION)-darwin-arm64.tar.gz
+provider_dist-windows-amd64: bin/$(PROVIDER)-v$(PROVIDER_VERSION)-windows-amd64.tar.gz
+provider_dist: provider_dist-linux-amd64 provider_dist-linux-arm64 provider_dist-darwin-amd64 provider_dist-darwin-arm64 provider_dist-windows-amd64
+.PHONY: provider_dist-linux-amd64 provider_dist-linux-arm64 provider_dist-darwin-amd64 provider_dist-darwin-arm64 provider_dist-windows-amd64 provider_dist
```
danielrbradley, line 252: This would be nicely grouped into an include for both makefiles:

```suggestion
include scripts/go-provider-dist.mk
```

danielrbradley (commented):
Good approach. Added some initial thoughts on minimising the duplication.

provider-ci/internal/pkg/templates.go
```diff
@@ -32,13 +37,15 @@ func getTemplateDirs(templateName string) ([]TemplateDir, error) {
 	switch templateName {
 	case "bridged-provider":
 		// Any Pulumi-owned bridged provider
-		return []TemplateDir{base, internal, bridged, internalBridged}, nil
+		return []TemplateDir{base, bridgedMakefile, internal, bridged, internalBridged}, nil
 	case "external-bridged-provider":
 		// third-party bridged providers
-		return []TemplateDir{base, external, bridged}, nil
+		return []TemplateDir{base, bridgedMakefile, external, bridged}, nil
```
danielrbradley, line 43: Technically you can implement this without moving the makefile into its own template as templates are rendered in order. So, as long as you put `parameterizedGo` after `base`, you just just override the Makefile.

t0yv0: Ugh really? Inheritance :) We'll curse this one day. I'll bear in mind though.

provider-ci/internal/pkg/templates/base/.github/workflows/prerequisites.yml
```diff
@@ -51,6 +51,7 @@ jobs:
         major-version: #{{ .Config.MajorVersion }}#
         set-env: 'PROVIDER_VERSION'
     - name: Cache examples generation
+      if: env.PULUMI_BUILD_SCHEMA != 'disabled'
```
danielrbradley, line 54: Please don't pepper the code with magic env 😬 

Adding a new boolean field in the config would be preferable, though I wonder if you can do this without any conditionals ...

Is the problem you're trying to solve for that you don't have a schema-embed file? If so, perhaps we just make it more robust so it doesn't fail if it's not there? This can be most simply done by making the artifact upload/download use a wildcard which can then match 0 files.

t0yv0: Hah I thought Bryce's direction was to remove configs? 

Schema-embed functionality is not needed for schema-less providers. We have to disable it. 

> This can be most simply done by making the artifact upload/download use a wildcard which can then match 0 files.

I do not agree with this in particular, let me explain briefly - if action A is redundant for scenario S1 but needed for scenario S2, weakening A failure checks to accommodate S2 would make S1 more brittle. This is not good in my book! It's just us being lazy and not segregating the logic so A only runs for S1 and not S2. 

Regardless on which method of factoring/deduplication we use I'd like to avoid running no-op schema-embed stuff if we can. Sounds like we can.

danielrbradley: Bryce's work was to statically type all config so we know exactly how providers can vary the output. See the `.Config.MajorVersion` above - this is described in code now:
https://github.com/pulumi/ci-mgmt/blob/bebe923d5312b1d1e63075152151c7d964c49db5/provider-ci/internal/pkg/config.go#L34

In this instance I'm not worried about weaking the failure check as Go will fail to compile without the embedded file.

Ok, another option to solve this is to remove the special logic for the schema-embed entirely and say we only capture and restore the `bin` directory. This is very general pattern which is easy to understand compared with specific files being handled in special ways. This would require us to modify the order of operations a little to accomodate the workflow:
1. Change `generate.to` in bridged providers to output the `schema-embed.json` to the bin directory (via gopatch migration) and do likewise with other providers. Or, even better, make `tfgen` output this straight to the `bin` and delete the `generate.go` files in every provider.
2. In the `makefile` `build_provider_cmd`, copy the `schema-embed.json` from `bin` to the cmd directory before calling `go build`.

This is quite a bit more involved than just doing the wildcard though.

t0yv0: I've added a NoConfig and this seems to be working out - working through end to end tests.

I've been heavily burned by schema-embed.json so I'm trying to avoid touching that in this PR. Let's tackle it separately? Indeed the are some bridge-specific assumptions there and the interface is not clean. The ideas you're suggesting are sound - we need a way to transport artifacts across phases, and bridged providers need to count on the mechanism of transporting schema-embed.json. That's a good goal. But getting there is tricky due to legacy.

t0yv0:
Got this to a good-enough place to run a release on the new provider.

t0yv0:
Let me address comments now.

provider-ci/internal/pkg/templates/parameterized-go/Makefile
t0yv0: Not quite, build_provider_cmd also does cross-builds in the same Makefile. I'm tempted to leave it, it works.

provider-ci/internal/pkg/templates/parameterized-go/Makefile
t0yv0: They're similar but different sets of targets. I'm not keen on extracting this just yet.

provider-ci/internal/pkg/templates/parameterized-go/Makefile
t0yv0: Doesn't hurt to have it, we assume these are smaller but what's the harm.

provider-ci/internal/pkg/templates/parameterized-go/Makefile
t0yv0: Yeah that's nice. All the cross-build code is identical. I've shared it as you suggest, that seems to still work. There's only one copy of it in ci-mgmt. We can edit it from here to make it better, as in abstracting singing script.

provider-ci/internal/pkg/templates/parameterized-go/Makefile
t0yv0: Nice but perhaps follow-up PR, not here.

provider-ci/internal/pkg/templates/parameterized-go/Makefile
t0yv0: This one ever so slightly different because go.mod location is different. It is a parameter though so perhaps I can add Go template expanders to customize it. I will try.

provider-ci/test-providers/acme/.golangci.yml
```diff
@@ -19,6 +19,8 @@ linters:
   - unused
   enable-all: false
 issues:
+  exclude-dirs:
```
t0yv0, line 22: Really handy to avoid linting 3-rd party code we include from OpenTOFU for example.

provider-ci/test-providers/acme/Makefile
```diff
@@ -305,86 +305,7 @@ debug_tfgen:
 	dlv  --listen=:2345 --headless=true --api-version=2  exec $(WORKING_DIR)/bin/$(CODEGEN) -- schema --out provider/cmd/$(PROVIDER)
 .PHONY: debug_tfgen
 
-# Provider cross-platform build & packaging
```
t0yv0, line 308: Moved as-is byte for byte into a shared file.

provider-ci/test-providers/aws/.github/workflows/run-acceptance-tests.yml
```diff
@@ -42,6 +42,7 @@ jobs:
     if: github.event_name == 'repository_dispatch' ||
       github.event.pull_request.head.repo.full_name == github.repository
     permissions:
+      contents: read
```
t0yv0, line 45: Without this non-public repos do not work.

provider-ci/internal/pkg/templates/parameterized-go/Makefile
t0yv0: Yep I think this code can be shared as well.

provider-ci/internal/pkg/templates.go
t0yv0: Used it to success.

provider-ci/internal/pkg/templates/base/.github/workflows/prerequisites.yml
t0yv0: Static configs. I think it might be OK for now.

provider-ci/internal/pkg/config.go
```diff
@@ -315,6 +315,14 @@ type Config struct {
 	// AllowMissingDocs controls whether we allow missing docs in the provider.
 	// Defaults to false.
 	AllowMissingDocs bool `yaml:"allowMissingDocs"`
+
+	// NoSchema is useful for providers such as parameterized providers that do not check in
```
guineveresaenger, line 319:
```suggestion
	// NoSchema is set by providers such as parameterized providers that do not check in
```

It seems like setting this is a necessity not just useful.

guineveresaenger (commented):
We currently have exactly one parameterized provider that we are hooking up. 

Do we have plans for more? Or to break out the `dynamic` features from the bridge?

t0yv0:
We have any TF one that has non-standard CI.

Sure we'll have tons more.

danielrbradley (approved):
This looks good to go 👍 

We can do a few follow-up simplifications at some point, but happy this isn't too repeated and gets you unblocked.

PR #1386, opened 2025-02-14
make provider target is now PHONY

t0yv0:


guineveresaenger (approved):
This works for me.

Please amend the title/description to mention this change is limited to the TF module so we can find it more easily later.

rquitales (approved):

danielrbradley (changes requested):
If the build doesn't produce a single file artifact, then it should use a sentinal target instead `.make/provider`.

provider-ci/internal/pkg/templates/parameterized-go/Makefile
```diff
@@ -102,11 +102,9 @@ build_provider_cmd = #{{ if .Config.BuildProviderPre -}}#
                      GOOS=$(1) GOARCH=$(2) CGO_ENABLED=0 go build $(PULUMI_PROVIDER_BUILD_PARALLELISM) -o "$(3)" -ldflags "$(LDFLAGS)" $(GO_MODULE)/cmd/$(PROVIDER)
 #{{- end }}#
 
+provider:
+	@$(call build_provider_cmd,$(shell go env GOOS),$(shell go env GOARCH),$(WORKING_DIR)/bin/$(PROVIDER))
 .PHONY: provider
```
danielrbradley, line 107:
```suggestion
provider: .make/provider
.PHONY: provider

.make/provider:
	$(call build_provider_cmd,$(shell go env GOOS),$(shell go env GOARCH),$(WORKING_DIR)/bin/$(PROVIDER))
```

t0yv0: No. That's precisely what we do not want here, because Go toolchain is much better at tracking dependencies than Make. This would defeat the purpose of this change, which is avoiding `make provider` accidentally forgetting to rebuild the provider in dev when sources change.

t0yv0:
> but anything that ends up depending on this in the future will also be rebuilt every time.

I would suggest - let's deal with this problem when we have it.

provider-ci/internal/pkg/templates/parameterized-go/Makefile
danielrbradley: Oh, right so you just need to add your source dependencies to the file target then ...

Either just add the files yourself or use Ian's tool to identify the files from the go sourcecode.

```diff
- bin/$(PROVIDER):
- 	$(call build_provider_cmd,$(shell go env GOOS),$(shell go env GOARCH),$(WORKING_DIR)/bin/$(PROVIDER))
+ bin/$(PROVIDER): $(shell find provider -type f)
+ 	$(call build_provider_cmd,$(shell go env GOOS),$(shell go env GOARCH),$(WORKING_DIR)/bin/$(PROVIDER))
```

Or

```diff
+ bin/$(PROVIDER): $(shell ${HELPMAKEGO} ./cmd/my_tool)
+ 	$(call build_provider_cmd,$(shell go env GOOS),$(shell go env GOARCH),$(WORKING_DIR)/bin/$(PROVIDER))
```

danielrbradley: Also, interesting that this is somewhat broken for the bridged providers right now too - we should add the missing dependencies there also.

danielrbradley (commented):
Unblocking by moving to a comment as I'm going offline, but I'd like the see the minimal fix of the `find` fix before this goes in to avoid us slipping into a world where makefiles are just task runners and instead focus on fixing the missing dependencies when we find them.

danielrbradley (commented):
Unblocking by moving to a comment as I'm going offline, but I'd like the see the minimal fix of the `find` fix before this goes in to avoid us slipping into a world where makefiles are just task runners and instead focus on fixing the missing dependencies when we find them.

danielrbradley (approved):
Moving it to an approve to unblock as GH doesn't let me just clear the request changes block.

provider-ci/internal/pkg/templates/parameterized-go/Makefile
t0yv0: Indeed. There's a lot riding on that "just". I'm not a fan of remaking Go caching in Make. Go knows better how to take care of things. It knows whether GOWORK changed, if sources changed or go.mod changed.

>  Ian's tool to identify the files from the go sourcecode

This seems ill-considered. It feels like rebuilding the Go toolchain and taking on maintaining that. It's not our core mission.

Trying to integrate Go cache hit/misses with Make seems possible in theory but I do not feel inspired to go that route. You could look at whether `go build` was a cache miss or a hit. You could teach make that. Just feels like hassle. 

It would be much easier to side-step the problem. If you have a developer-facing target, `make provider`, have it be PHONY. Go takes care of it.

t0yv0:
For completeness, if we were to teach Makefile how Go works, this *may* work:

```
need-provider: $(shell ./build-provider.sh)
	echo "Running code that needs provider.."
	touch "$@"
```

```
#!/usr/bin/env bash

set -euo pipefail

go build -o provider-binary-tmp

if cmp -s "provider-binary-tmp" "provider-binary"; then
    # "Files are identical, no copy needed.
    echo "^^ Provider up to date" >&2
    echo provider-binary
else
    echo "^^ Rebuilding" >&2
    cp provider-binary-tmp provider-binary
    echo provider-binary
fi
```

main.go:

```go

package main

import "fmt"

func main() {
	fmt.Println(5)
}

```

With this setup I can run `make need-provider` and it will rebuild after I change Go sources. 

I'm sure this breaks down somewhere though.

t0yv0:
I'd rather not merge this at all than do `$(shell find provider -type f)` - that is just something that pretends to do the right thing and then doesn't do it (e.g. you change your go.work and it ignores you). Discuss next week.

PR #1394, opened 2025-02-19
Bump upgrade-provider-action

VenelinMartinov:
Thomas spotted some warnings related to some recent changes to upgrade-provider-action: https://github.com/pulumi/pulumi-azuredevops/actions/runs/13382459741/job/37373285196

Looks like we never bumped the version of the action so the `allow-missing-docs` arg was not recognized.

Related to https://github.com/pulumi/ci-mgmt/pull/1379
Related to https://github.com/pulumi/upgrade-provider/issues/303


thomas11 (approved):
Any learnings on how we could catch such issues before rollout in the future?

Issue #1031, opened 2024-07-11
Make cache restore more targeted


danielrbradley:
Only specify caches when we're going to use the cache.

E.g. Only restore the go cache for `provider/go.mod` if we're going to build the provider.

This should help to reduce disk space usage issues.

PR #1406, opened 2025-02-24
Split up Go provider build caches by platform-arch

t0yv0:
In large providers like AWS we have cache stability issues - we run out of 10GB cache space, and we also have jobs failing when they hit out-of-disk issues on the runners, such as build_provider. One problem is that Go caches are very monolithic with one big file combining all architectures and platforms (in build_provider) job.

This PR looks at breaking down the caches per platform and architecture. Smaller caches should use less disk space on the runner. They also should evict more easily.

Relates https://github.com/pulumi/ci-mgmt/issues/1031

### DIFF
--- a/provider-ci/internal/pkg/templates/base/.github/workflows/build_provider.yml
+++ b/provider-ci/internal/pkg/templates/base/.github/workflows/build_provider.yml
@@ -58,6 +58,7 @@ jobs:
         uses: ./.github/actions/setup-tools
         with:
           tools: pulumictl, go
+          # use per-platform/arch caches instead since we are doing cross-builds
           cache-go: false
       # Based on https://github.com/actions/cache/blob/main/examples.md#go---modules
       - name: Get GOCACHE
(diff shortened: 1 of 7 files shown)


provider-ci/internal/pkg/templates/base/.github/workflows/build_provider.yml
```diff
@@ -58,6 +58,27 @@ jobs:
         uses: ./.github/actions/setup-tools
         with:
           tools: pulumictl, go
+          cache-go: false
```
guineveresaenger, line 61:
```suggestion
          cache-go: false # use per-platform/arch caches instead
```

just a quick comment to unconfuse a casual reader

guineveresaenger (approved):
This looks promising; have we tried it?

t0yv0:
Yeah trying it in pulumi-aws.

t0yv0:
It does build once. Remains to be seen if it operates sustainably in the steady state.

provider-ci/internal/pkg/templates/base/.github/workflows/build_provider.yml
t0yv0: Yep

rquitales (approved):
LGTM. Also noting that we have the option to use beefier runners if we keep hitting the 15 minute penalty for cache misses.

PR #1410, opened 2025-02-26
Allow testing of 3rd party contributions

### LINKED ISSUE #1409: Can't run acceptance tests for 3rd party PRs
A comment is added to third party PRs (with a different origin repo) instructing a maintainer to add the comment `/run-acceptance-tests`. This does run a workflow via command dispatch in the background but has two issues:
1. It doesn't check out the correct code (it runs on the default branch).
2. It doesn't report back success to the PR to unblock merging.

(2) has _never_ worked as this is a limitation of the command dispatch via repository_dispatch triggers. Repository dispatch always runs against the default branch and has no option to run elsewhere. Therefore the job statuses will never report back to the correct PR or commit automatically.

(1) used to work, but it appears got broken when we moved away from each job being defined in a single file and moved to re-usable workflows (around June 2024 https://github.com/pulumi/ci-mgmt/pull/942). This used to work by having a special case when checking out the code to specify the commit hash via an env variable which was populated when in repository_dispatch mode from the triggering payload.

# Possible solutions

## 1. Pass the SHA explicitly into all workflows

The most direct fix for (1) is to avoid ever using the `github` context for the current SHA. Most actions we use have the option for the SHA to be specified manually (but fallback to the default github context). 

We would then have a switch in the `run-acceptance-tests` workflow to check if we're in `repository_dispatch` mode and pull the SHA from the payload as we used to do.

Interestingly, this will also now fix (2) because we've since moved to the explicit sentinel status action on which we can specify the SHA, and this is the only check required to merge the PR.

### Drawbacks

This is tricky to prove that we've added overrides to all the correct places that require the current SHA. It's also possible that some actions are hard coded to use the SHA from the context and can't be overridden (including our versioning action, though we can fix this). Furthermore, this is very prone to breaking again in the future as the workflows evolve and people forget to specify the SHA from the input explicitly on some steps which could result in unexpected failures.

## 2. Change command dispatch to use workflow_dispatch

Workflow dispatch is the same mechanism used when running a job manually from the actions tab. This has the added features of having a `ref` parameter which allows the job to be run for a specific branch or commit. This would naturally operate on the correct SHA and report back all steps progress to the PR, avoiding manually overriding the SHA in every step.

### Challenges

Command dispatch has since also be leveraged for the `/release` command. It appears it's not possible to specify the type of dispatch per command so we'd need to also rework the release workflow which is heavily reliant on the repository dispatch payload which is not available when using workflow dispatch.

> A side question: do we still need the `/release` command? Is this a workflow we still encourage or do we all use the label based release process now?

It's possible we might be able to have two separate command dispatch handlers - one for each target command, but I'm not yet sure on if they would error if an unrecognised command was attempted.

Furthermore, this whole command dispatch process isn't currently implemented for 3rd party providers as it requires personal access tokens to be configured and therefore we need a process that works for them also.

## 3. Remove command dispatch and update instructions

It would be trivial to update the run-acceptance-tests workflow to allow workflow_dispatch. This allows it to be triggered in the GitHub actions UI. We could then update the instructions comment to link the maintainer to where to click the run button rather than implementing a whole complicated command dispatch mechanism.

When we originally wrote the command dispatch mechanism there was not a clickable UI for running actions directly which therefore mandated this complex setup, but perhaps it's a worthwhile simplification given the issues raised above.

danielrbradley:
Fixes #1409 

1. Allow the run-acceptance-tests workflow to be run manually.
2. Instruct the user to run the workflow via the UI.

See issue for discusson of alternative approaches and their drawbacks.

### DIFF
--- a/provider-ci/internal/pkg/templates/base/.github/workflows/pull-request.yml
+++ b/provider-ci/internal/pkg/templates/base/.github/workflows/pull-request.yml
@@ -8,21 +8,16 @@ jobs:
     name: comment-on-pr
     runs-on: #{{ .Config.Runner.Default }}#
     steps:
-    - name: Checkout Repo
-      uses: #{{ .Config.ActionVersions.Checkout }}#
-      with:
-        #{{- if .Config.CheckoutSubmodules }}#
-        submodules: #{{ .Config.CheckoutSubmodules }}#
-        #{{- end }}#
-        persist-credentials: false
-    - name: Comment PR
-      uses: #{{ .Config.ActionVersions.PrComment }}#
-      with:
-        github-token: ${{ secrets.GITHUB_TOKEN }}
-        message: >
-          PR is now waiting for a maintainer to run the acceptance tests.
+      - name: Comment PR
+        uses: #{{ .Config.ActionVersions.PrComment }}#
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          message: >
+            PR is now waiting for a maintainer to run the acceptance tests.
 
-          **Note for the maintainer:** To run the acceptance tests, please comment */run-acceptance-tests* on the PR
+            **Note for the maintainer:** To run the acceptance tests, go to the 
+            [run-acceptance-tests workflow in the Actions tab](${{ github.server_url }}/${{ github.repository }}/actions/workflows/run-acceptance-tests.yml)
+            and click the "Run workflow" button. Make sure to select the `${{ github.head_ref }}` branch.
 name: pull-request
 on:
   pull_request_target: {}
--- a/provider-ci/internal/pkg/templates/base/.github/workflows/pull-request.yml
+++ b/provider-ci/internal/pkg/templates/base/.github/workflows/pull-request.yml
@@ -18,6 +18,7 @@ jobs:
             **Note for the maintainer:** To run the acceptance tests, go to the 
             [run-acceptance-tests workflow in the Actions tab](${{ github.server_url }}/${{ github.repository }}/actions/workflows/run-acceptance-tests.yml)
             and click the "Run workflow" button. Make sure to select the `${{ github.head_ref }}` branch.
+            Alternatively, run `gh workflow run acceptance-test.yml --repo "${{ github.repository }}" --ref "${{ github.head_ref }}" on your local machine.`
 name: pull-request
 on:
   pull_request_target: {}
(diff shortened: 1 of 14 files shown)


danielrbradley:
To test this, we can merge the PR to xyz and open a PR from a fork.

provider-ci/internal/pkg/templates/base/.github/workflows/pull-request.yml
```diff
@@ -8,21 +8,16 @@ jobs:
...
-          PR is now waiting for a maintainer to run the acceptance tests.
+      - name: Comment PR
+        uses: #{{ .Config.ActionVersions.PrComment }}#
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          message: >
+            PR is now waiting for a maintainer to run the acceptance tests.
 
-          **Note for the maintainer:** To run the acceptance tests, please comment */run-acceptance-tests* on the PR
+            **Note for the maintainer:** To run the acceptance tests, go to the
```
thomas11, line 18: Another simple way of running it is `gh workflow run acceptance-test.yml --ref ${{ github.head_ref }}`.

thomas11 (approved):
LGTM. Option 3 is way better than what we have now. Thank you!

VenelinMartinov (approved):

danielrbradley:
Alas, this doesn't work because you can't dispatch a workflow for a ref that's not in the local repository. Finding another route forward instead.

Attempting to dispatch the workflow for the fork contribution in https://github.com/pulumi/pulumi-xyz/pull/906

```
$ gh workflow run run-acceptance-tests.yml --repo "pulumi/pulumi-xyz" --ref "danielrbradley:danielrbradley-patch-1" 
could not create workflow dispatch event: HTTP 422: No ref found for: danielrbradley:danielrbradley-patch-1 (https://api.github.com/repos/pulumi/pulumi-xyz/actions/workflows/90425114/dispatches)
$ gh workflow run run-acceptance-tests.yml --repo "pulumi/pulumi-xyz" --ref "danielrbradley-patch-1" 
could not create workflow dispatch event: HTTP 422: No ref found for: danielrbradley-patch-1 (https://api.github.com/repos/pulumi/pulumi-xyz/actions/workflows/90425114/dispatches)
$ gh workflow run run-acceptance-tests.yml --repo "pulumi/pulumi-xyz" --ref "7c493dab661dbe6a436a0ec8c50c6cd33cd0b2c0"
could not create workflow dispatch event: HTTP 422: No ref found for: 7c493dab661dbe6a436a0ec8c50c6cd33cd0b2c0 (https://api.github.com/repos/pulumi/pulumi-xyz/actions/workflows/90425114/dispatches)
```

PR #1412, opened 2025-02-26
Do not assume sed

t0yv0:


t0yv0:
First attempt failed as the relpaths were not resolved against outdir. Trying again.

t0yv0:
Looks like you solved this already! closing.

Issue #1141, opened 2024-11-12
upgrade-provider automation mistakenly mixes bridge and upstream upgrades


t0yv0:
Noticed this in https://github.com/pulumi/pulumi-aws/pull/4752

https://github.com/pulumi/ci-mgmt/blob/master/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/upgrade-provider.yml#L83

The code:

```
        run: upgrade-provider "${{ github.repository }}" --kind="all" --target-version="${{ steps.target_version.outputs.version }}" #{{ if .Config.javaGenVersion }}#--java-version="#{{ .Config.javaGenVersion }}#"#{{ end }}#
```

The `--kind="all"` flag is mixing bridge and upstream upgrades in a single change, which makes it harder to understand where changes are coming from and harder to bisect issues. Would it be possible to scope this to only upstream upgrades? And have a separate workflow handle upgrading the bridge

PR #1413, opened 2025-02-26
Upgrade provider should be upgrading upstream only

t0yv0:
Fixes https://github.com/pulumi/ci-mgmt/issues/1141

As a concrete example, https://github.com/pulumi/pulumi-gcp/pull/3004 fails because it starts hitting .NET 8.0 issues. This brings us close to violating SLA on upstream upgrades for no good reason.  I do believe we are better off with separate PRs for upstream upgrades vs Pulumi internal dependencies.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged/.github/workflows/upgrade-provider.yml
+++ b/provider-ci/internal/pkg/templates/bridged/.github/workflows/upgrade-provider.yml
@@ -86,7 +86,7 @@ jobs:
         continue-on-error: true
         uses: #{{ .Config.ActionVersions.UpgradeProviderAction }}#
         with:
-          kind: all
+          kind: upstream
           email: bot@pulumi.com
           username: pulumi-bot
           automerge: #{{ .Config.AutoMergeProviderUpgrades }}#
--- a/provider-ci/internal/pkg/templates/bridged/.github/workflows/upgrade-provider.yml
+++ b/provider-ci/internal/pkg/templates/bridged/.github/workflows/upgrade-provider.yml
@@ -86,7 +86,7 @@ jobs:
         continue-on-error: true
         uses: #{{ .Config.ActionVersions.UpgradeProviderAction }}#
         with:
-          kind: upstream
+          kind: provider
           email: bot@pulumi.com
           username: pulumi-bot
           automerge: #{{ .Config.AutoMergeProviderUpgrades }}#
--- a/provider-ci/test-providers/acme/.github/workflows/upgrade-provider.yml
+++ b/provider-ci/test-providers/acme/.github/workflows/upgrade-provider.yml
@@ -73,7 +73,7 @@ jobs:
         continue-on-error: true
         uses: pulumi/pulumi-upgrade-provider-action@ff5cb5907aecba099e61146c4d4d074c7fd6ca99 # v0.0.15
         with:
-          kind: upstream
+          kind: provider
           email: bot@pulumi.com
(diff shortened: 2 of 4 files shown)


t0yv0:
Hmm, I think it should be `"provider"` not `"upstream"`.

t0yv0:
Failure caused by .NET 8 issue I think unrelated to the change in question. This should clear up tomorrow AFAIK.

t0yv0:
Hm yeah it sounds like Eron remediated the failure separately. That's good. 

However I still think this is needed.

Why would we *ever* run with `all`?  It feels like an anti-pattern that confuses changes from different dependencies in one upgrade.

t0yv0:
> The failure in GCP is caused by the fact we don't pin the Pulumi CLI version in CI.

That is by the way factually inaccurate as GCP pins Pulumi CLI. However there are some issues with the reference to .NET Core SDK being unpinned that caused the failure.

t0yv0:
I'm not sure what all does in fact. From the docs it appears that:

```
- "all":     Upgrade the upstream provider and the bridge. Shorthand for "bridge,provider,code,pf".
```

We already run bridge upgrades when they are tested, separately. 

I doubt these are relevant anymore:

```
  - "pf": Upgrade the Plugin Framework only.
  - "code":     Perform some number of code migrations.
```

This might stop upgrading Java perhaps, but we should probably want to combine that with bridge releases and testing. What else? WDYT?

t0yv0:
Looks like upgrade-provider --kind=provider upgrades Java, which is a bug. GAH.

VenelinMartinov:
>That is by the way factually inaccurate as GCP pins Pulumi CLI.

```
Run pulumi/actions@13b8b7177d6fb736766875dac9b78aab07bd785f
Configured range: dev
/usr/local/bin/pulumi version
v3.150.0
warning: A new version of Pulumi is available. To upgrade from version '3.150.0' to '3.153.0', visit https://pulumi.com/docs/install/ for manual instructions and release notes.
Pulumi 3.150.0 does not satisfy the desired version dev. Proceeding to download
Matched version: v3.153.0-alpha.x252e880
```

Tests are run against the newest dev version. This might not have been the source of the dotnet issues however.

VenelinMartinov:
>This might stop upgrading Java perhaps, but we should probably want to combine that with bridge releases and testing. What else? WDYT?

Yeah, that sound sensible.

t0yv0:
Makefile ignores that version and downloads a pinned version to run with https://github.com/pulumi/pulumi-gcp/blob/master/scripts/plugins.mk#L31  AFAIK.

VenelinMartinov:
We don't use make for running tests... https://github.com/pulumi/pulumi-gcp/blob/4c7f9008368392fefb777bdf86f79c8eec8f3d2a/.github/workflows/test.yml#L98

t0yv0:
That's a .. good find? That's an interesting situation though. So we use one version of Pulumi CLI to build the schema, and another version (latest) to run tests against. I doubt this was though through but it possibly matches our current preference..

t0yv0:
> The failure in GCP is caused by the fact we don't pin the Pulumi CLI version in CI.

I wonder just out of curiosity if this is true or false. I think still it's false because even if we used unpinned CLI for testing, the fact that we pin CLI for schema and code generation means that nothing much was affected .NET-side. The new CLI binary is probably fine. I guess this was something else.

VenelinMartinov (approved):

PR #1408, opened 2025-02-26
Update dotnet to 8.0 LTS

EronWright:
This PR updates provider-ci and native-provider-ci to use .NET 8 (LTS) at the tool version and as the target framework for examples. See individual commits for breakdown.

Part of #1307 

Open questions:
1. the migration script looks to `examples/` and to `tests/` for .csproj files; should it look elsewhere? 
1. unrelated: why go 1.21?

Provider Test PRs:
- https://github.com/pulumi/pulumi-xyz/pull/910
- https://github.com/pulumi/pulumi-gcp/pull/3012

PR #1414, opened 2025-02-26
Upgrade to dotnet 8

danielrbradley:
.NET 6 is now out of support and the Pulumi .NET SDK is needing to move to .NET 8. Therefore we should use version 8 when building too.


danielrbradley:
Dupe of https://github.com/pulumi/ci-mgmt/pull/1408

EronWright:
A dup of: https://github.com/pulumi/ci-mgmt/pull/1408

PR #1418, opened 2025-02-27
Carry env vars into upgrade-bridge.yml

t0yv0:
This came up during investigating bridge downstream checks and in particular:

https://github.com/pulumi/pulumi-mongodbatlas/pull/714

There is an environment variable that affects whether preview resources are included in the schema or not:

https://github.com/pulumi/pulumi-mongodbatlas/blob/1a78223a600a451687b4a0b6ae76a6e21db06a71/.ci-mgmt.yaml#L7

This environment variable is set in all workflows but not in upgrade-bridge.yml. Because it is not set there, the wrong schema/sdk is generated during bridge upgrade. 

This PR suggests to set env vars uniformly. 

Alternatively we may consider wiring this env var into the Go code in mongodbatlas.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged/.github/workflows/upgrade-bridge.yml
+++ b/provider-ci/internal/pkg/templates/bridged/.github/workflows/upgrade-bridge.yml
@@ -57,8 +57,8 @@ permissions:
 
 env:
   GH_TOKEN: ${{ secrets.PULUMI_PROVIDER_AUTOMATION_TOKEN || secrets.PULUMI_BOT_TOKEN || secrets.GITHUB_TOKEN }}
-  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        
+#{{ .Config.Env | toYaml | indent 2 }}#
+
 jobs:
   upgrade_provider:
     name: upgrade-provider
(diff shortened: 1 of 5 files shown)


guineveresaenger (approved):
Feels a little overkill/messy to have _all_ of these env vars everywhere, but that's our MO in other workflows as well. 

I think I'd prefer this over putting it in the provider code, since that would add yet another place to look? But it's really six of one, half a dozen of the other.

rquitales (approved):

t0yv0:
Checking with Ramon what other work-flows do not set .Config.Env

```
 ./base/.github/actions/download-prerequisites/action.yml
./base/.github/actions/download-provider/action.yml
./base/.github/actions/download-sdk/action.yml
./base/.github/actions/setup-tools/action.yml
./base/.github/actions/upload-prerequisites/action.yml
./base/.github/actions/upload-sdk/action.yml
./base/.github/workflows/build_provider.yml      (????)
./base/.github/workflows/main-post-build.yml
./bridged/.github/workflows/upgrade-provider.yml (????)
./external/.github/workflows/resync-build.yml    (????)
./internal/.github/workflows/community-moderation.yml
./internal/.github/workflows/release_command.yml
```

t0yv0:
HTTP 403: API rate limit exceeded for user ID 30351955 :/

PR #1451, opened 2025-03-21
Revert "Optimize run-acceptance-tests: only build linux binaries"

t0yv0:
Reverts pulumi/ci-mgmt#1430

@ringods requested verifying that code works on Windows PR builds.


ringods:
@t0yv0 I'm not asking to revert everything, only adding a Windows build as this can be different regarding Go imports.

t0yv0:
That's more work I cannot undertake now. This change was an optimization so if we lose features easiest is to just revert. PRs are welcome for anything better than this :)

PR #1452, opened 2025-03-21
Adding Windows build to verify whether platform specific Go packages are used.

ringods:
Improves: #1430 
Supersedes: #1451


t0yv0 (approved):
I think it should be fine.

PR #1453, opened 2025-03-21
Switch AWS credentials to esc for native providers

corymhall:
This switches the AWS Access from hardcoded access keys that are pulled from GitHub secrets to use Pulumi ESC environments.

re #3731


native-provider-ci/src/workflows.ts
```diff
@@ -511,9 +511,23 @@ export class PrerequisitesJob implements NormalJob {
...
+    const awsCredentialSteps = steps.ConfigureAwsCredentialsForTests(
+      opts.provider == "aws-native"
+    );
+    if (awsCredentialSteps.length > 0) {
+      // If you specify _any_ permissions then by default the rest of the permissions
+      // that you don't specify will be set to "none"
+      this.permissions = {
+        // Pulumi esc auth requires 'id-token' permission
+        "id-token": "write",
+        // commont-on-pr requires 'pull-requests' permission
```
mjeffryes, line 527:
```suggestion
        // comment-on-pr requires 'pull-requests' permission
```

Was this just something we overlooked in a previous PR so the commenting on the PR was just broken, or is this a new requirement of the ESC auth stuff?

mjeffryes: Oh I think, this is sort of answered by the first comment in the block. `"pull-requests": "write"` is the default, but if we set any permission explicitly, we have to specify everything we need and not just rely on the default. Tricky!

mjeffryes (approved):

guineveresaenger (approved):
This looks good to me. What is the reason to hold off on this for bridged providers? Is the intent to use native providers as a canary of sorts?

corymhall:
> This looks good to me. What is the reason to hold off on this for bridged providers? Is the intent to use native providers as a canary of sorts?

Just splitting things up into separate PRs

PR #1454, opened 2025-03-21
publish_sdk no longer consults the Go cache

t0yv0:
Motivated by:

https://github.com/pulumi/pulumi-aws/issues/5337
https://github.com/pulumi/pulumi-aws/issues/5122

I was only able to resolve 5337 that kept failing on the publish_sdk step by clearing caches and a targeted re-run.  The AWS caches are too large. In particular the default Go cache. I think publish_sdk is better off opting out of Go cache just like `publish` is opting out of Go cache as they are not doing any particularly heavy Go compilation, and doing this alleviates the problem of downloading large caches and running out of disk space.


guineveresaenger (approved):
Seems sensible.

rquitales (approved):

PR #1480, opened 2025-04-01
Move native providers to templates

mjeffryes:
We'd like to make CI for our native providers more consistent with the rest of our provider fleet. The first step is to start using the same template system for these providers. 

I've taken the rather tedious, but safe route of creating templates that produce almost identical results to the current native-provider-ci (modulo the ordering of a few env vars).  I've moved all the templated workflows for these providers from native-provider-ci/providers to provider-ci/test-providers to verify. This should make it easier review and simplifies adding a .ci-mgmt.yaml to each of these providers.

After this lands we can add a .ci-mgmt.yaml to each of these providers, add them to providers.json, and remove native-provider-ci. Then we'll be free to start migrating these templates to share more with other provider templates.

## Provider PRs
- https://github.com/pulumi/pulumi-command/pull/733
- https://github.com/pulumi/pulumi-aws-native/pull/2165
- https://github.com/pulumi/pulumi-docker-build/pull/511


mjeffryes:
Looks like removing the directories from native-provider-ci/providers caused us to drop them from our infra program. Will fix.

mjeffryes:
infra project is fixed, this is good to go

provider-ci/internal/pkg/templates.go
```diff
@@ -44,6 +48,10 @@ func getTemplateDirs(templateName string) ([]TemplateDir, error) {
 		return []TemplateDir{base, internal}, nil
 	case "parameterized-go":
 		return []TemplateDir{base, parameterizedGo /* overrides Makefile */}, nil
+	case "native":
+		return []TemplateDir{native}, nil // Can't use base because it has a Makefile that would conflict
+	case "aws-native":
+		return []TemplateDir{native, awsNative}, nil // AWS native has 2 extra workflows
```
VenelinMartinov, line 54: I believe providers are allowed to have custom workflows in their repos and ci-mgmt should respect that. You might be able to get rid of the aws-native specialization here and just move the workflows to the repo.

Here is an example from aws-classic: https://github.com/pulumi/pulumi-aws/blob/master/.github/workflows/aws-upstream-tests.yml

VenelinMartinov (approved):
LGTM. Would be great if we can remove the aws-native-specific template.

provider-ci/internal/pkg/templates.go
mjeffryes: Yes, For this change I wanted to just move everything that was in native-providers-ci over intact to make it easy to verify, but I agree that once we have aws-native fully switched over we can look at migrating it's custom workflows into the repo and removing this.

PR #1485, opened 2025-04-08
Drop wildcard to prevent accidental collisions

ringods:
I’m helping @scraly from OVH to adopt ci-mgmt. In their repositories, they have a top-level folder called LICENSES which is populated by a central process.

When rolling out the external-provider-template for the ovh/pulumi-ovh repository, the copy command in .make/build_nodejs includes ../../LICENSE* (with wildcard). This wildcard matches folder LICENSES but the copy command fails because it only expects files.

I checked the filenames of the LICENSE files in pulumi and pulumiverse organization and all seem to be an exact match with LICENSE without file extension.

https://github.com/search?q=org%3Apulumi%20LICENSE.txt&type=code

https://github.com/search?q=org%3Apulumi%20LICENSE.md&type=code

https://github.com/search?q=org%3Apulumiverse+LICENSE.txt&type=code&p=1

https://github.com/search?q=org%3Apulumiverse+LICENSE.md&type=code&p=1

So I think this is a safe change.

### DIFF
--- a/provider-ci/internal/pkg/templates/base/Makefile
+++ b/provider-ci/internal/pkg/templates/base/Makefile
@@ -164,7 +164,7 @@ build_nodejs: .make/build_nodejs
 	cd sdk/nodejs/ && \
 		yarn install && \
 		yarn run tsc && \
-		cp ../../README.md ../../LICENSE* package.json yarn.lock ./bin/
+		cp ../../README.md ../../LICENSE package.json yarn.lock ./bin/
 	@touch $@
 .PHONY: generate_nodejs build_nodejs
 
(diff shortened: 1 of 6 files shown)


iwahbe:
We can't know that the change is safe since 3rd parties may use `ci-mgmt`, and they could have a `LICENSE.md` file.

That said, I think this is *probably* fine to take. I have tagged in the team that owns the repo for a review.

danielrbradley:
I also _think_ this is safe based on not being able to find any remaining uses of "LICENSE.txt", but would be good for @pulumi/iac-ecosystem to sign off to ensure they're aware of the change going out.

danielrbradley (approved):
Approving to unblock as I think ecosystem are lacking bandwidth to take a look right now

PR #1491, opened 2025-04-09
Run integration tests for Pulumi YAML

### LINKED ISSUE #1487: [Native] Run integration tests for Pulumi YAML
We control integration tests by running `go test` with a language specific tag, for example `go test -tag python`. Since we also have a Pulumi YAML language, we should support writing integration tests with it.

iwahbe:
This PR adds Pulumi YAML as a language in the integration test matrix. Because Pulumi YAML doesn't have SDKs, we skip the `Download SDK`, `UnTar SDK folder` and `Install dependencies` steps for Pulumi YAML.

The first commit is the manual change. The second commit is the result of running `make build` to regenerate examples.

Fixes #1487

### DIFF
--- a/provider-ci/internal/pkg/templates/native/.github/workflows/build.yml
+++ b/provider-ci/internal/pkg/templates/native/.github/workflows/build.yml
@@ -411,6 +411,7 @@ jobs:
         - dotnet
         - go
         - java
+        - yaml
     name: test
     permissions:
       contents: read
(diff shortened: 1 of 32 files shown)


mjeffryes (approved):
Good to merge this as long as the sanity check in  https://github.com/pulumi/pulumi-docker-build/pull/517 looks good.

blampe:
Doesn't this break workflows for anything that doesn't have a `tests/yamls` directory? I assume that's most providers?

iwahbe:
> Doesn't this break workflows for anything that doesn't have a `tests/yaml` directory? I assume that's most providers?

No. Why would it? This will run any tests in the `examples` dir that is flagged behind the `yaml` build tag. Thats it.

blampe:
@iwahbe false alarm, seems p-k is the odd one out here. It expects the directory to exist, https://github.com/pulumi/pulumi-kubernetes/pull/3616 should take care of it.

    - name: Run tests
      run: cd tests/sdk/${{ matrix.language }} && go test -v -count=1 -cover -timeout

PR #1510, opened 2025-04-23
Remove native-provider-ci

mjeffryes:
All the providers that had been using native-provider-ci to manage their workflows have been moved to use a template in provider-ci instead. (google-native is currently unmanaged as it is on the road to deprecation.)

This means we can delete native-provider-ci and all the workflows & tools that were pointing to it so we only have one system to maintain going forward!


mjeffryes:
I think the pulumi update fails because it's trying to remove labels from the google-native provider repo (which are protected resources); we should probably just be using `retain_on_delete` for labels rather than protect though?

Issue #1514, opened 2025-04-28
Bridged providers no longer build ARM binaries for windows


mjeffryes:
As reported in community slack: https://pulumi-community.slack.com/archives/C84L4E3N1/p1745597638246989

The last pulumi-aws release to include Windows Arm binaries was https://github.com/pulumi/pulumi-aws/releases/tag/v6.44.0

It looks like this change https://github.com/pulumi/pulumi-aws/pull/4202 is where we stopped building them. I think the omission was accidental.

blampe:
Tracking here https://github.com/pulumi/ci-mgmt/issues/1513

PR #1515, opened 2025-04-28
Also build ARM binaries for windows

mjeffryes:
We seem to have accidentally dropped ARM builds for windows when we switched away from goreleaser. This change should start building them again. 

Fixes https://github.com/pulumi/ci-mgmt/issues/1514

### DIFF
--- a/provider-ci/internal/pkg/templates/base/.github/workflows/build_provider.yml
+++ b/provider-ci/internal/pkg/templates/base/.github/workflows/build_provider.yml
@@ -19,7 +19,8 @@ on:
               {"os": "linux",   "arch": "arm64"},
               {"os": "darwin",  "arch": "amd64"},
               {"os": "darwin",  "arch": "arm64"},
-              {"os": "windows", "arch": "amd64"}
+              {"os": "windows", "arch": "amd64"},
+              {"os": "windows", "arch": "arm64"}
             ]
           }
 


danielrbradley (commented):
This looks insufficient as it doesn't implement the build process in the makefile. 

Confusingly, it looks like Ringo add the windows-*amd* build in https://github.com/pulumi/ci-mgmt/pull/1452 during the acceptance testing phase, but your change only affects the _release_ builds, so wasn't actually tested until you attempted to merge. Here's the release testing failure: https://github.com/pulumi/pulumi-xyz/actions/runs/14719345106/job/41309976000

Here's the file where we need some new targets added: https://github.com/pulumi/ci-mgmt/blob/master/provider-ci/internal/pkg/templates/base/scripts/crossbuild.mk#L12-L21. We also need to establish if, for windows _arm_ builds, if we should be doing the same signing process as for windows-amd .. I expect we should, but I'm also not sure if `jsign` will support arm, so we should verify that too.

danielrbradley:
Gone ahead and implemented the makefile ..

main build without signing: https://github.com/pulumi/pulumi-xyz/actions/runs/14726403970
main build with signing: https://github.com/pulumi/pulumi-xyz/actions/runs/14726430432

danielrbradley:
As a related item, it would also be good to implement release verification for windows ARM by using the preview [windows-11-arm](https://github.com/actions/partner-runner-images/blob/main/images/arm-windows-11-image.md) public runner. This would mean that we can verify the actual binaries are usable in the xyz release before rolling this out to all providers.

mjeffryes:
Thanks for cleaning up after me @danielrbradley! 

> it would also be good to implement release verification for windows ARM by using the preview [windows-11-arm](https://github.com/actions/partner-runner-images/blob/main/images/arm-windows-11-image.md) public runner. This would mean that we can verify the actual binaries are usable in the xyz release before rolling this out to all providers. Also, we could add [ubuntu-24.04-arm](https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-24-image.md) for linux arm verification too.

I assume we'd need to edit the release verification action to add support for these first before we can turn it on in the provider workflow?

danielrbradley:
> I assume we'd need to edit the release verification action to add support for these first before we can turn it on in the provider workflow?

Don't think so ... probably just needs the ARM runner in the matrix here: https://github.com/pulumi/ci-mgmt/blob/master/provider-ci/internal/pkg/templates/base/.github/workflows/verify-release.yml#L58

t0yv0:
Ah good catch. This must have been why the issue happened, at some point when editing these files I must have assumed Windows-ARM was not supported because signing was not supported for this combination, and this assumption got enshrined in the code, even when the former goreleaser builds used to build for it.

danielrbradley:
I've just opened this to confirm if the release verification works on arm platforms .. if it doesn't, we should fix it! https://github.com/pulumi/verify-provider-release/pull/144

danielrbradley:
Oh .. this is an issue:

```
Error: Unsupported operating system - Pulumi CLI is only released for Darwin (x64, arm64), Linux (x64, arm64) and Windows (x64)
```

mjeffryes:
"Unsupported operating system - Pulumi CLI is only released for Darwin (x64, arm64), Linux (x64, arm64) and Windows (x64)" 🤦

danielrbradley:
I think is this wrong: https://github.com/pulumi/actions/blob/9519177da243fd32cab35cdbf19cce1ab7472fcc/src/libs/pulumi-cli.ts#L65

danielrbradley:
Opened: https://github.com/pulumi/actions/issues/1364

PR #1558, opened 2025-06-05
Use provider-boilerplate as a native canary

blampe:
It's not possible to reference outputs from a matrix job, so I combined these two jobs into one. Otherwise things are essentially the same, just parameterized on repo name. I also tightened up some timeouts.


.github/workflows/test-provider-ci.yml
```diff
@@ -119,33 +111,35 @@ jobs:
       - name: Await main build start
         id: main_build
         run: |
-          until (gh run list --repo "pulumi/pulumi-xyz" --workflow main --json headSha | grep -q "${{ steps.merge_commit.outputs.oid }}"); do sleep 1; done
-          database_id=$(gh run list --repo "pulumi/pulumi-xyz" --workflow main --json "number,headSha,databaseId" | jq '.[] | select(.headSha == "${{ steps.merge_commit.outputs.oid }}") | .databaseId')
+          until (gh run list --repo "${{ matrix.repo }}" --workflow main --json headSha | grep -q "${{ steps.merge_commit.outputs.oid }}"); do sleep 1; done
+          database_id=$(gh run list --repo "${{ matrix.repo }}" --workflow main --json "number,headSha,databaseId" | jq '.[] | select(.headSha == "${{ steps.merge_commit.outputs.oid }}") | .databaseId')
```
mjeffryes, line 115: does provider-boilerplate have a "main" workflow? I think it has "build" instead.

.github/workflows/test-provider-ci.yml
```diff
@@ -119,33 +111,35 @@ jobs:
...
-          until (git ls-remote --tags "https://github.com/pulumi/pulumi-xyz.git" | grep -q "${{ steps.merge_commit.outputs.oid }}"); do sleep 1; done
+          until (git ls-remote --tags "https://github.com/${{ matrix.repo }}.git" | grep -q "${{ steps.merge_commit.outputs.oid }}"); do sleep 1; done
           # Also handle annotated tags in the format refs/tags/v0.1.0^{}
-          tag=$(git ls-remote --tags "https://github.com/pulumi/pulumi-xyz.git" | grep "${{ steps.merge_commit.outputs.oid }}" | cut -d '/' -f 3 | sed -E 's/\^\{\}$//')
+          tag=$(git ls-remote --tags "https://github.com/${{ matrix.repo }}.git" | grep "${{ steps.merge_commit.outputs.oid }}" | cut -d '/' -f 3 | sed -E 's/\^\{\}$//')
           echo "Tag for release is ${tag}"
           echo "tag=${tag}" >> "${GITHUB_OUTPUT}"
 
-      - name: Wait for release workflow run
+      - name: Await release workflow run
```
mjeffryes, line 134: This probably wont trigger for the native provider because we don't have the release label machinery

mjeffryes (commented):
wow I didn't realize how intricate this sanity check is! I'm excited to add this as a guard rail for changes to native provider workflows, but I'm concerned that some of the checks probably won't translate today as native and bridged workflows behave a bit differently.

blampe:
@mjeffryes yeah that's all valid. I'd like to keep these as close as possible so I don't mind fixing the tagging while I'm in here. This won't be able to merge until they're at parity.

mjeffryes (approved):
Cool let's see if it works!

Issue #1481, opened 2025-04-03
Replace Github secrets use in providers CI with ESC


mjeffryes:
We want to remove all secrets from GitHub in favor of using Pulumi ESC. We can setup secret rotation, dynamic secrets, better auditng, and access controls.
# Providers

## Native Providers

- [x] https://github.com/pulumi/pulumi-aws-apigateway/pull/439
- [x] https://github.com/pulumi/pulumi-aws-native/pull/2449
- [x] https://github.com/pulumi/pulumi-awsx/pull/1691 (blocked on https://github.com/pulumi/pulumi-awsx/issues/1692)
- [x] https://github.com/pulumi/pulumi-azure-native/pull/4302 (not managed)
- [x] https://github.com/pulumi/pulumi-command/pull/914
- [x] https://github.com/pulumi/pulumi-docker-build/pull/594
- [x] https://github.com/pulumi/pulumi-eks/pull/1895
- [x] https://github.com/pulumi/pulumi-kubernetes/pull/3827
- [x] https://github.com/pulumi/pulumi-kubernetes-cert-manager/pull/617
- [x] https://github.com/pulumi/pulumi-kubernetes-ingress-nginx/pull/363
- [x] https://github.com/pulumi/pulumi-kubernetes-coredns/pull/378

### Tier 1

- [x] https://github.com/pulumi/pulumi-aws/pull/5829
- [x] https://github.com/pulumi/pulumi-gcp/pull/3383
- [x] https://github.com/pulumi/pulumi-azure/pull/3274
- [x] https://github.com/pulumi/pulumi-azuread/pull/2101

### Tier 2

- [x] https://github.com/pulumi/pulumi-auth0/pull/953
- [x] https://github.com/pulumi/pulumi-cloudflare/pull/1361
- [x] https://github.com/pulumi/pulumi-datadog/pull/939
- [x] https://github.com/pulumi/pulumi-digitalocean/pull/1138
- [x] https://github.com/pulumi/pulumi-docker/pull/1516
- [x] https://github.com/pulumi/pulumi-github/pull/993
- [x] https://github.com/pulumi/pulumi-mongodbatlas/pull/848
- [x] https://github.com/pulumi/pulumi-postgresql/pull/776
- [x] https://github.com/pulumi/pulumi-random/pull/1868
- [x] https://github.com/pulumi/pulumi-tls/pull/845

### Tier 3

- [x] https://github.com/pulumi/pulumi-aiven/pull/983
[...]

PR #1570, opened 2025-06-11
Add a provider workflow to export repo secrets

blampe:
This adds a new "Export secrets to ESC" workflow to our internal providers. The
workflow itself is copy-pasted from Pat's prior work. This workflow is
necessary in cases where the provider relies on repo-specific secrets. It may
need to be run at most once per provider, and we can remove this later when all
secrets have been migrated.


blampe:

- #1569
    - #1570 ◀

PR #1576, opened 2025-06-11
Mark aws-native and native templates as internal

blampe:
This marks `native` and `aws-native` templates as internal.

This has the following beneficial side effects:
* The native providers now get the same issue templates as bridged.
* Code of conduct is consolidated.
* `/run-acceptance-tests` is consolidated, e.g. native providers now get things
  like the "you're modifying SDK directly" warning. If there's any
  bridge-specific logic in here we can address it later.

This also carves out a new template `external-native-provider` in the spirit of
`external-bridged-provider`. We can move pulumi-provider-boilerplate over to
this and eventually add some exceptions for things like doc publishing
https://github.com/pulumi/registry/pull/7825.


blampe:

- #1576 ◀

mjeffryes (approved):
This is a nice consolidation

PR #1578, opened 2025-06-11
Add a provider workflow to export repo secrets

blampe:
This adds a new "Export secrets to ESC" workflow to our internal providers. The
workflow itself is copy-pasted from Pat's prior work. This workflow is
necessary in cases where the provider relies on repo-specific secrets. It may
need to be run at most once per provider, and we can remove this later when all
secrets have been migrated.

Confirmed working on docker-build:

```
❯ esc env get pulumi/github-secrets/pulumi-pulumi-docker-build

   Value

    {
      "environmentVariables": {
        ...
      }
    }

   Definition

    imports:
      - imports/github-secrets
    values:
      environmentVariables:
        CODECOV_TOKEN:
          ...
        DOCKER_HUB_PASSWORD:
          ...
```

### DIFF
--- a/provider-ci/internal/pkg/templates/defaults.config.yaml
+++ b/provider-ci/internal/pkg/templates/defaults.config.yaml
@@ -224,7 +224,7 @@ sdkModuleDir: "sdk"
 
 esc:
   enabled: false
-  environment: imports/github-secrets
+  environment: github-secrets/${{ github.repository_owner }}-${{ github.event.repository.name }}
   organization: pulumi
   requestedTokenType: urn:pulumi:token-type:access_token:organization
   environmentVariables: []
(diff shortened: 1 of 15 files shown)


provider-ci/internal/pkg/config.go
```diff
@@ -412,6 +416,14 @@ type toolVersions struct {
 	PulumiCTL string `yaml:"pulumictl"`
 }
 
+type escConfig struct {
```
mjeffryes, line 419: this and the next file are redundant with stacked PR

blampe: Yeah downside to stacks is PRs look weird until they're rebased.

PR #1577, opened 2025-06-11
Expose ESC config to ci-mgmt.yaml

blampe:
This will make it possible for a provider to request their own ESC environment,
and it will enable us to move repo-specific secrets out of `.ci-mgmt.yaml`'s
`env:` key.

Refs https://github.com/pulumi/ci-mgmt/issues/1481

PR #1588, opened 2025-06-18
Add an ESC shim action

blampe:
This adds a small actions which has the same "interface" as ESC, but
which works with our existing CI environments. The action is trivial and
implemented as vanilla JavaScript to eliminate packaging pitfalls. (Only
JavaScript actions support the dynamic outputs we need.)

The action takes the CI job's environment and makes it available as an
output on the step, in the same way we will eventually consume secrets
as outputs from the esc-action step.

A new template method, `renderESCStep`, is also introduced which allows
us to toggle between the shim action and the actual ESC action. This
will allow us to roll out ESC gradually, and without needing to
introduce ESC conditionals around all of our steps.

This is a no-op. A subsequent PR will actually consume the new action.

### DIFF
--- a/provider-ci/internal/pkg/generate.go
+++ b/provider-ci/internal/pkg/generate.go
@@ -335,6 +335,8 @@ func toYAML(v interface{}) (string, error) {
 	return strings.TrimSuffix(string(data), "\n"), nil
 }
 
+// renderESCStep generates either the real ESC action or our shim action which
+// re-exports the existing environment.
 func renderESCStep(v any) (string, error) {
 	config, ok := v.(Config)
 	if !ok {
--- a/provider-ci/internal/pkg/generate.go
+++ b/provider-ci/internal/pkg/generate.go
@@ -318,7 +318,8 @@ func parseTemplate(fsys fs.FS, inPath string) (*template.Template, error) {
 	}
 
 	tmpl, err := template.New(inPath).Funcs(template.FuncMap{
-		"toYaml": toYAML,
+		"toYaml":        toYAML,
+		"renderEscStep": renderESCStep,
 	}).Funcs(sprig.FuncMap()).Delims("#{{", "}}#").Parse(string(inData))
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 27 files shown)


blampe:

- #1577
    - #1588 ◀

provider-ci/internal/pkg/templates/internal/.github/actions/esc-action/action.yaml
```diff
@@ -0,0 +1,14 @@
+name: "Load secrets"
+description: |
+  This is a temporary action which assists with our migration to ESC. Instead
+  of surrounding every step that references secrets with an "if ESC" block, we
+  instead modify those steps to consume their secrets from this step's outputs.
+  Then, later, we can replace this action with esc-action to actually load
+  secrets from ESC.
+#{{ if not .Config.ESC.Enabled -}}#
```
mjeffryes, line 8: why bother with this conditional if renderESCStep only inserts it when  `.Config.ESC.Enabled == False`

provider-ci/internal/pkg/templates/internal/.github/actions/esc-action/index.js
```diff
@@ -0,0 +1,14 @@
+const fs = require("fs");
+
+for (const [name, value] of Object.entries(process.env)) {
+  const file = process.env["GITHUB_OUTPUT"];
+  var stream = fs.createWriteStream(file, { flags: "a" });
```
mjeffryes, line 5: why open the file inside the loop?

provider-ci/internal/pkg/generate.go
```diff
@@ -332,3 +333,44 @@ func toYAML(v interface{}) (string, error) {
 	}
 	return strings.TrimSuffix(string(data), "\n"), nil
 }
+
+func renderESCStep(v any) (string, error) {
```
mjeffryes, line 337: So, we could just copy paste in conditional blocks into the template right? I assume you're adding a function just to make it a little more concise and avoid copy pasta errors?

mjeffryes (approved):
Exporting all environment variables feels a little icky to me, but I suppose it works. Does GHA actually let you pass secrets in `GITHUB_OUTPUT`? I thought they had a heuristic to sanitize it.

PR #1589, opened 2025-06-18
Update steps to reference the ESC shim

blampe:
This updates steps which currently consume the global secrets context to
instead grab their secrets from the ESC step.

If ESC is disabled these outputs will match the existing environmental
secrets.


blampe:

- #1577
    - #1588
        - #1589 ◀

.github/actionlint.yaml
```diff
@@ -1,3 +1,118 @@
 self-hosted-runner:
   labels:
     - pulumi-ubuntu-8core
+
+# Most of these ignores are in place because our esc-secrets action has dynamic
+# outputs, and actionlint assumes all outputs are explicit.
```
mjeffryes, line 6: bummer. :/ Are we going to have to maintain a list of exceptions like this going forward even after switching to the proper ESC action?

mjeffryes (approved):

Issue #1611, opened 2025-07-16
Update pulumi-java to v1.16.0


t0yv0:
We need to rollout pulumi/pulumi-java v1.16.0 update to restore the ability to publish to Maven Central after OSSRH was sunset.

# Providers

## Native Providers

- [x] https://github.com/pulumi/pulumi-aws-apigateway/pull/381
- [x] https://github.com/pulumi/pulumi-aws-native/pull/2333
- [x] https://github.com/pulumi/pulumi-awsx/pull/1657
- [x] https://github.com/pulumi/pulumi-azure-native/pull/4215
- [x] https://github.com/pulumi/pulumi-command/pull/828
- [x] https://github.com/pulumi/pulumi-docker-build/pull/530
- [x] https://github.com/pulumi/pulumi-eks/pull/1836
- [x] [google-native](https://github.com/pulumi/pulumi-google-native) (deprecated)
- [x] https://github.com/pulumi/pulumi-kubernetes/pull/3714
- [x] https://github.com/pulumi/pulumi-kubernetes-cert-manager/pull/541
- [x] https://github.com/pulumi/pulumi-kubernetes-ingress-nginx/pull/298
- [x] https://github.com/pulumi/pulumi-kubernetes-coredns/pull/293

### Tier 1

- [x] https://github.com/pulumi/pulumi-aws/pull/5660
- [x] https://github.com/pulumi/pulumi-gcp/pull/3281
- [x] https://github.com/pulumi/pulumi-azure/pull/3224
- [x] https://github.com/pulumi/pulumi-azuread/pull/2059

### Tier 2

- [x] https://github.com/pulumi/pulumi-auth0/pull/904/
- [x] https://github.com/pulumi/pulumi-cloudflare/pull/1300
- [x] https://github.com/pulumi/pulumi-datadog/pull/892
- [x] https://github.com/pulumi/pulumi-digitalocean/pull/1086
- [x] https://github.com/pulumi/pulumi-docker/pull/1473
- [x] https://github.com/pulumi/pulumi-github/pull/958
- [x] https://github.com/pulumi/pulumi-mongodbatlas/pull/809
- [x] https://github.com/pulumi/pulumi-postgresql/pull/743
- [x] https://github.com/pulumi/pulumi-random/pull/1835
- [x] https://github.com/pulumi/pulumi-tls/pull/812

### Tier 3

- [x] https://github.com/pulumi/pulumi-aiven/pull/950
[...]

PR #1612, opened 2025-07-16
Sketch an experimental job to update Java everywhere

t0yv0:
This job is intended to help with https://github.com/pulumi/ci-mgmt/issues/1611 rollout. 

It is currently experimental and a few follow-up PRs may be coming. With GHA it is difficult to test things out before they are on main.

### DIFF
--- a/.github/workflows/update-java.yml
+++ b/.github/workflows/update-java.yml
@@ -1,0 +1,86 @@
+name: update-java
+
+description: Update pulumi/pulumi-java used in all providers to a target version.
+
+permissions: write-all # Equivalent to default permissions plus id-token: write
+
+on:
+  workflow_dispatch:
+    inputs:
+      pulumi-java-version:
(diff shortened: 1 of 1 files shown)


.github/workflows/update-java.yml
```diff
@@ -0,0 +1,85 @@
...
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+        with:
+          repository: pulumi/pulumi-${matrix.provider}
+          path: provider
+
+      - name: Install upgrade-provider
+        run: go install github.com/pulumi/upgrade-provider@main
+        shell: bash
+
+      - name: Install pulumictl
```
VenelinMartinov, line 63: Providers have a `setup-tools` action: https://github.com/pulumi/pulumi-aws/blob/4a5d90d31974e995964d89d260ae56e0e76d6bef/.github/workflows/upgrade-provider.yml#L52

That might be useful here too

Issue #1605, opened 2025-07-01
Separate java upgrade ci job for bridged providers


VenelinMartinov:
Bridged providers should not upgrade their java version with provider upgrades but instead have a separate job dedicated to that.

t0yv0:
That's really needed now.

t0yv0:
Example of the workflow working end to end:

Calling https://github.com/pulumi/ci-mgmt/actions/runs/16328313981 with provider=random version=1.16.0

Dispatches this job: https://github.com/pulumi/pulumi-random/actions/runs/16328321164/job/46124153106

Creates this PR: https://github.com/pulumi/pulumi-random/pull/1835

t0yv0:
There is a problem with PRs generated by this job.

https://github.com/pulumi/pulumi-auth0/pull/904 is opened by [github-actions](https://github.com/apps/github-actions) bot

In contrast:

https://github.com/pulumi/pulumi-auth0/pull/889 is opened by [pulumi-bot](https://github.com/pulumi-bot)

which is probably what we would want, because github-actions opened PRs need closing and reopening to kick off tests runs.

I wonder what is different in the environment

guineveresaenger:
> pull request create failed: GraphQL: No commits between master and upgrade-java-version-to-1.16.0-ci (createPullRequest)
cf: https://github.com/pulumi/pulumi-postgresql/actions/runs/16348201837/job/46187781210#step:6:18

Running this separately is great; but failure to upgrade because the Java version is already updated should not result in p1s.

Filed https://github.com/pulumi/upgrade-provider/issues/346.

PR #1623, opened 2025-07-16
Propagate GH_TOKEN

t0yv0:
This attempts to address an issue that java update PRs have the wrong owner and do not trigger tests properly: https://github.com/pulumi/ci-mgmt/issues/1605#issuecomment-3081399781


t0yv0:
This does seem to fix the issue, and PR generated using this method is https://github.com/pulumi/pulumi-akamai/pull/798 using pulumi-bot and is automatically running the tests.

PR #1629, opened 2025-07-17
Adding mise.toml file generation.

pose:
Projects managed by `ci-mgmt` lack consistent version enforcement, leading to discrepancies between CI and local development environments. Having to manual synchronize CI and local environments is error prone. For example, various CONTRIBUTING.md files specify the need for tools like golangci-lint or pulumictl without indicating a valid version range.

This PR generates a `mise.toml` on all the projects managed by `ci-mgmt` that matches the tool and runtime versions used by CI.

### DIFF
--- a/.github/workflows/test-provider-ci.yml
+++ b/.github/workflows/test-provider-ci.yml
@@ -38,6 +38,8 @@ jobs:
         run: git config --global init.defaultBranch master
       - name: Install mise
         uses: jdx/mise-action@v2
+        with:
+          version: 2025.7.12
       - name: Build & test
         run: cd provider-ci && make all
       - name: Check worktree clean


provider-ci/internal/pkg/templates/mise/mise.toml
```diff
@@ -0,0 +1,13 @@
+# WARNING: This file is autogenerated - changes will be overwritten when regenerated by https://github.com/pulumi/ci-mgmt
+
+[tools]
+pulumi = 'latest'
+"go:github.com/pulumi/pulumictl/cmd/pulumictl" = 'latest'
+go = '#{{ trimAll ".x" .Config.ToolVersions.Go }}#'
```
pose, line 6: The way on how I'm extracting the `.x` is a bit brittle. If someone writes the version such as `1.x.3` it would fail and install `1.3`. However, will that case ever occur in the while?

provider-ci/internal/pkg/templates/mise/mise.toml
```diff
@@ -0,0 +1,13 @@
...
+node = '#{{ trimAll ".x" .Config.ToolVersions.Nodejs }}#'
+python = '#{{ trimAll ".x" .Config.ToolVersions.Python }}#'
+dotnet = '#{{ trimAll ".x" .Config.ToolVersions.Dotnet }}#'
+gradle = '#{{ trimAll ".x" .Config.ToolVersions.Gradle }}#'
+
+# Corretto version used as Java SE/OpenJDK version no longer offered
```
pose, line 12: Happy to change to another JDK provider. This is the one I'm familiar with.

provider-ci/test-providers/acme/mise.toml
```diff
@@ -0,0 +1,13 @@
+# WARNING: This file is autogenerated - changes will be overwritten when regenerated by https://github.com/pulumi/ci-mgmt
+
+[tools]
+pulumi = 'latest'
+"go:github.com/pulumi/pulumictl/cmd/pulumictl" = 'latest'
```
pose, line 5: (question) Do you think I should least these two binaries at the end and group it with gradle? They are not runtimes. I'm thinking about adding a `# runtimes` section to make this clearer. What do you think?

provider-ci/internal/pkg/templates/mise/mise.toml
```diff
@@ -0,0 +1,13 @@
+# WARNING: This file is autogenerated - changes will be overwritten when regenerated by https://github.com/pulumi/ci-mgmt
+
```
pose, line 2: This file is missing the open overrides section that we need to implement where `.ci-mgmt.yml` will inject the local overrides (e.g. `kubectl` if you are developing `pulumi-kubernetes`).

provider-ci/internal/pkg/templates/mise/mise.toml
```diff
@@ -0,0 +1,13 @@
...
+[tools]
+pulumi = 'latest'
+"go:github.com/pulumi/pulumictl/cmd/pulumictl" = 'latest'
+go = '#{{ trimAll ".x" .Config.ToolVersions.Go }}#'
+node = '#{{ trimAll ".x" .Config.ToolVersions.Nodejs }}#'
+python = '#{{ trimAll ".x" .Config.ToolVersions.Python }}#'
+dotnet = '#{{ trimAll ".x" .Config.ToolVersions.Dotnet }}#'
+gradle = '#{{ trimAll ".x" .Config.ToolVersions.Gradle }}#'
```
Graham-Pedersen, line 10: Another silly question, is this logic of the trim performed by MISE in some form or is it make that does these substitutions before mise is invoked?

provider-ci/Makefile
```diff
@@ -51,6 +51,9 @@ test-provider/%: bin/provider-ci $(ACTIONLINT)
 	cp -r test-providers/$(PROVIDER_NAME) bin/test-provider
 	cd bin/test-provider/$(PROVIDER_NAME) && git init
 	cd bin/test-provider/$(PROVIDER_NAME) && ../../../$(ACTIONLINT) -config-file ../../../../.github/actionlint.yaml
+	# Experimental mise flag required to support building golang binaries (e.g. pulumictl)
+	mise settings experimental=true
```
pose, line 55: Oops, missed adding the Mise github action: https://github.com/jdx/mise-action. This is currently failing since it cannot find `mise`. I'll fix tomorrow 🙏

provider-ci/internal/pkg/templates/mise/mise.toml
mjeffryes: FWIW, this template is using the same approach and is has worked for the last 18 mo just fine: [ci-mgmt/provider-ci/internal/pkg/templates/base/devbox.json at master · pulumi/ci-mgmt](https://github.com/pulumi/ci-mgmt/blob/master/provider-ci/internal/pkg/templates/base/devbox.json)

provider-ci/internal/pkg/templates/mise/mise.toml
mjeffryes: ci-mgmt is using go templating library [template package - text/template - Go Packages](https://pkg.go.dev/text/template) to translate these template files when it generates files for a repository ([ci-mgmt/provider-ci/internal/pkg/generate.go at master · pulumi/ci-mgmt](https://github.com/pulumi/ci-mgmt/blob/master/provider-ci/internal/pkg/generate.go#L320))

provider-ci/internal/pkg/templates/mise/mise.toml
mjeffryes: I've been using corretto too FWIW

provider-ci/test-providers/acme/mise.toml
mjeffryes: I don't think it's a big deal at this point, the list is short. 

I have mixed feelings about using latest for pulumi CLI as I think most of our repos have it pinned, but I don't think the version is in ci-mgmt config right now, so maybe we should just leave this as latest for now.

provider-ci/internal/pkg/templates.go
```diff
@@ -27,6 +27,8 @@ const (
 	native = "native"
 	// workflows for aws-native provider
 	awsNative = "aws-native"
+	// mise.toml file generation
+	mise = "mise"
```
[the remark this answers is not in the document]
mjeffryes: maybe we should name this something slightly more generic and move the devbox + devcontainer templates in here too.

mjeffryes (approved):
Left some comments, but I think this could ship as is once we add mise to the ci job

provider-ci/internal/pkg/templates.go
```diff
@@ -39,21 +41,21 @@ func getTemplateDirs(templateName string) ([]TemplateDir, error) {
 	switch templateName {
 	case "bridged-provider":
 		// Any Pulumi-owned bridged provider
-		return []TemplateDir{base, internal, bridged, internalBridged}, nil
+		return []TemplateDir{base, internal, bridged, internalBridged, mise}, nil
```
danielrbradley, line 44: In terms of layering, I think "mise" should go after "base", but before the more specific layers - in case they need to override the mise file.

```suggestion
		return []TemplateDir{base, mise, internal, bridged, internalBridged}, nil
```

danielrbradley (approved):
What's here looks like a good point to start at.

Minor nit-pick level comments:
1. I'm not convinced that very small per-feature template directories will make things easier to manage. I'd still lean towards just placing this in the "base" directory personally, but happy to see how this goes with other features.
2. If we are doing the feature-based template directory, I think this default feature should be applied before the more specific provider templates, rather than after, so specific types of providers could override their mise file.
3. For bridged providers, they definitely don't use pulumictl any more, so wouldn't need this downloaded via mise. One option could be in the bridged template directory to add a different variation of the mise.toml, without pulumictl. The other option would be to replicate the work from the bridged providers intot the native providers to remove the need for pulumictl.

provider-ci/internal/pkg/templates/mise/mise.toml
pose: Then, I'm going to leave this as is. Thanks to both!

provider-ci/test-providers/acme/mise.toml
pose: Sounds good. Let's leave `pulumi` CLI here and we change it if we need it in the future.

provider-ci/internal/pkg/templates.go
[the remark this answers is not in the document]
pose: Also, another question we need to answer is whether we extend the Dockerfile in `.devcontainer` to use Mise as well.

provider-ci/internal/pkg/templates/mise/mise.toml
pose: Thanks! I replaced this file and the dev containers one with `trimSuffix`.

provider-ci/internal/pkg/templates/mise/mise.toml
pose: (I'm planning to add this on the next PR)

provider-ci/Makefile
pose: Done. CI is now passing 🥳

PR #1642, opened 2025-08-07
Fix prerequisite permissions

thatliuser:
On Formal's org (and presumably other orgs out there), the default GitHub CI token permissions do not include commenting on PRs, which breaks CI. It's fixable by editing the yaml file, but that's an autogenerated file, so ideally this should be fixed in the CI generator.

### DIFF
--- a/provider-ci/internal/pkg/templates/base/.github/workflows/main.yml
+++ b/provider-ci/internal/pkg/templates/base/.github/workflows/main.yml
@@ -4,6 +4,9 @@ env:
 #{{ .Config.Env | toYaml | indent 2 }}#
 jobs:
   prerequisites:
+    permissions:
+      contents: read
+      pull-requests: write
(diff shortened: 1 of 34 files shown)


thatliuser:
Hey everyone, I ran into an issue with the prerequisites CI while setting up a Pulumi provider using the boilerplate repository here as a template: https://github.com/pulumi/pulumi-tf-provider-boilerplate

The issue was that the GitHub CI runner didn't have access commenting on a pull request, which was fixed by the change added above in the `templates/` repo. I then ran `make test-providers` to sync the other `prerequisite.yml` files. Let me know if I need to make any other changes or if this is a mistake on our end!

thatliuser:
@mjeffryes Thank you for reviewing the PR! I just added another commit that also adds the token permissions for the other yaml files that import the same job. Also it looks like there's a lint error in the sample Pulumi provider, but I think this is an unrelated issue.

PR #1649, opened 2025-08-07
Remove java key secrets from the global environment

blampe:
We only need these when publishing the Java SDK.

Refs #1481.


blampe:
> Just noticed that we have both a "build" and a "prerelease" workflow in native providers templates that are both set up to push prerelease SDKs! I wonder if both are actually used and if so how!

Yeah, weird. Looks like build ignores tags and prerelease doesn't (but also expects something like vX.Y.Z-alpha, which we rarely do). Seems like the build workflow should be able to handle both of these cases with the generic version stuff.

PR #1650, opened 2025-08-07
Remove PYPI secrets from the global environment

blampe:
These are only needed when publishing the Python SDK.

Refs #1481.


provider-ci/internal/pkg/templates/native/.github/workflows/prerelease.yml
```diff
@@ -20,8 +20,6 @@ env:
   NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
   NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
   NUGET_PUBLISH_KEY: ${{ secrets.NUGET_PUBLISH_KEY }}
-  PYPI_USERNAME: __token__
-  PYPI_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
```
mjeffryes, line 24: Do we not need to add these to the publish step below? I don't see where that's picking them up if not from this global env

PR #1651, opened 2025-08-07
Remove GITHUB_TOKEN secret from global environment

blampe:
We don't need to expose our GitHub token to every action.

This is the most likely the break things since we (probably) have a lot of scripts that implicitly depend on `GITHUB_TOKEN` being set. I'll be merging this one after some of the other cleanup.

Refs #1481.


Graham-Pedersen:
I would love to give this a ship it, but I wanted to ask about two things. 

First the check that failed. Seems like that should pass
Second, as to Matt's comment, how are we planning to roll this out. When we ship it can we run the CI management changes on a couple of providers by hand quickly, and verify they are working? Seems like the blast radius of this change is very large and I would like to catch problems quickly so we can roll back.

PR #1655, opened 2025-08-12
Fix p-k test infra

blampe:
We need to generate and provider a Pulumi token to the cluster setup and teardown steps.


mjeffryes (approved):

blampe:
Working here https://github.com/pulumi/pulumi-kubernetes/pull/3756

PR #1656, opened 2025-08-12
Gradle v4

EronWright:
Updates various scripts to use `gradle/actions/setup-gradle@v4` rather than `gradle/gradle-build-action@v3`.

Note that `gradle-build-action@v3` actually has a superset of functionality as seen here:
https://github.com/pulumi/pulumi-azure-native/pull/4256/files#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7L288-L292

There may be scripts that still use `gradle-build-action@v3` for build or publishing.  This PR is scoped to the setup task.

### DIFF
--- a/provider-ci/internal/pkg/templates/native/.github/workflows/prerelease.yml
+++ b/provider-ci/internal/pkg/templates/native/.github/workflows/prerelease.yml
@@ -287,7 +287,7 @@ jobs:
         distribution: temurin
         cache: gradle
     - name: Setup Gradle
-      uses: gradle/gradle-build-action@ac2d340dc04d9e1113182899e983b5400c17cda1 # v3.5.0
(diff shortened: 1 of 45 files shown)


provider-ci/internal/pkg/templates/native/.github/workflows/prerelease.yml
```diff
@@ -800,7 +800,7 @@ jobs:
       run: tar -zxf ${{github.workspace}}/sdk/java.tar.gz -C
         ${{github.workspace}}/sdk/java
     - name: Publish Java SDK
-      uses: gradle/gradle-build-action@ac2d340dc04d9e1113182899e983b5400c17cda1 # v3.5.0
+      uses: gradle/actions/setup-gradle@017a9effdb900e5b5b2fddfb590a105619dca3c3 # v4.4.2
```
EronWright, line 803: Dang, I see this is a mistake

Issue #1543, opened 2025-05-14
Use `golangci-lint` v2


ringods:
The `lint_provider<.fix>` targets use an installed `golangci-lint` tool. Locally, I had a v2 installed, but it seems that the config file is still for a V1.

Is there anything blocking us to migrate to `golangci-lint` v2?

t0yv0:
Ran into this as well today, it's quite pesky to lint with golangci-lint. I think it might be tied to migrating Go to 1.24. 

I think preferred path forward may be to go ahead with @blampe suggestion here and pin golangci-lint used in the project with mise, and subsequently migrate to the latest:

https://github.com/pulumi/ci-mgmt/issues/1263

kangasta:
The `golangci-lint` v2 (more specifically version `>= 2.4.0`) and golangci-lint-action v7 would also be required for Go 1.25 support.

Our (@UpCloudLtd) Terraform provider is built with Go 1.25 now, so upgrade provider workflow in our external bridged provider automatically bumped Go version to 1.25 also in the Pulumi provider configuration, which now causes the linting job to fail with:

```
Error: can't load config: the Go language version (go1.24) used to build golangci-lint is lower than the targeted Go version (1.25.0).
```

PR #1689, opened 2025-09-05
Upgrade to golangci-lint v2

ringods:
Fixes: #1543

* Upgraded the `.golangci.yml` config file to `v2`
* Upgraded the `lint.yml` workflow to `golangci-lint` v2.4.0 and `go` v1.24.0. The Go version in this workflow is separate from the configured Go version for the rest of the workflows.
* Added the tool to `mise.toml` for consistency

Changes from this branch were tested manually on a Pulumiverse provider:

* failing run before upgrade: https://github.com/pulumiverse/pulumi-talos/actions/runs/17473609254/job/49627875748#step:6:34
* succeeding run after upgrade: https://github.com/pulumiverse/pulumi-talos/actions/runs/17488569455/job/49672936588

(need to investigate the license check error in my runs, but thes are unrelated to the lint changes)


blampe (changes requested):
Doing a blanket upgrade like this will break a lot of provider builds until we manually resolve these lint issues. That's not a good use of time.

Instead it should be possible for a provider to opt-in to their own version of the linter and we should respect that in the workflows.

t0yv0:
I'm deep down on Neo work so I'll let @blampe tackle this unless you need me! TYVM.

PR #1705, opened 2025-09-10
Fix GITHUB_TOKEN permissions for ESC

blampe:
When ESC is enabled our GITHUB_TOKEN needs `id-token: write` permissions. I added this to each job that uses the ESC step. As a bonus our steps have less write access than they normally would by default.

As part of this I'm also going to force the boilerplate to enable ESC. Normally we would enable it in the repo's `.ci-mgmt.yml`, but we have a chicken-and-egg problem while the ESC-enabled workflows are broken.

Refs https://github.com/pulumi/ci-mgmt/issues/1481.

### DIFF
--- a/provider-ci/internal/pkg/generate.go
+++ b/provider-ci/internal/pkg/generate.go
@@ -373,7 +373,7 @@ func renderESCStep(v any) (string, error) {
 		step := map[string]any{
 			"name": "Fetch secrets from ESC",
 			"id":   "esc-secrets",
-			"uses": "pulumi/esc-action@f93f6337967c9937a15826438a6f817d6bc7f872",
+			"uses": "pulumi/esc-action@9eb774255b1a4afb7855678ae8d4a77359da0d9b",
 			"env":  env,
 		}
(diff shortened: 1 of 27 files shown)


provider-ci/internal/pkg/generate.go
```diff
@@ -165,6 +165,14 @@ func renderTemplateDir(template TemplateDir, opts GenerateOpts) error {
...
 	projName := strings.TrimPrefix(opts.RepositoryName, "pulumi/")
 
+	if projName == "pulumi-provider-boilerplate" {
+		config.ESC.Enabled = true // Temporarily forcing ESC to test workflows.
+		config.ESC.Environment = "imports/github-secrets"
+	}
+	if projName == "pulumi-xyz" {
+		config.ESC.Enabled = true // Temporarily forcing ESC to test workflows.
+	}
+
```
mjeffryes, line 175: Let's file a ticket to make sure we clean this up later?

mjeffryes (approved):

PR #1730, opened 2025-09-16
Replace setup-tools with mise

### LINKED ISSUE #1610: Use jdx/mise-action for tool setup if a provider has a .mise.toml present

corymhall:
fixes #1610


provider-ci/internal/pkg/templates/base/.github/workflows/build_provider.yml
```diff
@@ -61,12 +61,12 @@ jobs:
...
-        uses: ./.github/actions/setup-tools
+      - name: Setup mise
+        uses: jdx/mise-action@v3
         with:
-          tools: pulumictl, go
-          # use per-platform/arch caches instead since we are doing cross-builds
-          cache-go: false
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          cache_key: "mise-{{platform}}-{{file_hash}}"
+          cache_save: false
```
blampe, line 69: Why not cache the tools?

blampe (commented):
This should essentially take care of https://github.com/pulumi/ci-mgmt/pull/1689.

provider-ci/internal/pkg/templates/base/.github/actions/setup-tools/action.yml
```diff
@@ -1,100 +0,0 @@
-name: Setup tools
-description: Installs Go, Pulumi, pulumictl, schema-tools, Node.JS, Python, dotnet and Java.
```
blampe, line 2: Thinking out loud, we're replacing this with an individual step (which is great) but I wonder if there's any value in keeping this action around. It's kinda nice that it's already factored out, so if we ever need to do something else during setup we only have one place to fix. Normally I would say YAGNI but we tend to touch this stuff a lot.

Edit: actually after reading a bit more, it seems like it would already be natural for this to handle mise + the Go cache step. We already have the cache-go input, because just like mise we might only want to cache in one place.

provider-ci/internal/pkg/templates/base/.config/mise.toml
```diff
@@ -0,0 +1,20 @@
...
+# Runtimes
+go = '#{{ trimSuffix ".x" .Config.ToolVersions.Go }}#'
+node = '#{{ trimSuffix ".x" .Config.ToolVersions.Nodejs }}#'
+python = '#{{ trimSuffix ".x" .Config.ToolVersions.Python }}#'
+dotnet = '#{{ trimSuffix ".x" .Config.ToolVersions.Dotnet }}#'
+# Corretto version used as Java SE/OpenJDK version no longer offered
+java = 'corretto-#{{trimSuffix ".x" .Config.ToolVersions.Java }}#'
+
+# Executable tools
+pulumi = 'latest'
```
blampe, line 15: Is it worth copying over your go.mod trick from the bridge while we're in here?

If not, we should probably at least track `mise.lock`. This file isn't generated by default, but `touch mise.lock && mise install` will add it.

provider-ci/internal/pkg/templates/base/.config/mise.toml
```diff
@@ -0,0 +1,20 @@
+# WARNING: This file is autogenerated - changes will be overwritten when regenerated by https://github.com/pulumi/ci-mgmt
+# You can create your own mise.toml file to override/augment this. See https://mise.jdx.dev/configuration.html
```
blampe, line 2: Let's add a comment here explaining that a root-level `mise.toml` can be used to override these defaults.

provider-ci/internal/pkg/templates/base/.config/mise.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[tools]
+
+# Runtimes
+go = '#{{ trimSuffix ".x" .Config.ToolVersions.Go }}#'
+node = '#{{ trimSuffix ".x" .Config.ToolVersions.Nodejs }}#'
+python = '#{{ trimSuffix ".x" .Config.ToolVersions.Python }}#'
+dotnet = '#{{ trimSuffix ".x" .Config.ToolVersions.Dotnet }}#'
+# Corretto version used as Java SE/OpenJDK version no longer offered
+java = 'corretto-#{{trimSuffix ".x" .Config.ToolVersions.Java }}#'
```
blampe, line 12: We don't necessarily need to do this now, but with this change we can stop specifying tool overrides in ci-mgmt.yml and instead specify them in the root mise.toml. This file can then just specify whatever our defaults are. _In theory_ a tool like Renovate could then handle bumping those defaults over time, but in practice we have so many other build issues that it won't work smoothly.

provider-ci/internal/pkg/templates/base/.github/workflows/build_provider.yml
blampe: Ah that totally makes sense. It would be worth adding a comment so we don't accidentally undo this.

provider-ci/internal/pkg/migrations/ignoreMiseLocal.go
```diff
@@ -0,0 +1,40 @@
...
+// committed to source control. see https://mise.jdx.dev/configuration.html
+type ignoreMiseLocal struct{}
+
+func (ignoreMiseLocal) Name() string {
+	return "Add mise.local.toml to .gitignore"
+}
+func (ignoreMiseLocal) ShouldRun(templateName string) bool {
+	return templateName == "bridged-provider" || templateName == "mise"
+}
+func (ignoreMiseLocal) Migrate(templateName, outDir string) error {
```
blampe, line 21: 👍 love to see migrations like this!

blampe (approved):
This is great. A lot of my feedback is commentary on how things are currently and doesn't need to block you.

provider-ci/internal/pkg/templates/base/.github/actions/setup-tools/action.yml
corymhall: Sure I can add it back. I personally like the flattened workflows because I think the separated actions make it harder to reason about what is happening, but that is the status quo.

pose (commented):
Some minor comments to, mostly, educate myself. Thanks again for driving this 🙏

provider-ci/internal/pkg/generate.go
```diff
@@ -165,6 +172,10 @@ func renderTemplateDir(template TemplateDir, opts GenerateOpts) error {
 
 	projName := strings.TrimPrefix(opts.RepositoryName, "pulumi/")
 
+	if projName == "pulumi-provider-boilerplate" {
+		config.ModulePath = "."
+	}
```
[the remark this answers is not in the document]
corymhall: Yes we should and then we can remove this.

provider-ci/test-providers/acme/scripts/get-versions.sh
```diff
@@ -0,0 +1,54 @@
...
+                exit
+            }
+        }
+    }
+' "$gomod")
+
+if [[ -z "${raw_version:-}" ]]; then
+  echo "failed to determine Pulumi version from $gomod" >&2
+  # default to latest
+  raw_version="latest"
```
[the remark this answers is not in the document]
corymhall: I think the only time this would fail is in the tests for this repo. The tests were failing because the tests don't have the `go.mod` file to parse.

provider-ci/internal/pkg/migrations/createMiseLock.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+		pulumiVersion, goVersion, err := getVersions(outDir)
+		if err != nil {
+			return fmt.Errorf("error getting go version from go.mod: %w", err)
+		}
+		cmd := exec.Command("mise", "install")
+		cmd.Dir = outDir
+		cmd.Env = append(os.Environ(),
+			fmt.Sprintf("PULUMI_VERSION=%s", pulumiVersion),
+			fmt.Sprintf("GO_VERSION=%s", goVersion),
```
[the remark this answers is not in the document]
corymhall: In our `.config/mise.toml` we use these env vars to resolve the version to use.

provider-ci/internal/pkg/templates/all/scripts/get-versions.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# This script can be simplified to use go when https://github.com/jdx/mise/discussions/6374 is fixed
+# e.g. go list -m -f '{{.GoVersion}}'
+
+module_path="github.com/pulumi/pulumi/pkg/v3"
+gomod="#{{ .Config.ModulePath }}#/go.mod"
```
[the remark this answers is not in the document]
corymhall: Yes we should probably move to using mise plugins for this. It would also allow us to install the dev version of the pulumi cli instead of just latest. I'll add a backlog item.

provider-ci/internal/pkg/templates/all/scripts/get-versions.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
pose: Ah! Sorry, now I see what you mean. I thought we only wanted to `source` the script but we are also printing the values so we can later parse the values.

PR #1736, opened 2025-09-22
Fix `esc-action` shim generation for external providers

ringods:
The introduction of the ESC secrets support introduced a regression. External providers which have the ESC support disabled did not have the `esc-action` shim generated. Also the `terraform-module` test provider didn't have the shim in its setup.

I refactored the `mise` template into an `all` template for files which should be generated for all providers without exception. Moving the `esc-action` shim into `all` resolves the problem for external providers and `terraform-module`, but has the side-effect that the shim is also generated for `pulumi-provider-boilerplate`. I did it in this smallest possible way because restructuring any other way lead me to a rabbit hole every single time.


mjeffryes (approved):
This seems like a good approach to fix this for external providers. I think it's good that the shim is getting generated for pulumi-provider-boilderplate too; since that's the starting place for external providers.

 Unfortunately, it looks like the boilerplate is still failing on ESC; it probably should be configured not to use ESC by default?

PR #1756, opened 2025-09-30
Rename mise env vars

corymhall:
Turns out that `MISE_` env vars [are
special](https://github.com/jdx/mise/discussions/6476) and will prevent
the version from being written to the lockfile (or from MISE_ENV from working)


guineveresaenger (approved):
🤯 

LGTM!

PR #1760, opened 2025-10-01
Ignore mise.lock in `git-status-check`

corymhall:
The mise.lock file is only updated when a machine first downloads a tool
version. If there is a cache on the machine, the lockfile is not updated
with the checksum entry. It also only adds a checksum entry for the arch
of the machine that is installing the tool. This means that sometimes CI
will update the mise.lock file and cause the git-status-check to fail.

### DIFF
--- a/provider-ci/internal/pkg/templates/base/.github/workflows/build_sdk.yml
+++ b/provider-ci/internal/pkg/templates/base/.github/workflows/build_sdk.yml
@@ -94,6 +94,8 @@ jobs:
             sdk/nodejs/package.json
             sdk/python/pyproject.toml
             sdk/java/build.gradle
+            **/mise.lock
+            **/mise.*.lock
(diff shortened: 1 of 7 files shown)


mjeffryes (approved):

blampe:
If a user added a tool shouldn't the lock file have already been updated / committed? If it's getting updated on CI that seems like a problem.

corymhall:
> If a user added a tool shouldn't the lock file have already been updated / committed? If it's getting updated on CI that seems like a problem.

@blampe I can't find a way for mise to update the lockfile for _all_ architectures. e.g. I run `mise install` locally and it adds checksums for macos and then in CI it tries to add checksums for linux.

blampe:
Yeah thanks for explaining, seems like a mise bug more than anything.

PR #1792, opened 2025-10-16
Migrate ci-mgmt `plugins` entries to mise

### LINKED ISSUE #1474: Auto-update provider pins
.ci-mgmt.yaml pins providers such as aws=6.32.0. This is needed because it affects example generation and examples are serialized into the schema which is checked into the repository, as well as into SDKs which are also checked into the repository.  The suggestion here is to automate keeping these updated via PRs:

```
provider: random
major-version: 4
makeTemplate: bridged
plugins:
  - name: std
    version: "1.6.2"
  - name: terraform
    version: "1.0.16"
    kind: converter
  - name: aws
    version: "6.32.0"
  - name: azure
    version: "5.24.0"
team: ecosystem
license:
  ignore:
    - github.com/terraform-providers/terraform-provider-random/shim
pulumiConvert: 1

```

The automation would need to update e.g. plugins[name='aws'] version and call make generate_sdks and check in the result.

corymhall:
This PR migrates the `plugins` entry in `ci-mgmt.yaml` to
`.config/mise.toml`. To simplify the mise migrations I also refactored a
little so that we now have the following flow:

- `deleteOldMiseConfig`: Runs first and deletes the old root level
  `mise.toml` if it exists and is not the overrides file
- `migrateCimgmtOverrides`: Migrates things from `ci-mgmt.yaml` to the
  root level override file (currently just `toolVersions`).
- `migrateCimgmtToMise`: Migrates things from `ci-mgmt.yaml` to the
  `.config/mise.toml` file.
- `maintainMiseLock`: Runs last and runs `mise install` to update the
  lockfiles.

I've set the plugins version to "latest" for all the plugins so that we
can implement an upgrade workflow (e.g. via `mise upgrade`). Also, since
the plugins are added to the `.config/mise.toml` we can add specific
version overrides to the root level `mise.toml`.

## Additional things

- I've added some generic functionality for dealing with `toml` and `yaml` files so it will be easier to maintain these going forward
- Since `mise` can automatically install `plugins`, I've also created a backend plugin for installing the plugins https://github.com/pulumi/vfox-pulumi
  - This plugin will be automatically installed with `mise install` so that the tools defined using the plugins can be installed.
- Added `mise_install` and `mise_env` make targets. This should make it so that as long as you have `mise` installed, running any of the `make` targets should automatically use the `mise` environment. 

closes #1474

### DIFF
--- a/.github/workflows/test-windows.yml
+++ b/.github/workflows/test-windows.yml
@@ -10,5 +10,7 @@ jobs:
       - name: Install mise
         uses: jdx/mise-action@e3d7b8d67a7958d1207f6ed871e83b1ea780e7b0 # v3
         with:
+          # Latest working version. See https://github.com/jdx/mise/discussions/6781
+          version: 2025.10.16
           github_token: ${{ secrets.GITHUB_TOKEN }}
 
--- a/.github/workflows/update-workflows.yml
+++ b/.github/workflows/update-workflows.yml
@@ -95,7 +95,8 @@ jobs:
       - name: Install mise
         uses: jdx/mise-action@e3d7b8d67a7958d1207f6ed871e83b1ea780e7b0 # v3
         with:
-          version: 2025.10.15
+          # Latest working version. See https://github.com/jdx/mise/discussions/6781
+          version: 2025.10.16
           install: false
           cache: false
           working_directory: pulumi-${{ inputs.provider_name }}
--- a/provider-ci/internal/pkg/templates/base/.github/workflows/build_provider.yml
+++ b/provider-ci/internal/pkg/templates/base/.github/workflows/build_provider.yml
@@ -64,6 +64,8 @@ jobs:
       - name: Setup mise
         uses: jdx/mise-action@v3
         with:
+          # Latest working version. See https://github.com/jdx/mise/discussions/6781
+          version: 2025.10.16
           github_token: ${{ secrets.GITHUB_TOKEN }}
           cache_key: "mise-{{platform}}-{{file_hash}}"
           # only saving the cache in the prerequisites job
--- a/provider-ci/internal/pkg/templates/base/.github/workflows/build_sdk.yml
+++ b/provider-ci/internal/pkg/templates/base/.github/workflows/build_sdk.yml
@@ -55,6 +55,8 @@ jobs:
       - name: Setup mise
         uses: jdx/mise-action@v3
         with:
+          # Latest working version. See https://github.com/jdx/mise/discussions/6781
+          version: 2025.10.16
           github_token: ${{ secrets.GITHUB_TOKEN }}
           cache_key: "mise-{{platform}}-{{file_hash}}"
           # only saving the cache in the prerequisites job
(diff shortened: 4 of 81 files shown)


provider-ci/internal/pkg/migrations/migrate_mise_config.go
```diff
@@ -1,149 +1,48 @@
...
+			return nil
+		}
+		return fmt.Errorf("error reading old mise.toml: %w", err)
 	}
 
-	// If the old root-level mise.toml exists, remove it first.
-	if miseExists && strings.Contains(string(oldFile), "# WARNING: This file is autogenerated - changes will be overwritten when regenerated by https://github.com/pulumi/ci-mgmt") {
+	// If the old root-level mise.toml exists, remove it.
+	// The new overrides file will not have the `# WARNING` at the top
+	if strings.Contains(string(oldFile), "# WARNING: This file is autogenerated - changes will be overwritten when regenerated by https://github.com/pulumi/ci-mgmt") {
```
blampe: 👍

provider-ci/internal/pkg/migrations/migrate_cimgmt_to_mise.go
```diff
@@ -0,0 +1,124 @@
...
+
+	plugins := cimgmt.getFieldNode("plugins")
+	if plugins == nil {
+		return nil
+	}
+
+	// If we have any plugins overrides, move them to the .config/mise.toml
+	pluginList := nodeToPluginEntries(plugins)
+	entries := pluginsToToolEntries(pluginList)
+	updated, err := mise.ensureToolsEntries(entries)
```
blampe: Love this. We're really starting to decouple ci-mgmt's workflow management responsibilities from its version management responsibilities.

provider-ci/internal/pkg/migrations/ci_mgmt_yaml.go
```diff
@@ -0,0 +1,102 @@
...
+	"fmt"
+	"os"
+
+	"gopkg.in/yaml.v3"
+)
+
+// cimgmtYaml wraps the contents of a .ci-mgmt.yaml file along with its source path.
+type cimgmtYaml struct {
+	path string
+	// we operate on the yaml.Node so that we can preserve comments and ordering
```
blampe: 👍

provider-ci/internal/pkg/migrations/ci_mgmt_yaml_test.go
```diff
@@ -0,0 +1,65 @@
+package migrations
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestCimgmtYamlDeleteKeyAndWrite(t *testing.T) {
```
blampe: This is great, it might be our first real Go test for the repo. We should double check these are run on CI -- I wouldn't be surprised if they aren't.

blampe (approved):

blampe (approved):

PR #1793, opened 2025-10-16
Fix `/run-acceptance-tests` for bridged providers

blampe:
This adds an optional `sha` argument to our reusable workflows. When we get a community PR, we propagate the PR's merge commit SHA to those workflows.

`run-acceptance-tests.yml` is modified to create and update a status check for Sentinel for these community PR workflows.

We aren't able to use the native "Approve workflows to run" button because our ESC secrets require `id-token: write` permission but PRs from forks can only have `read`.

Fixes https://github.com/pulumi/ci-mgmt/issues/1409


provider-ci/internal/pkg/templates/base/.github/workflows/test.yml
```diff
@@ -5,15 +5,15 @@ name: "Test Provider"
...
+         type: string
+         description: The commit to build, primarily for community PRs.
        version:
          required: true
          type: string
          description: Version of the provider to test
 
 env:
-  PR_COMMIT_SHA: ${{ github.event.client_payload.pull_request.head.sha }}
-  MISE_ENV: test
```
EronWright: What is the impact of removing `MISE_ENV`

provider-ci/test-providers/acme/.github/workflows/run-acceptance-tests.yml
```diff
@@ -9,9 +9,9 @@ on:
   repository_dispatch:
     types:
     - run-acceptance-tests-command
+  workflow_dispatch: {}
```
EronWright: Is workflow_dispatch actually being used?

EronWright (commented):
Looking good.

An oddity is that the natural run-acceptable-test still waits for maintainer approval.
- https://github.com/pulumi/pulumi-xyz/actions/runs/18576585729

PR #1808, opened 2025-11-04
Add gates for backported version releases

### LINKED ISSUE #1806: Add gates for backported releases
When we release a stable provider version that is not intended as `latest`, but is also not a prerelease, we need to ensure our tooling doesn't make inappropriate assumptions:

- GitHub release should not be set as `latest` automatically when a non-latest backport release is cut
- No docs build should happen on releases that are not a latest release (our registry site only supports one version of the docs, and since that is the case, we should only ever publish `latest` docs.)

guineveresaenger:
This pull request:

- Determines whether the tag that triggered the `publish` Workflow was based off of the default branch (i.e. a regular release) or off a feature branch (as would be the case for a backport release).
- Uses the `make_latest` [input to the release Action](https://github.com/softprops/action-gh-release?tab=readme-ov-file#-customizing) to explicitly state whether this release should be marked as `latest` on GitHub, using whether we're on `main` or not as the logic to determine the value
- Gates a docs build on whether we're on `main`/`master` as well.

Considerations:

- This change allows for _any_ tags on default branch to set the release as `latest`; it doesn't actually have to be the tip of default. As long as default is the head ref for the tag, the release will be marked as latest. This is the case today; this code does not change that.

Fixes #1806.

- **Determine the branch name of the event's head ref. Set make_latest to the value of whether the branch name is default or not. If we're not pushing this tag from default, make-latest will be set to false.**
- **Do not build docs for backported releases. Add comments**

### DIFF
--- a/provider-ci/test-providers/acme/.github/workflows/main.yml
+++ b/provider-ci/test-providers/acme/.github/workflows/main.yml
@@ -76,6 +76,7 @@ jobs:
     with:
       version: ${{ needs.prerequisites.outputs.version }}
       isPrerelease: true
+      setLatestRelease: false
       skipGoSdk: true
       skipJavaSdk: true
 
--- a/provider-ci/test-providers/acme/.github/workflows/prerelease.yml
+++ b/provider-ci/test-providers/acme/.github/workflows/prerelease.yml
@@ -59,6 +59,7 @@ jobs:
     with:
       version: ${{ needs.prerequisites.outputs.version }}
       isPrerelease: true
+      setLatestRelease: false
 
   test:
     uses: ./.github/workflows/test.yml
--- a/provider-ci/test-providers/acme/.github/workflows/publish.yml
+++ b/provider-ci/test-providers/acme/.github/workflows/publish.yml
@@ -7,6 +7,9 @@ on:
       version:
         required: true
         type: string
+      setLatestRelease:
+        required: true
+        type: boolean
       isPrerelease:
         required: true
         type: boolean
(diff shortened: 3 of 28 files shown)


pose (approved):
Thanks for driving this.

provider-ci/internal/pkg/templates/base/.github/workflows/release.yml
```diff
@@ -76,6 +76,8 @@ jobs:
     with:
       version: ${{ needs.prerequisites.outputs.version }}
       isPrerelease: false
+      # Only tags on the default branch should set the release as `latest`; backported releases (tagged on a feature branch) should not.
+      setLatestRelease: ${{ github.event.base_ref == format('refs/heads/{0}', github.event.repository.default_branch) }}
```
corymhall: I think `github.event.base_ref` will be empty for a push->tag event, so that expression will be false and won't mark default-branch tags as latest.

For a push-of-a-tag workflow I think we need to detect whether the tag's commit is contained in the default branch, maybe through a separate step that tests the ancestry (via `merge-base --is-ancestor`).

Maybe something like this

```yaml
- name: Checkout tag
  uses: actions/checkout@v4
  with:
    ref: ${{ github.ref }}
    fetch-depth: 1

- name: Fetch default branch
  run: git fetch origin ${{ github.event.repository.default_branch }} --depth=1

- name: Check whether tag commit is on default branch
  id: is_default
  run: |
    if git merge-base --is-ancestor ${{ github.sha }} origin/${{ github.event.repository.default_branch }}; then
      echo "is_default=true" >> $GITHUB_OUTPUT
    else
      echo "is_default=false" >> $GITHUB_OUTPUT
    fi

# ...later, where you call the release action:
with:
  version: ${{ needs.prerequisites.outputs.version }}
  isPrerelease: false
  # Only tags on the default branch should set the release as `latest`; backported releases (tagged on a feature branch) should not.
  setLatestRelease: ${{ steps.is_default.outputs.is_default == 'true' }}

```

corymhall (approved):

Issue #1810, opened 2025-11-05
Run mise for native providers


blampe:
For parity with bridged.

PR #1811, opened 2025-11-05
mise for native providers

blampe:
* Extract duplicated steps into download-provider, download-sdk, and setup-tool composite actions.
* As part of setup-tools, use jdx/mise-action as we do for bridged providers. We preserve the behavior where caching is only enabled during prerequisites.

The actions are all still part of the "native" template but they're nearly identical to what we use for bridged, making it easier to converge these later.

Testing here:
* https://github.com/pulumi/pulumi-docker-build/pull/651
* https://github.com/pulumi/pulumi-kubernetes/pull/3929 (has flaking tests)

Fixes https://github.com/pulumi/ci-mgmt/issues/1810
Fixes https://github.com/pulumi/pulumi-docker-build/issues/636

### DIFF
--- a/provider-ci/internal/pkg/templates/native/.github/workflows/build.yml
+++ b/provider-ci/internal/pkg/templates/native/.github/workflows/build.yml
@@ -498,19 +498,19 @@ jobs:
         token: ${{ secrets.GITHUB_TOKEN }}
     #{{- if eq .Config.Provider "kubernetes-cert-manager" }}#
     - name: Setup KinD cluster
-      uses: helm/kind-action@a1b0e391336a6ee6713a0583f8c6240d70863de3 # v1.12.0
+      uses: helm/kind-action@92086f6be054225fa813e0a4b13787fc9088faab # v1.13.0
       with:
         cluster_name: kind-integration-tests-${{ matrix.language }}
         node_image: kindest/node:v1.29.2
     #{{- else if eq .Config.Provider "kubernetes-coredns" }}#
     - name: Setup KinD cluster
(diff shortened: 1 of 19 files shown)


provider-ci/internal/pkg/templates/native/.github/actions/download-sdk/action.yml
```diff
@@ -0,0 +1,20 @@
...
+  steps:
+    - name: Download SDK
+      uses: actions/download-artifact@018cc2cf5baa6db3ef3c5f8a56943fffe632ef53 # v6.0.0
+      with:
+        name: ${{ inputs.language }}-sdk.tar.gz
+        path: ${{ github.workspace}}/sdk/
+
+    - name: UnTar SDK folder
+      shell: bash
+      run: tar -zxf ${{ github.workspace}}/sdk/${{ inputs.language}}.tar.gz -C ${{ github.workspace}}/sdk/${{ inputs.language}}
```
pose: (nitpick) Is there any linter we could use? This irks my blood:

```
...${{ inputs.language }}...
```

vs

```
...${{ github.workspace}}/sdk/...
```

Notice the missing final whitespace on the ${{ }} expression

provider-ci/internal/pkg/templates/native/.github/workflows/build.yml
```diff
@@ -44,24 +44,10 @@ jobs:
         set-env: PROVIDER_VERSION
```
pose: Sweet! 👏

pose:
@blampe This test is failing: https://github.com/pulumi/ci-mgmt/actions/runs/19231363612/job/54970579726 and preventing the merge. Seems like it's not generating any new/modified files (which is expecting) and that's skipping the step.

PR #1822, opened 2025-11-11
Switching to new SDK provider tool for makefiles and adding to mise

Graham-Pedersen:
This change is part of our refactoring of how SDK generation works for providers. It alters the makefile targets to use the tool, and adds the new tool to mise. 

This change is ready to be merged and rolled out to the bridged providers

As part of this rollout we are also going to migrate overlays into a new package structure. This has been done for AWS, and I have identified the other 3 packages that need that change (pulumi-gcp, pulumi-openstack, and pulumi-azure). In addition each bridged provider will need another commit to remove the 4 SDKs that no longer need to be included in source

### DIFF
--- a/.github/workflows/update-workflows.yml
+++ b/.github/workflows/update-workflows.yml
@@ -77,13 +77,14 @@ jobs:
           rm pulumi-${{ inputs.provider_name }}/.github/workflows/cleanup-artifacts.yml || echo "not found"
           rm pulumi-${{ inputs.provider_name }}/.github/workflows/pull_request.yml || echo "not found"
           rm pulumi-${{ inputs.provider_name }}/.github/workflows/master.yml || echo "not found"
-      - name: Install mise
-        uses: jdx/mise-action@146a28175021df8ca24f8ee1828cc2a60f980bd5 # v3
+      - name: Setup mise
+        uses: jdx/mise-action@v3
+        env:
+          MISE_FETCH_REMOTE_VERSIONS_TIMEOUT: 30s
         with:
-          # Latest working version. See https://github.com/jdx/mise/discussions/6781
-          version: 2025.10.16
-          cache: false
-          working_directory: pulumi-${{ inputs.provider_name }}
+          version: 2025.11.6
+          github_token: ${{ steps.esc-secrets.outputs.PULUMI_BOT_TOKEN }}
+          cache_save: false
       - name: Generate workflow files into pulumi-${{ inputs.provider_name }}
         run: |
           cd ci-mgmt/provider-ci && go run ./... generate \


provider-ci/internal/pkg/migrations/migrate_cimgmt_to_mise.go
```diff
@@ -69,16 +71,15 @@ func (migrateCimgmtToMise) Migrate(templateName, outDir string) error {
...
-		}
-		return nil
-	}
 
 	// If we have any plugins overrides, move them to the .config/mise.toml
 	pluginList := nodeToPluginEntries(plugins)
 	entries := pluginsToToolEntries(pluginList)
+
+	//Manually adding the new SDK tool to this list until we support dynamically adding new dependencies
+	sdkToolPlugin := sectionEntry{"go:github.com/pulumi/provider-sdk-builder", "latest"}
```
corymhall: long term we'll probably want to switch to using a binary which is published on the GitHub release.

```
"github:pulumi/provider-sdk-builder" = "latest"
```

provider-ci/internal/pkg/templates/base/Makefile
```diff
@@ -77,21 +77,18 @@ mise_env:
...
 prepare_local_workspace: .make/mise_install upstream
 prepare_local_workspace: | mise_env
-# Creates all generated files which need to be committed
-generate: generate_sdks schema#{{- if .Config.RegistryDocs }}# build_registry_docs#{{- end }}#
-generate_sdks:#{{ range .Config.Languages }}# generate_#{{ . }}##{{ end }}##{{- if .Config.RegistryDocs }}# build_registry_docs#{{- end }}#
-build_sdks:#{{ range .Config.Languages }}# build_#{{ . }}##{{ end }}##{{- if .Config.RegistryDocs }}# build_registry_docs#{{- end }}#
-install_sdks:#{{ range .Config.Languages }}# install_#{{ . }}#_sdk#{{ end }}#
-.PHONY: development only_build build generate generate_sdks build_sdks install_sdks mise_install mise_env
+# Creates all generated files of which the schema is committed
+generate: build_sdks schema#{{- if .Config.RegistryDocs }}# build_registry_docs#{{- end }}#
```
corymhall: Are we still generating the SDKs? If so I think we should still separate out the generate from build. Building the sdks can take a while and it's nice to not have to run it if you don't have to.

provider-ci/internal/pkg/templates/base/Makefile
```diff
@@ -121,11 +116,6 @@ help:
 	@echo "  mise_install             Install tools with mise"
 	@echo "  upstream                 Initialize the upstream submodule, if present"
 	@echo ""
-	@echo "Language-Specific Targets"
-	@echo "  generate_[language]    Generate the SDK files ready for committing"
-	@echo "  build_[language]       Build the SDK to check correctness"
-	@echo "  install_[language]_sdk Install the SDK ready for testing"
-	@echo ""
```
corymhall: Don't we still need all these targets? If I'm running locally and just want the nodejs SDKs installed I can just run the language specific targets.

provider-ci/internal/pkg/migrations/migrate_cimgmt_to_mise.go
Graham-Pedersen: We can do that. I would just need to get the binaries getting built and included on the release. Any reason we prefer the binaries to go installation though?

provider-ci/internal/pkg/templates/base/Makefile
```diff
@@ -77,21 +77,18 @@ mise_env:
 	@touch $@
 
 # Build the provider and all SDKs and install ready for testing
-build: .make/mise_install provider build_sdks install_sdks#{{- if .Config.RegistryDocs }}# build_registry_docs#{{- end }}#
```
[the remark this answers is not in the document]
Graham-Pedersen: In the latest version I actually have changed this. You can pass in an SDK to build. It is run as part of the build in root though also

provider-ci/internal/pkg/templates/base/Makefile
```diff
@@ -229,8 +134,14 @@ build_registry_docs: .make/build_registry_docs
...
 #{{- end }}#
 
+build_sdks: .make/build_sdks
+.make/build_sdks: .make/schema
+	provider-sdk-builder build-sdks --language all --providerName $(PACK)
+	@touch $@
+.PHONY: build_sdks
+
 clean:
-	rm -rf sdk/{dotnet,nodejs,go,python}
```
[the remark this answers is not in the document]
Graham-Pedersen: So we use to not ever delete the Java SDK for some reason. It is now safe to delete all of them.

Also funny story, there is a big bug in this make clean target because all the overlay files used to be hand written and put in the SDK folders so this clean target was actually very unsafe for providers with overlays!

.github/workflows/update-workflows.yml
```diff
@@ -97,7 +93,7 @@ jobs:
         with:
           # Latest working version. See https://github.com/jdx/mise/discussions/6781
           version: 2025.10.16
-          install: false
+          install: true
```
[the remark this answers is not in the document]
blampe: Isn't `install: true` the default, so we can omit this?

Since we're no longer using lockfiles it sounds like this is OK?

provider-ci/internal/pkg/migrations/migrate_cimgmt_to_mise.go
```diff
@@ -67,6 +67,11 @@ func (migrateCimgmtToMise) Migrate(templateName, outDir string) error {
...
+	entries = append(entries, sdkToolPlugin)
```
blampe: I don't think you need this, the `mise.toml` in this directory should already capture the default.

provider-ci/internal/pkg/templates/base/.github/workflows/verify-release.yml
```diff
@@ -76,6 +77,10 @@ jobs:
       #{{- .Config | renderEscStep | indent 6 }}#
       #{{- range .Config.Languages }}#
       #{{- if eq . "python" }}#
+      - name: Download python SDK
+        uses: ./.github/actions/download-sdk
+        with:
+          language: python
```
blampe: This seems counter to what the workflow is trying to do, and if we needed to add this step this might point towards a bug somewhere.

In particular I think this workflow is trying to test the end-to-end experience from the user's perspective, which means that it expects to find all the relevant SDKs and binaries published and available from the usual places.

What we're doing here is grabbing an SDK that was generated earlier. Why weren't we able to grab the SDK from npm, pypi, etc.?

provider-ci/internal/pkg/migrations/migrate_cimgmt_to_mise.go
Graham-Pedersen: I was testing off an old branch of this package, but I know we did need it for a time. Repos that already had a mise.toml file to add it to the existing ones. 

Let me rip it out and check when I retest.

Graham-Pedersen: I know this file has changed a lot since I wrote that code, because I had like 10 sets of merge conflicts in here when I was rebasing my changes onto mainline haha

provider-ci/internal/pkg/templates/base/.github/workflows/test.yml
```diff
@@ -152,15 +152,15 @@ jobs:
...
 #{{- end }}#
 #{{- if not .Config.Shards }}#
-    - name: Install dependencies
-      run: make install_${{ matrix.language}}_sdk
 #{{- if .Config.Actions.PreTest }}#
 #{{ .Config.Actions.PreTest | toYaml | indent 4 }}#
 #{{- end }}#
     #{{- if .Config.IntegrationTestProvider }}#
+    - name: Install dependencies
+      run: provider-sdk-builder install  --sdkLocation ./sdk/${{ matrix.language}} -l ${{ matrix.language}} --installLocation ${{ github.workspace }}
```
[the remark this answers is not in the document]
Graham-Pedersen: Happy to make those changes, and yes that was probably not the correct name

provider-ci/internal/pkg/templates/base/.github/workflows/test.yml
```diff
@@ -152,15 +152,15 @@ jobs:
     #{{- end }}#
 #{{- if and .Config.Shards (not .Config.NoSchema) }}#
     - name: Install prebuilt SDKs
-      run: make install_sdks
+      run: provider-sdk-builder install
 #{{- end }}#
 #{{- if not .Config.Shards }}#
-    - name: Install dependencies
-      run: make install_${{ matrix.language}}_sdk
```
Graham-Pedersen: I need to add this back in but update it to the latest way. Next version will have a change. This is why the XYZ provider fails to build

provider-ci/internal/pkg/templates/base/.github/workflows/test.yml
```diff
@@ -83,7 +83,7 @@ jobs:
       with:
         language: ${{ matrix.language }}
     - name: Restore makefile progress
-      run: make --touch provider schema build_${{ matrix.language }}
+      run: make --touch provider schema build_sdks SDK_LANG=${{ matrix.language }}
```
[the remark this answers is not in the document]
corymhall: Yeah this is using `--touch` so it's not actually building. We do this so that later when we run `make test` it thinks (correctly) that the SDKs have already been built. Without this `make test` will re-build the SDKs.

provider-ci/internal/pkg/templates/base/Makefile
corymhall: I think I'm more suggesting that when iterating I might just want to regenerate the SDKs and _not_ build the SDKs. But that's probably pretty uncommon.

provider-ci/test-providers/acme/.github/actions/download-sdk/action.yml
```diff
@@ -14,6 +14,13 @@ runs:
       with:
         name: ${{ inputs.language }}-sdk.tar.gz
         path: ${{ github.workspace}}/sdk/
-    - name: Uncompress SDK folder
+    - name: Uncompress SDK folder (Unix)
+      if: runner.os != 'Windows'
       shell: bash
-      run: tar -zxf ${{ github.workspace }}/sdk/${{ inputs.language }}.tar.gz -C ${{ github.workspace }}/sdk/${{ inputs.language }}
+      run: mkdir -p ${{ github.workspace }}/sdk/${{ inputs.language }} && tar -zxf ${{ github.workspace }}/sdk/${{ inputs.language }}.tar.gz -C ${{ github.workspace }}/sdk/${{ inputs.language }}
+    - name: Uncompress SDK folder (Windows)
```
corymhall: Is this only needed because of verify-release?

Graham-Pedersen: It is used for more steps than just that. Several steps download the SDK. We needed to make this change because tar will fail if the destination folder does not exist without the flag I added. Since the SDK folders are not checked in this is needed for the verify and test workflows.

provider-ci/internal/pkg/templates/base/.github/workflows/verify-release.yml
Graham-Pedersen: I believe it was one of them yes

provider-ci/internal/pkg/migrations/migrate_cimgmt_to_mise.go
blampe: golangci-lint is a bit of an odd duck and that guidance isn't true in general. Sticking with `go install` should be fine for something this simple.

provider-ci/test-providers/acme/.github/actions/download-sdk/action.yml
corymhall: Per the other comment we don't want to download SDKs in the verify-release job since we want to use the published SDKs.

Graham-Pedersen: Would you mind if I tackled that at a different time when I go after the verify-release job?

provider-ci/internal/pkg/templates/base/Makefile
```diff
@@ -227,8 +134,27 @@ build_registry_docs: .make/build_registry_docs
...
 
+build_sdks: .make/build_sdks
+.make/build_sdks: .make/schema
+	provider-sdk-builder build-sdks --providerName $(PACK) --language $(SDK_LANG) --version $(PROVIDER_VERSION)
+	@touch $@
+.PHONY: build_sdks
+
+install_sdks: .make/install_sdks
+.make/install_sdks: .make/build_sdks
+	provider-sdk-builder build-sdks --language $(SDK_LANG)
```
EronWright: Does `provider-sdk-builder build-sdks` install too? This seems too similar to `.make/build_sdks`.

provider-ci/internal/pkg/templates/base/.github/workflows/publish.yml
```diff
@@ -160,7 +160,7 @@ jobs:
         registry-url: https://registry.npmjs.org
     - name: Publish SDKs
       if: inputs.skipJavaSdk == false
-      uses: pulumi/pulumi-package-publisher@c1672c7928591d563dccb12729e05e315c21f8c2 # v0.0.22
+      uses: pulumi/pulumi-package-publisher@3ec1409d3e894142b9825c7859be8e57d362762a
```
EronWright: nit: can we keep the `# v0.0.22` comment? (did a tool make this change?)

provider-ci/internal/pkg/templates/base/.github/workflows/verify-release.yml
```diff
@@ -55,7 +55,8 @@ jobs:
     strategy:
       matrix:
 #{{- if .Config.ReleaseVerification }}#
-        runner: ["ubuntu-latest", "windows-latest", "macos-latest"]
+        runner: ["ubuntu-latest", "macos-latest"] 
+        #TODO add "windows-latest" back in to this list. https://github.com/pulumi/ci-mgmt/issues/1827?issue=pulumi%7Cci-mgmt%7C1857
```
EronWright: Is this TODO for now or later?

provider-ci/internal/pkg/templates/base/.github/workflows/publish.yml
Graham-Pedersen: A tool made that change. When I compile the ci-mgmt it does that

provider-ci/internal/pkg/templates/base/.github/workflows/verify-release.yml
Graham-Pedersen: It is a TODO for later, but hopefully not much later. I cut a github issue in the epic for the SDK rollout

provider-ci/test-providers/acme/.github/workflows/build_sdk.yml
```diff
@@ -70,33 +70,21 @@ jobs:
...
-            sdk/go/**/pulumiUtilities.go
-            sdk/nodejs/package.json
-            sdk/python/pyproject.toml
-            sdk/java/build.gradle
       - name: Commit ${{ matrix.language }} SDK changes for Renovate
         # If the worktree is dirty and this is a Renovate PR to bump
         # dependencies, commit the updated SDK and push it back to the PR. The
         # job will still be marked as a failure.
-        if: failure() && steps.worktreeClean.outcome == 'failure' && contains(github.actor, 'renovate') && github.event_name == 'pull_request'
+        if: failure() && contains(github.actor, 'renovate') && github.event_name == 'pull_request'
```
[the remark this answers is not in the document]
Graham-Pedersen: Uhh no it was not. My intent was to remove the worktree clean step. Very good catch let me rework it.

Graham-Pedersen:
Also enough people have asked for a compile target I will add that back in

provider-ci/test-providers/acme/.github/workflows/build_sdk.yml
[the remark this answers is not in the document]
Graham-Pedersen: I have just ripped this whole step of the workflow out

provider-ci/internal/pkg/migrations/migrate_cimgmt_to_mise.go
```diff
@@ -67,6 +67,7 @@ func (migrateCimgmtToMise) Migrate(templateName, outDir string) error {
 	// If we have any plugins overrides, move them to the .config/mise.toml
 	pluginList := nodeToPluginEntries(plugins)
 	entries := pluginsToToolEntries(pluginList)
+
```
guineveresaenger: nit: surprise blank line

provider-ci/test-providers/acme/.github/actions/download-sdk/action.yml
Graham-Pedersen: Sounds good

provider-ci/internal/pkg/templates/base/.github/workflows/verify-release.yml
corymhall: I think that is just a side-effect of how the setup-java action works. It must expect a local java project in certain cases and it was a coincidence that we had the SDKs checked out. I _think_ it might be because the setup-java action tries to restore the gradle cache and looks for local projects. Maybe we can just turn that off or trick it somehow?

provider-ci/internal/pkg/templates/base/.github/workflows/test.yml
EronWright: Is this still outstanding?

provider-ci/internal/pkg/templates/base/Makefile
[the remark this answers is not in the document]
EronWright: Can you speak to this?  Does build include install or not?

PR #1838, opened 2025-11-17
Move back `build` target to top of Makefile for `base`

### LINKED ISSUE #1825: Move back `build` target to top of Makefile
Run `make` on `eks`, `awsx`, or `aws-apigateway`.

## What should happen?

It should invoke the `build` target as described by the `make help` docs:

```
Usage: make [target]

Main Targets
  build (default)     Build the provider and all SDKs and install for testing
  generate            Generate all SDKs, documentation and schema
  provider            Build the local provider binary
```

## What happens instead?

It runs the `mise_install` and shows not output.

---

## Root cause

From the [Makefile manual](https://www.gnu.org/software/make/manual/html_node/Goals.html):

> By default, the goal is the first target in the makefile (not counting targets that start with a period). Therefore, makefiles are usually written so that the first target is for compiling the entire program or programs they describe. 

When we [introduced `mise_install` target on top of `base/Makefile`](https://github.com/pulumi/ci-mgmt/blob/master/provider-ci/internal/pkg/templates/base/Makefile#L62-L74), we replaced the `build` target as goal.

pose:
Fixes #1825

From the Makefile manual:

> By default, the goal is the first target in the makefile (not counting
targets that start with a period). Therefore, makefiles are usually written so that the first target is for compiling the entire program or programs they describe.

When we introduced mise_install target on top of base/Makefile, we replaced the build target as goal.

## Testing

Checked out `test-providers` and verified they worked as before the `mise` change:
* [x] acme
* [x] aws
* [x] cloudflare
* [x] docker
* [x] eks
* [x] xyz

### DIFF
--- a/provider-ci/internal/pkg/templates/base/Makefile
+++ b/provider-ci/internal/pkg/templates/base/Makefile
@@ -59,23 +59,10 @@ LDFLAGS=$(LDFLAGS_PROJ_VERSION) $(LDFLAGS_UPSTREAM_VERSION) $(LDFLAGS_EXTRAS) $(
 # Ensure all directories exist before evaluating targets to avoid issues with `touch` creating directories.
 _ := $(shell mkdir -p .make bin .pulumi/bin)
 
-# Installs all necessary tools with mise and records completion in a sentinel
-# file so dependent targets can participate in make's caching behaviour. The
-# environment is refreshed via an order-only prerequisite so it still runs on
-# every invocation without invalidating the sentinel.
-mise_install: .make/mise_install | mise_env
-
-.PHONY: mise_env
-mise_env:
-	@mise env -q  > /dev/null
-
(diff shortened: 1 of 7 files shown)


blampe (approved):
Makes sense -- we changed the behavior of `make` with no args.

pose:
> This sounds reasonable to me, but what are the current repercussions of this?

Hey, I am doing this change is to comply with what we are documenting on `make help`. If you run `make help` on on of the providers we manage with `mise` (like `eks`), we print this message:

```
> $ make help
Usage: make [target]

Main Targets
  build (default)     Build the provider and all SDKs and install for testing
  generate            Generate all SDKs, documentation and schema
  provider            Build the local provider binary
```

Let me know if that's what you were referring to or if I missed anything else 🙏

PR #1848, opened 2025-11-18
mise: Stop using lockfiles

blampe:
mise: Stop using lockfiles

(Resubmitting #1846 as I had some git trouble while setting up the stack.)

Lock files are currently experimental and have been problematic for us. In
particular "latest" still seems to resolve inconsistently, and older lock files
can cause panics due to some breaking changes.

Let's remove lock files and specify explicit versions in mise.toml for now.

* Update the migration to remove mise.lock instead of generating it.
* Update mise to latest since https://github.com/jdx/mise/discussions/6781 is no longer relevant.
* Update default mise.toml to specify exact versions instead of latest.
* Remove mise.lock from list of allowed-changes on SDK gen.


blampe:

- #1848 ◀
    - #1847

Issue #1727, opened 2025-09-15
[Epic]: Migrate tool installation/setup to mise


corymhall:
This is a tracking ticket for the work needed to migrate all of our tool setup to fully use [mise](https://mise.jdx.dev/).

[Design doc.](https://docs.google.com/document/d/1wMi3ZujTQ7E4uYTVaTuk2WO6FTHsApmj4vICl3RSs5c/edit?pli=1&tab=t.0#heading=h.4k906iv6294)

PR #1850, opened 2025-11-18
mise: Use mise in verify-release

blampe:
After this change all of our workflows should be using mise consistently.

Refs https://github.com/pulumi/ci-mgmt/issues/1727.


guineveresaenger:
Example failure in verify-release:

https://github.com/pulumi/pulumi-aws/issues/6040

Issue #591, opened 2023-09-18
Make Gradle version configurable


guineveresaenger:
We should make the Gradle version configurable, just like the javaGenVersion.

Currently we [hardcode it in the templates](https://github.com/pulumi/ci-mgmt/blob/master/provider-ci/internal/pkg/templates/bridged-provider/.github/workflows/run-acceptance-tests.yml#L57), which doesn't allow for overrides or updates to individual providers.

EronWright:
Re-upping this since the ubuntu-latest image was recently bumped from Gradle 8 to 9, and that caused an SDK compilation error.  The templates still use 7.6, and this error suggests some efforts are needed soon.

https://github.com/pulumi/pulumi-kubernetes/actions/runs/16949066809/job/48040265267#step:10:3395

```
pulumi package gen-sdk --language java provider/cmd/pulumi-resource-kubernetes/schema.json \
	--overlays provider/pkg/gen/java-templates --out sdk
SDK has been written to sdk/java
cd sdk/java/ && \
	echo "module fake_java_module // Exclude this directory from Go tools\n\ngo 1.17" > go.mod && \
	gradle --console=plain build

Welcome to Gradle 9.0.0!

Here are the highlights of this release:
 - Configuration Cache is the recommended execution mode
 - Gradle requires JVM 17 or higher to run
 - Build scripts use Kotlin 2.2 and Groovy 4.0
 - Improved Kotlin DSL script compilation avoidance

For more details see https://docs.gradle.org/9.0.0/release-notes.html

Starting a Gradle Daemon (subsequent builds will be faster)


[Incubating] Problems report is available at: file:///home/runner/work/pulumi-kubernetes/pulumi-kubernetes/sdk/java/build/reports/problems/problems-report.html
FAILURE: Build failed with an exception.


Deprecated Gradle features were used in this build, making it incompatible with Gradle 10.
* What went wrong:

Configuring project ':lib' without an existing directory is not allowed. The configured projectDirectory '/home/runner/work/pulumi-kubernetes/pulumi-kubernetes/sdk/java/lib' does not exist, can't be written to or is not a directory.

* Try:
> Make sure the project directory exists and is writable.
> Run with --scan to generate a Build Scan (Powered by Develocity).

BUILD FAILED in 11s
You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

[...]

PR #1868, opened 2025-11-24
mise: Update migration logic to maintain `.config/mise.toml`

blampe:
The migration responsible for establishing the `.config/mise.toml` file was only running once, because it was moving plugins from `.ci-mgmt.yaml` into `mise.toml`. Running the migration again wouldn't see any plugins left in `.ci-mgmt.yaml`.

This is problematic because it means changes made to our default `mise.toml` aren't getting reflected in downstream projects after they've onboarded. For example in https://github.com/pulumi/pulumi-databricks/pull/955 it's using `gradle = '7.6'` but that's broken! We want it to use 7.6.0 (https://github.com/pulumi/ci-mgmt/pull/1835).

This PR updates our logic to extract plugins from any existing `mise.toml` and then merge them back into the default configuration.

Refs https://github.com/pulumi/ci-mgmt/issues/1727

### DIFF
--- a/provider-ci/internal/pkg/migrations/mise.toml
+++ b/provider-ci/internal/pkg/migrations/mise.toml
@@ -20,7 +20,7 @@ java = 'corretto-11'
 pulumi = "{{ get_env(name='PULUMI_VERSION_MISE', default='latest') }}"
 "github:pulumi/pulumictl" = '0.0.50'
 "github:pulumi/schema-tools" = "0.6.0"
-gradle = '7.6.0'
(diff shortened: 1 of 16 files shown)


guineveresaenger (approved):
Let's try it!

To recap from our offsides discussion - we should probably have the `mise.toml` be a template, so we can have at-a-glance solutions for e.g. https://github.com/pulumi/ci-mgmt/issues/591

(full disclosure I thought it already _was_ a template until seeing this PR!)

PR #1871, opened 2025-11-25
mise: Install tools when updating workflows

blampe:
We should have the provider's toolchain available when generating workflows to more closely match the behavior we expect to see (i.e. when running locally).

The version of golangci-lint is used to determine whether you get v1 or v2 lint config, for example.


guineveresaenger (approved):
Seems to fix pulumi-gitlab.

PR #1879, opened 2025-12-01
enable journaling in ci-mgmt repos

tgummerer:
This is the next step for the rollout plan described in https://github.com/pulumi/home/issues/4404. We probably want to wait to merge this until
https://github.com/pulumi/pulumi-service/issues/34813 has a resolution, but I wanted to at least get this ready for review in the meantime.

/cc @pgavlin


blampe (approved):
Functionality is only for testing and isn't user-facing so this seems good to roll out broadly.

PR #1881, opened 2025-12-01
Update aws-native nightly-sdk-generation to use mise

corymhall:
fixes https://github.com/pulumi/pulumi-aws-native/issues/2601

Ran the workflow update for aws-native and manually triggered the nightly workflow to confirm it works https://github.com/pulumi/pulumi-aws-native/actions/runs/19826814292/job/56802222791


blampe:
@corymhall @pose FYI nowadays it's possible for the provider to define some custom workflows that don't get overridden by ci-mgmt. Basically rename this file to `aws-native-nightly-sdk-generation.yml` and it will persist. That way you can make changes like this directly in the repo without having to go through this extra ci-mgmt hoop.

Issue #1869, opened 2025-11-24
Move mise plugins to top-level mise.toml


blampe:
Followup to #1474. @guineveresaenger pointed out on #1868 that we could get simpler automatic bumping of `.config/mise.toml` by moving provider plugins to the top-level `mise.toml`, since they are optional and something provider authors need to explicitly specify anyway.

In other words:
- Update the migration to move any vfox-pulumi plugins from `.config/mise.toml` to `mise.toml`.
- Make `.config/mise.toml` a normal file that ci-mgmt ships as part of the template.

That gets ci-mgmt out of the game of managing provider plugins, while still allowing it to ship reasonable defaults. We should be able to rely on something like Renovate or Dependabot to handle bumping plugin versions.

PR #1906, opened 2025-12-09
mise: Move pulumi plugins to root mise.toml

blampe:
Right now we distribute a default `.config/mise.toml` but then apply a
migration in order to preserve a provider's custom pulumi plugin requirements,
if it has any.

Instead, we would like `.config/mise.toml` to be distributed (and templated)
the same as all the other static files. The root `./mise.toml` remains the
customization point for providers, so these custom plugin requirements are more
appropriately handled there.

This PR adds a migration to move vfox-pulumi tools (pulumi plugins) from
`.config/mise.toml` to `./mise.toml`. Once this migration has been applied
everywhere we can remove it and distribute the default `./config/mise.toml`
normally.

Refs https://github.com/pulumi/ci-mgmt/issues/1869.

### DIFF
--- a/provider-ci/internal/pkg/migrations/activate_mise.go
+++ b/provider-ci/internal/pkg/migrations/activate_mise.go
@@ -1,7 +1,6 @@
 package migrations
 
 import (
-	"bytes"
 	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


.github/workflows/test-provider-ci.yml
```diff
@@ -2,8 +2,6 @@ permissions: write-all # Equivalent to default permissions plus id-token: write
 name: Test Provider CI
 on:
   pull_request:
-    branches:
-      - master
```
pose: Is this intended?

pose (approved):

.github/workflows/test-provider-ci.yml
blampe: Ah, yes intentional. This lets tests run on stacked PRs.

PR #1852, opened 2025-11-18
mise: Use an env plugin

blampe:
We currently set env vars using `_.source` but this doesn't work on Windows.

[Environment plugins](https://mise.jdx.dev/env-plugin-development.html) do work on Windows.

This PR migrates our `get-versions.sh` script to an environment plugin / Lua script. This will enable us to use mise during verify-release.

Refs https://github.com/pulumi/ci-mgmt/issues/1742.

PR #1907, opened 2025-12-09
mise: Distribute `.config/mise.toml` as a normal file

### LINKED ISSUE #1869: Move mise plugins to top-level mise.toml

blampe:
This consolidates our pulumi plugin migrations and makes `.config/mise.toml` a
normal file. Importantly, this means we will apply our usual templating to the
file so we can now parameterize it with `${{ .Config }}`.

Migrations are consolidated so we now have a single migration responsible for
moving pulumi plugins from `.ci-mgmt.yaml` _or_ `./config/mise.toml` to the
root `mise.toml`. This should allow migrations to continue onboarding existing
third-party provider.

This is the second half of #1869 and should be merged once #1906 has gone out
broadly.

Fixes #1869.

### DIFF
--- a/provider-ci/internal/pkg/migrations/merge_plugins_to_root_test.go
+++ b/provider-ci/internal/pkg/migrations/merge_plugins_to_root_test.go
@@ -292,12 +292,6 @@ go = "latest"
 	if !strings.Contains(contentStr, `"vfox-pulumi:pulumi/pulumi-bridge-eks" = "1.2.3"`) {
 		t.Fatalf("expected pulumi-bridge-eks tool from .ci-mgmt.yaml in root mise.toml, got:\n%s", contentStr)
 	}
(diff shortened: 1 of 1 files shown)


blampe:

- #1906
    - #1907 ◀
        - #1852
            - #1850

PR #1913, opened 2025-12-10
Update workflows using GitHub app auth

blampe:
I've created a GitHub app (Pulumi Provider Automation) and added its ID/key to the `pulumi/imports/github-secrets` environment.

This PR starts using it in ci-mgmt's workflows where we saw some throttling of the pulumi-bot user.

I'm also adding plumbing so we can hook this up in provider jobs as well. If pulumi-bot is bumping into rate limits then this is something we'll eventually need to do anyway.

When we do start using this in other places, one thing to watch out for is a lot of our automation special-cases the pulumi-bot user specifically. With this app token the user will be different, so we need to account for that.

Fixes https://github.com/pulumi/ci-mgmt/issues/1910.

### DIFF
--- a/.github/workflows/update-workflows.yml
+++ b/.github/workflows/update-workflows.yml
@@ -127,5 +127,7 @@ jobs:
         # `continue-on-error: true`.
         continue-on-error: true
         run: "gh pr merge --auto --squash ${{ steps.create-pr.outputs.pull-request-url }}"
+        env:
+          GH_TOKEN: ${{ steps.app-auth.outputs.token }}
     outputs:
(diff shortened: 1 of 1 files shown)


provider-ci/internal/pkg/templates/native/.goreleaser.yml
```diff
@@ -1,5 +1,4 @@
 # WARNING: This file is autogenerated - changes will be overwritten if not made via https://github.com/pulumi/ci-mgmt
-
```
blampe: No-op change to confirm we can actually deploy the downstream changes.

PR #1927, opened 2025-12-16
Flag upgrade-provider build failures

blampe:
We have a visibility gap right now around upgrade-provider failures. These failures get flagged in issues like [this one](https://github.com/pulumi/pulumi-azure/issues/3367), but not in a way that surfaces them to our dashboards. For example I can't tell right now if this workflow is broken for _all_ providers or just azure.

This change causes the upgrade-provider workflow to fail if we weren't able to create a PR. This will get the failure picked up by the same automation we use for all our other workflow failures: an issue will be created and visible on our dashboards, and the issue will auto-resolve when the problem is fixed.

### DIFF
--- a/provider-ci/internal/pkg/templates/bridged/.github/workflows/upgrade-provider.yml
+++ b/provider-ci/internal/pkg/templates/bridged/.github/workflows/upgrade-provider.yml
@@ -98,7 +98,6 @@ jobs:
       - name: Call upgrade provider action
         id: upgrade_provider
         if: steps.target_version.outputs.version != ''
-        continue-on-error: true
         uses: #{{ .Config.ActionVersions.UpgradeProviderAction }}#
         with:
           kind: provider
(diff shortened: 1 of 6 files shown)


corymhall (commented):
> These failures get flagged in issues like https://github.com/pulumi/pulumi-azure/issues/3367, but not in a way that surfaces them to our dashboards.

Should we instead try to surface these issues in our dashboards? My only concern is  having two issues to track the same thing. We also get those issues if the upgrade-provider workflow fails so I would vote on tracking failures in one or the other (create workflow failure issue, or the existing upgrade issue)

blampe:
> > These failures get flagged in issues like [pulumi/pulumi-azure#3367](https://github.com/pulumi/pulumi-azure/issues/3367), but not in a way that surfaces them to our dashboards.
> 
> Should we instead try to surface these issues in our dashboards? My only concern is having two issues to track the same thing. We also get those issues if the upgrade-provider workflow fails so I would vote on tracking failures in one or the other (create workflow failure issue, or the existing upgrade issue)

Yeah, I thought about alternatively trying to apply labels to these existing issues as well.

I don't want to get rid of the existing issues since they might have their own tracking around them, but I agree the "failed to create PR" messages are basically re-creating what we're already doing with build failures. They also complicate the workflow.

So I think what I'll do is get rid of that comment so the workflow is simpler and everything "just works" as expected.

PR #1941, opened 2025-12-18
weekly-pulumi-update: specify verison in go get

pgavlin:
These changes update the `go get` command used by `weekly-pulumi-update`
to include the version the upgrade is targeting. While this won't
prevent these jobs from failing if the versions of p/p/pkg and p/p/sdk
drift, it should help make that class of failures clearer.


blampe (approved):

pgavlin:
@blampe how do I update the testdata here?

blampe:
@pgavlin just run `make` at the repo root and it will re-generate and test everything.

PR #1952, opened 2025-12-19
mise: pin by default

blampe:
`mise use foo` should pin to the version fetched instead of using "latest" in mise.toml.


provider-ci/internal/pkg/templates/all/.config/mise.toml
```diff
@@ -28,6 +28,7 @@ golangci-lint = "1.64.8" # See note about about overrides if you need to customi
 experimental = true # Required for Go binaries (e.g. pulumictl).
 lockfile = false
 http_retries = 3
+pin = true # `mise use` should pin versions instead of defaulting to latest.
```
pose: Wow! This is something that should be a default.

PR #1959, opened 2025-12-23
Fix schema path

blampe:
I got this path wrong, so the SDK commit step is still broken.

```
git add sdk provider/cmd/docker-build/schema.json

...

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   provider/cmd/pulumi-resource-docker-build/schema.json
```

Noticed in https://github.com/pulumi/pulumi-docker-build/pull/717.


pose:
Nice! That explains an error I didn't understand on one of our logs

PR #1975, opened 2026-01-12
Fixes for external providers which have ESC & Github App disabled

ringods:
Fixes for external providers (tested by the `acme` test provider) which have ESC & Github App:

* Only use `id-token: write` permission when ESC is enabled.
* Disabled Github App in the `acme` test provider and regenerated files
* External providers use the Github secret named `PULUMI_PROVIDER_AUTOMATION_TOKEN`
* Drop `pull-requests: write` for lint & license check jobs.


ringods:
For now, I don't see an alternative in this glorified Go templating setup.

ringods:
Fixed in forked repo https://github.com/pulumi-labs/ci-mgmt/pull/1

PR #1976, opened 2026-01-12
[comment-on-stale-issues] Point back to aws-action instead of fork

pose:
The fork is no longer required since the fix has been merged upstream 🎉

Fixes https://github.com/pulumi/home#4448


blampe (approved):
Wonderful!

PR #1991, opened 2026-01-14
Fix more runner size configs

guineveresaenger:
It turns out there's a `publish` job inside the `build` Workflow. It also needs to be configureable.

Prerequisite for https://github.com/pulumi/pulumi-kubernetes/issues/4052.

- **Publish job in build.yml also needs configurable runners**
- **All providers should have ESC for prerelease prerequisites**
- **make test providers**

### DIFF
--- a/provider-ci/test-providers/command/.github/workflows/prerelease.yml
+++ b/provider-ci/test-providers/command/.github/workflows/prerelease.yml
@@ -23,6 +23,9 @@ jobs:
   prerequisites:
     runs-on: ubuntu-latest
     name: prerequisites
+    permissions:
+      id-token: write # For ESC secrets.
+      pull-requests: write # For schema check comment.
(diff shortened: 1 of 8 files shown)


provider-ci/internal/pkg/templates/native/.github/workflows/prerelease.yml
```diff
@@ -22,11 +22,9 @@ jobs:
   prerequisites:
     runs-on: #{{ if .Config.Runner.Prerequisites }}##{{ .Config.Runner.Prerequisites }}##{{ else }}#ubuntu-latest#{{ end }}#
     name: prerequisites
-    #{{- if eq .Config.Provider "aws-native" }}#
```
blampe: So odd... thanks for getting rid of this.

PR #2021, opened 2026-01-27
Setup claude github action

corymhall:
This PR does a couple of things:

### 1. Setup Claude

Created a [claude.yml](./provider-ci/internal/pkg/templates/internal/.github/workflows/claude.yml) workflow following the [claude docs](https://code.claude.com/docs/en/github-actions). This splits the functionality into two modes:

- *Review Mode*: On a PR if you comment with `@claude review` it will kick off the claude PR review task.
- *Work Mode*: On an issue _or_ PR if you comment with `@claude` (without "review") it will kick off claude to address your comment. This can be used to ask claude to work on an issue or address feedback on a PR.

For the initial roll out I chose not to auto trigger on PRs since we have a lot of repos. If we notice that we always request the review anyway we can revisit this.

### 2. Add CLAUDE.md and skills

I've only done this initially for bridged providers since I'm not sure native providers are generic enough to template this out.

- [.claude/CLAUDE.md](./provider-ci/internal/pkg/templates/internal-bridged/.claude/CLAUDE.md): Created this under `.claude` so that repos can add a root level CLAUDE.md with custom content.
- [provider-code-review](./provider-ci/internal/pkg/templates/internal-bridged/.claude/skills/provider-code-review): This includes specific instructions on performing code reviews for bridged providers. The idea is to be able to add specific instructions. Initially I have included specific instructions on reviewing changes to patches and resources.go (the `resources.go` is very bare at the moment).
- [upstream-patches](./provider-ci/internal/pkg/templates/internal-bridged/.claude/skills/upstream-patches): Skill that provides info on how to work with patches
- [pulumi-upgrade-provider](./provider-ci/internal/pkg/templates/internal-bridged/.claude/skills/pulumi-upgrade-provider): Skill for fixing issues with provider upgrades. I've been using this locally, but want to test and see if we can trigger `@claude` to fix it.


provider-ci/internal/pkg/templates/internal/.github/workflows/claude.yml
```diff
@@ -0,0 +1,88 @@
...
+  issues:
+    types: [opened]
+  pull_request_review:
+    types: [submitted]
+
+jobs:
+  claude:
+    # Only run when @claude is mentioned
+    if: |
+      (github.event_name == 'issue_comment' && contains(github.event.comment.body, '@claude')) ||
```
pose: Don't we need to put any guardrails around people just randomly running `@claude` on the public repos? What does `@claude` actually has access to?

Issue #1844, opened 2025-11-18
Move release and verify workflows to a merge queue


blampe:
We currently perform releases after merging to master. This is problematic because we can sometimes merge changes (especially to CI workflows) which break releases however we don't know that until after the change has already merged.

Instead, we should adopt a merge queue model. This would keep PRs as "light" as possible, and only execute the heavier prerelease logic when attempting to merge.

PR #2025, opened 2026-01-27
Allow repos to opt-in to merge queues

blampe:
This preserves our existing behavior while allowing repos to selectively enable merge queues for testing.

Currently:
* Tests run as part of PR.
* PR is merged.
* Push event publishes pre-release artifacts and re-runs tests.

A big problem with this is we only find out about failures well after the fact (e.g. Java publishing is broken, verify release doesn't work on Windows, etc.).

After this change, if a repo enables merge queues:
* Tests run as part of PR.
* Before a PR is merged, pre-release artifacts are published and tests are re-run.

This does a few things:
* It gives us stronger guarantees around commits on master.
  * If it was committed, we know it was released and its tests passed.
  * Semantic merge conflicts are no longer possible.
* Any release or test failures are now impossible to miss.
* We will have the ability to run slower integration tests only during the merge group for faster PR feedback.
* It also lets us merge more changes at the same time, but this isn't really an issue for providers.

Backward compatibility is achieved by ignoring the "github-merge-queue[bot]" actor on push events. This actor is responsible for pushing the result of the merge group back to the main branch, and we ignore these events because we know the checks already ran.

The release workflow remains unchanged & is still triggered by tags.

Refs #1844.


corymhall (approved):
What's the process to opt-in/out?

PR #2034, opened 2026-02-04
Customizable targets

blampe:
This lays groundwork for https://github.com/pulumi/ci-mgmt/issues/1131 by shipping some default targets (currently just lint) and including those in Makefiles.

The list of targets in defaults.mk will essentially become our CI/provider contract. That is, as we start extracting default behaviors we should also be updating CI to invoke those targets directly for local & CI to share overridden behaviors.

Related to https://github.com/pulumi/ci-mgmt/pull/2024.


provider-ci/internal/pkg/templates/base/Makefile
```diff
@@ -262,17 +264,23 @@ install_nodejs_sdk: .make/install_nodejs_sdk
...
+# `lint.fix` is a utility target meant to be run manually
 # that will run the linter and fix errors when possible.
-lint_provider.fix: upstream
+.PHONY: lint.fix
+lint.fix: upstream lint.pre default.lint.fix lint.post
+
+.PHONY: lint.pre
+lint.pre:
 	git grep -l 'go:embed' -- provider | xargs perl -i -pe 's/go:embed/ goembed/g'
-	cd provider && golangci-lint run --path-prefix provider -c ../.golangci.yml --fix
```
pose: Why is this required? Could we add some comment on why we need to do `lint.pre` and `lint.post`? Thanks!

Issue #2042, opened 2026-02-05
Modify repo structure to follow `provider/` convention path


pose:
Related to: https://github.com/pulumi/pulumi-std/pull/118

blampe:
> After moving `go.mod` and `go.sum` to the `providers` folder in `pulumi-pulumiservice`

If the module is already at the repo root we should try to keep it there (https://github.com/pulumi/ci-mgmt/pull/1238 explains why). There's a ci-mgmt config that lets us override the module location.

pose:
1. `.PHONY` target

OR

1. empty go file + override path

OR

1. GOSRC/GOMODSUM

pose:
Went with the "empty go file + change glob to use {{ .Config.ModulePath }}/go.*" approach. Resolving in favor of https://github.com/pulumi/ci-mgmt/pull/2047

PR #2043, opened 2026-02-05
Fix `bin/$(CODEGEN)` wildcard rule

pose:
Part of #2042 

When trying to build the sdks for `pulumi-pulumiservice`, I was getting the following error:

```
> $ make
make: *** No rule to make target `provider/*.go', needed by `bin/pulumi-gen-pulumiservice'.  Stop.
```

This is caused since the provider doesn't have files on that folder. Here's the relevant `Makefile` snippet:

```Makefile
bin/$(CODEGEN): provider/*.go provider/go.* .make/upstream
	(cd provider && go build $(PULUMI_PROVIDER_BUILD_PARALLELISM) -o $(WORKING_DIR)/bin/$(CODEGEN) -ldflags "$(LDFLAGS_PROJ_VERSION) $(LDFLAGS_EXTRAS)" $(PROJECT)/$(PROVIDER_PATH)/cmd/$(CODEGEN))
```

The wildcard expansion happens in the rule for `bin/$(CODEGEN)`, so that each existing ‘.go’ file becomes a prerequisite and will be recompiled if necessary. But that make rule is wrong. Adapted from [the Make manual section "[Pitfalls of Using Wildcards](https://www.gnu.org/software/make/manual/html_node/Wildcard-Pitfall.html)":

> When a wildcard matches no files, it is left as it is, so then `bin/$(CODEGEN)` will depend on the oddly-named file  `*.go`. Since no such file is likely to exist, make will give you an error saying it cannot figure out how to make `*.go`

Which is exactly the error I am getting above.

I believe the original author/s intention was to signal a dependency on `go.mod`, `go.sum`, and all the `.go` files in the `providers/` folder. Hence, the target needs to be re-built every time those files are changed. The correct way to achieve that is:

```Makefile
GOSRCS = $(shell find ./provider -type f -name '*.go')
GOMODSUM = $(shell find ./provider -type f -name 'go.*')
bin/$(CODEGEN): $(GOSRCS) $(GOMODSUM) .make/upstream
	(cd provider && go build $(PULUMI_PROVIDER_BUILD_PARALLELISM) -o $(WORKING_DIR)/bin/$(CODEGEN) -ldflags "$(LDFLAGS_PROJ_VERSION) $(LDFLAGS_EXTRAS)" $(PROJECT)/$(PROVIDER_PATH)/cmd/$(CODEGEN))
```


blampe:
This generic template is kinda weird... It gets a Makefile for a bridged provider and still needs a "gen" binary?

Thinking out lout, would this make more sense as a customizable target like in https://github.com/pulumi/ci-mgmt/pull/2034? Or it's at least thinking through what this would look like in that model. Long term we don't want any of these gen/tfgen binaries right? So the targets to override would look more like gen_node, gen_go, etc.

PR #2045, opened 2026-02-06
Add pulumi-std

blampe:
https://github.com/pulumi/pulumi-std/issues/64


pose (approved):
Thanks for catching one for the team on this one! 🙏

Issue #2037, opened 2026-02-04
Provide opt-in Makefile to use `pulumi package gen-sdk`


pose:
## Background

Bridge providers use a build-time helper utility to generate schemas and SDKs. By convention these utilities are named `pulumi-tfgen-$provider` (e.g. `pulumi-tfgen-aws`, etc.) How do these helpers work? They invoke either `pftfgen.Main` or `pftfgen.MainWithMuxer` provided by the `pkg/pf/tfgen` package of the `pulumi-terraform-bridge` module.

_(I believe that)_ When attempting to unify all the providers, a `pulumi-gen-$provider` was introduced to non-bridged providers (e.g. `pulumi-eks`, `pulumi-aws`, and `pulumi-aws-apigateway`). Example: https://github.com/pulumi/pulumi-eks/blob/master/provider/cmd/pulumi-gen-eks/main.go#L146-L157 

From a backporting perspective, we would like to converge everything to use `pulumi package gen-sdk` command. In addition, we are in the process of extracting the SDK generation out of the code as well. For instance, https://github.com/pulumi/pulumi/issues/21638 tracks moving out C# codegen code from [github.com/pulumi/pulumi/pkg/v3/codegen/dotnet](http://github.com/pulumi/pulumi/pkg/v3/codegen/dotnet) into the `pulumi/pulumi-dotnet` repo.

## Known blockers for SDK generation

* https://github.com/pulumi/pulumi/issues/16099: Providers need to keep rebuilding schemas for every target language because description fields can reference property names or other resources. These resources can be named differently in different language's SDKs. Example: `outboundPolicy` or `outbound_policy` depending on the SDK. All that's left as part of "tfgen" is the filtering functionality that can be taken out of the bridge and hooked up to an SDK gen pipeline (or added to SDKgen in Core)[^1].
* 

[^1]: https://github.com/pulumi/pulumi-terraform-bridge/pull/3188

## Desired State

Our desired _near term_ state is:
1. Keep using `pulumi-tfgen-$provider` to generate the schemas for bridge providers. Remove as much code as possible if this is redundant for non-bridge providers. Non-bridge providers **can be called with `gen-sdk`** and there's not necessary filter that needs to take place.
2. Move all SDK generation to use  `pulumi package gen-sdk` (this is a temporary solution until https://github.com/pulumi/ci-mgmt/issues/1712)

[...]

PR #2062, opened 2026-02-17
Add package gen-sdk toggles for Dotnet, Go, NodeJS, and Python SDKs

pose:
Extends PR #2048 to add opt-in toggles for the remaining four SDK languages, enabling progressive migration from provider-specific codegen binaries to the unified `pulumi package gen-sdk` command.

Key changes:
- Add config fields: useDotnetPackageGenSdk, useGoPackageGenSdk, useNodejsPackageGenSdk, usePythonPackageGenSdk (all default to false)
- Update Makefile template with conditional logic for each language
- Schema dependency added (conditionally skipped for noSchema providers)
- Remove version.txt creation from Dotnet (handled via --version flag)
- Remove README.md copy from Python (embedded in schema)

Part of #2037

### DIFF
--- a/provider-ci/internal/pkg/config.go
+++ b/provider-ci/internal/pkg/config.go
@@ -337,6 +337,18 @@ type Config struct {
 	// UseJavaPackageGenSdk controls whether we use the Java SDK generation via package gen-sdk
 	UseJavaPackageGenSdk bool `yaml:"useJavaPackageGenSdk"`
 
+	// UseDotnetPackageGenSdk controls whether we use the Dotnet SDK generation via package gen-sdk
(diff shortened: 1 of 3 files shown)


blampe (approved):
Re: the pending status check that's waiting on https://github.com/pulumi/ci-mgmt/pull/2063

Just curious re: the per-language toggles, do we have providers that can use gen-sdk for one language but not all of them?

pose:
> Re: the pending status check that's waiting on #2063
> 
> Just curious re: the per-language toggles, do we have providers that can use gen-sdk for one language but not all of them?

Thanks for the review and the fix 🙏 . No, but I wanted granularity to gate it in case we have problems with specific providers and specific languages without having to do a new release from `ci-mgmt`.

PR #2070, opened 2026-02-19
Add repo automation with Linear + Claude

blampe:
NB: This is squarely in the “experiment with AI to see what works” category. Would appreciate some feedback and 🚢. As an experiment I'd like to try to use this for https://github.com/pulumi/home/issues/4529.

--

This introduces orchestration to bridge Linear issues with AI automation across all providers.

The workflow looks like this:

- An issue is created in Linear. (You can use the "All Providers" template to automatically create sub-tasks for each provider.)
- When the issue is assigned to "Provider Agent" (an OAuth app I already provisioned in Linear) it starts a Linear [agent session](https://linear.app/developers/agent-interaction).
- A lambda receives a webhook for that session and invokes a GitHub workflow (linear-agent.yml).
- This workflow parses out the issue description and any impacted repositories. It then dispatches additional workflows to process each of those repositories.
- Each worker workflow posts status updates back to Linear as it works.

```
Linear Issue Assigned to "Provider Agent"
          |
          v
   AWS Lambda Webhook  <-- deployed via infra/linear-agent (Pulumi stack)
          |
          | workflow_dispatch
          v
   ci-mgmt/.github/workflows/linear-agent.yml
          |
          |
          v
   Claude Code Orchestrator (claude-code-action)
    - Posts "thought" to Linear
    - Queries Linear for child sub-issues
    - Parses issue body for provider names
          |
          |  gh workflow run (one per provider)
          +--------+--------+--------+
          |        |        |        |
          v        v        v        v
    worker/aws  worker/gcp  worker/..  worker/..
    (linear-agent-worker.yml)
          |
          v
   [per-repo work: open PRs, update workflows, etc.]
          |
          v
   Claude posts "response" back to Linear session
```

Issues (the parent tracking issue or sub-issues) can be commented on with corrections or clarifications and the orchestrator will apply those updates to the relevant repos.

The webhook stack is included as part of the merge group's "must deploy" infra, and `.github/workflows/deploy-infra-providers.yml` has been renamed to `deploy-infra.yml`.

--- 

Could we do this today using GitHub and our existing issue template? Yes.

Could we do this with an alternative agent like Copilot? Yes.

So what does Linear bring to the table? It has some nice quality of life features, e.g. issue templates that can automatically create sub-issues, so large "fix all providers" can be better planned and tracked across milestones.

But if we decide not to go with Linear it's not too much trouble to make this work with GitHub instead. Our handler would trigger on GitHub events instead, and it would invoke essentially the same workflow.

### DIFF
--- a/infra/linear-agent/handler.ts
+++ b/infra/linear-agent/handler.ts
@@ -1,4 +1,12 @@
 import * as crypto from "crypto";
+import { SSMClient, GetParameterCommand, PutParameterCommand } from "@aws-sdk/client-ssm";
+
+const log = {
+  info: (msg: string, extra?: Record<string, unknown>) =>
+    console.log(JSON.stringify({ level: "info", message: msg, ...extra })),
+  error: (msg: string, extra?: Record<string, unknown>) =>
+    console.error(JSON.stringify({ level: "error", message: msg, ...extra })),
+};
 
 // Lambda Function URL event shape (payload format version 2.0)
 interface LambdaFunctionUrlEvent {
@@ -41,7 +49,6 @@ async function getLinearToken(): Promise<string> {
   const paramName = process.env.LINEAR_TOKEN_PARAM;
   if (!paramName) return "";
 
-  const { SSMClient, GetParameterCommand } = require("@aws-sdk/client-ssm");
   const ssm = new SSMClient({});
   const response = await ssm.send(
     new GetParameterCommand({ Name: paramName, WithDecryption: true }),
@@ -80,7 +87,7 @@ async function handleOAuthCallback(
   const paramName = process.env.LINEAR_TOKEN_PARAM;
 
   if (!clientId || !clientSecret || !callbackUrl || !paramName) {
-    console.error("Missing OAuth configuration");
+    log.error("Missing OAuth configuration");
     return { statusCode: 500, body: "Server configuration error" };
   }
 
@@ -98,9 +105,7 @@ async function handleOAuthCallback(
   });
 
   if (!tokenResponse.ok) {
-    console.error(
-      `Token exchange failed: ${tokenResponse.status} ${await tokenResponse.text()}`,
-    );
+    log.error("Token exchange failed", { status: tokenResponse.status, body: await tokenResponse.text() });
     return { statusCode: 502, body: "Token exchange failed" };
   }
 
@@ -109,7 +114,6 @@ async function handleOAuthCallback(
   };
 
   // Store token in SSM Parameter Store and invalidate cache
-  const { SSMClient, PutParameterCommand } = require("@aws-sdk/client-ssm");
   const ssm = new SSMClient({});
   await ssm.send(
     new PutParameterCommand({
@@ -121,7 +125,7 @@ async function handleOAuthCallback(
   );
   cachedLinearToken = undefined;
 
-  console.log("OAuth token stored successfully");
+  log.info("OAuth token stored successfully");
   return {
     statusCode: 200,
     headers: { "Content-Type": "text/html" },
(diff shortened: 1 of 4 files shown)


infra/linear-agent/handler.ts
```diff
@@ -0,0 +1,290 @@
...
+export const handler = async (
+  event: LambdaFunctionUrlEvent,
+): Promise<LambdaFunctionUrlResult> => {
+  const { method, path } = event.requestContext.http;
+
+  if (method === "GET" && path === "/oauth/callback") {
+    return handleOAuthCallback(event);
+  }
+
+  if (method === "POST") {
```
pose: (minor) Don't we want to restrict the route if it doesn't match the webhook one?

infra/linear-agent/handler.ts
```diff
@@ -0,0 +1,290 @@
...
+      grant_type: "authorization_code",
+      code,
+      redirect_uri: callbackUrl,
+      client_id: clientId,
+      client_secret: clientSecret,
+    }).toString(),
+  });
+
+  if (!tokenResponse.ok) {
+    console.error(
```
pose: (minor) Could we log in structured JSON format to have easy queryable CloudWatch logs?

infra/linear-agent/index.ts
```diff
@@ -0,0 +1,81 @@
...
+// Note on LINEAR_CALLBACK_URL: this is a bootstrapping config value.
+// 1. Run `pulumi up` once (set linearCallbackUrl to a placeholder or omit via config.get).
+// 2. Note the exported webhookUrl, compute: <webhookUrl>oauth/callback
+// 3. Run `pulumi config set linearCallbackUrl <url>` then `pulumi up` again.
+// The callback URL is only used during OAuth installation, not for webhook processing.
+const callbackUrl = config.get("linearCallbackUrl") ?? "";
+
+const func = new aws.lambda.CallbackFunction("linear-webhook-handler", {
+  runtime: aws.lambda.Runtime.NodeJS22dX,
+  callback: require("./handler").handler,
```
pose: Why do we need to do the `require` this way? Could we move it to the top with the request of imports?

infra/linear-agent/handler.ts
```diff
@@ -0,0 +1,290 @@
...
+
+  if (
+    !sig ||
+    sig.length !== expected.length ||
+    !crypto.timingSafeEqual(Buffer.from(sig), Buffer.from(expected))
+  ) {
+    return { statusCode: 401, body: "Invalid signature" };
+  }
+
+  const payload: AgentSessionEvent = JSON.parse(event.body ?? "{}");
```
pose: (major) I think we should throw or return a `400 Bad Request` if `event.body` is not JSON-parseable. It's going to be hard to debug if we default to "{}".

infra/linear-agent/handler.ts
```diff
@@ -0,0 +1,290 @@
...
+  if (payload.type !== "AgentSessionEvent") {
+    return { statusCode: 200, body: "Ignored" };
```
pose: Since we are ignoring a webhook call, we should be logging we are doing it: if not it will lead to some serious head-scratching.

infra/linear-agent/handler.ts
```diff
@@ -0,0 +1,290 @@
...
+  }
+
+  const { action, agentSession, agentActivity } = payload;
+  const session = agentSession ?? { id: "", status: "" };
+  const issue = session.issue ?? { id: "", title: "", description: "" };
```
pose: (minor) Shouldn't we validate these?

infra/linear-agent/handler.ts
```diff
@@ -0,0 +1,290 @@
...
+          event_type: action === "create" ? "created" : "prompted",
+          prompt_body: agentActivity?.body ?? "",
+        },
+      }),
+    },
+  );
+
+  if (!dispatchResponse.ok) {
+    const errorBody = await dispatchResponse.text();
+    console.error(`GitHub dispatch failed: ${dispatchResponse.status} ${errorBody}`);
```
pose: (same as above about logging)

infra/linear-agent/handler.ts
```diff
@@ -0,0 +1,290 @@
...
+// Mirrors what `actions/create-github-app-token` does in CI.
+async function generateGitHubInstallationToken(): Promise<string> {
+  const appId = process.env.GITHUB_APP_ID;
+  const installationId = process.env.GITHUB_APP_INSTALLATION_ID;
+  // Private key may have literal '\n' sequences in the env var — normalize them.
+  const privateKey = (process.env.GITHUB_APP_PRIVATE_KEY ?? "").replace(/\\n/g, "\n");
+
+  if (!appId || !installationId || !privateKey) {
+    throw new Error("Missing GITHUB_APP_ID, GITHUB_APP_INSTALLATION_ID, or GITHUB_APP_PRIVATE_KEY");
+  }
```
pose: (Same as above about failing the function earlier)

infra/linear-agent/handler.ts
```diff
@@ -0,0 +1,290 @@
...
+}
+
+async function postLinearActivity(
+  token: string,
+  sessionId: string,
+  type: string,
+  body: string,
+): Promise<void> {
+  if (!token) {
+    console.error("No Linear token available — OAuth installation may be incomplete");
```
pose: (minor) (same as above)

infra/linear-agent/handler.ts
```diff
@@ -0,0 +1,290 @@
...
+    return;
+  }
+
+  const escapedBody = body.replace(/\\/g, "\\\\").replace(/"/g, '\\"');
+  const response = await fetch("https://api.linear.app/graphql", {
```
pose: (minor) constants

PR #2095, opened 2026-02-26
Linear agent fixes

blampe:
Key changes:
  - Token lifecycle: Lambda now handles OAuth refresh tokens; ESC environment reads LINEAR_API_KEY dynamically from SSM so workers always get a fresh token
  - Worker workflow: Provider repo is pre-checked out at ./provider-repo (no more gh repo clone in the agent); agent fetches full issue context (including parent sub-issue) via GraphQL at runtime instead of receiving it as an input
  - Branch naming: Derived from Linear issue identifier (e.g. linear/eng-123) fetched at runtime
  - Tooling: mise installed in worker to supply pinned claude and bun executables; bun package cache added; max-turns increased to 200
  - Permissions: Expanded allowedTools and settings.permissions.allow to include echo, jq, cd, mise; fixed shell sandbox rules in prompt to avoid permission denials
  - Orchestrator: Added cancel-in-progress: false concurrency group; prefixed PR titles; fixed double-dispatch and other orchestration bugs
  - Added infra/linear-agent/README.md documenting the architecture

### DIFF
--- a/.github/workflows/linear-agent-worker.yml
+++ b/.github/workflows/linear-agent-worker.yml
@@ -80,8 +80,8 @@ jobs:
 
       - name: Configure git user for provider repo
         run: |
-          git -C provider-repo config user.email "pulumi-bot@pulumi.com"
-          git -C provider-repo config user.name "Pulumi Bot"
+          git -C provider-repo config --global user.email "pulumi-bot@pulumi.com"
+          git -C provider-repo config --global user.name "Pulumi Bot"
 
       - name: Get claude and bun executable paths
         id: claude


.github/workflows/linear-agent-worker.yml
```diff
@@ -59,22 +68,71 @@ jobs:
...
+          repository: pulumi/pulumi-${{ inputs.repo }}
+          path: provider-repo
+          token: ${{ steps.esc-secrets.outputs.PULUMI_BOT_TOKEN }}
+          fetch-depth: 0
+          submodules: recursive
+
+      - name: Configure git user for provider repo
+        run: |
+          git -C provider-repo config user.email "pulumi-bot@pulumi.com"
+          git -C provider-repo config user.name "Pulumi Bot"
```
corymhall:
```suggestion
          git -C provider-repo config --global user.email "pulumi-bot@pulumi.com"
          git -C provider-repo config --global user.name "Pulumi Bot"
```

I ran into this recently. If work ever needs to be done in submodules we need `--global`

PR #2102, opened 2026-03-06
Add azure-native to providers list

fnune:
Enables versioned documentation generation for azure-native in pulumi/registry.

Related: https://github.com/pulumi/registry/pull/10159

### DIFF
--- a/provider-ci/providers.json
+++ b/provider-ci/providers.json
@@ -10,6 +10,7 @@
     "aws-native",
     "awsx",
     "azure",
+    "azure-native",
     "azuread",
     "azuredevops",
     "cloudamqp",


iwahbe (commented):
I don't have context anymore. It isn't clear to me how provider-ci/providers.json relates to versioned documentation.

I'll let @pose take this one.

pose:
@fnune we had a meeting with @Zaid-Ajaj, who's technically leading that space, and we decided we won't be on-boarding azure-native to `ci-mgmt` for now. If you could keep this list as the source of truth, import it and add it from the consumption side; that would be our preferred option. Thanks!

PR #2181, opened 2026-04-27
Add `integrationTestCoverage` opt-in for native providers

iwahbe:
## Summary

Wires up runtime coverage collection from the SDK integration test matrix for native providers that opt in via `integrationTestCoverage: true` in their `.ci-mgmt.yaml`.

When enabled, the generated workflow:

1. Builds the provider with `make provider_cover` instead of `make provider`. The new target invokes `go build -cover -covermode=atomic`, which produces a binary that writes coverage data files to `\$GOCOVERDIR` whenever it exits.
2. Sets `GOCOVERDIR` on the integration test step so each invocation of the provider (spawned by the Pulumi engine as a subprocess) emits coverage data into a shared directory.
3. After tests finish, runs `go tool covdata textfmt` to convert the collected data into a coverage profile and uploads it to Codecov under an `integration-<language>` flag so the matrix shards merge.

The opt-in is gated so there is no behaviour change for existing providers. Test fixture coverage in \`provider-ci/test-providers/command/\` shows the rendered workflow with the flag enabled.

## Motivation

Today, Codecov for native providers only reflects coverage from `make test_provider` (Go unit tests). The SDK integration matrix runs the actual provider binary as a subprocess but contributes nothing to coverage, so patches whose new code is exercised end-to-end (e.g. CRUD paths driven through the Pulumi engine) show artificially low patch coverage.

Go 1.20+ supports `go build -cover` to instrument binaries; this PR plumbs it through the workflow and Codecov upload.

## Caveats

- The `provider_cover` target is added to the base Makefile template, so providers that use the templated Makefile pick it up automatically. Native providers (\`command\`, \`pulumiservice\`, etc.) have hand-written Makefiles and must add the target locally for the workflow to invoke it. This is documented in the new field's godoc.
- The workflow only enables this for the matrix \`test\` job, not the \`prerequisites\` \`Test Provider Library\` step (that step already produces \`coverage.txt\` from \`make test_provider\`).

## Test plan

- [x] `make test-providers` regenerates fixtures and passes actionlint.
- [x] `provider-ci/test-providers/command/` fixtures show the new steps under the opt-in flag; other test providers are untouched aside from the new `provider_cover` target appearing in their Makefile fixtures.
- [ ] Wire a downstream provider PR (\`pulumi-command\`) that opts in and observe Codecov reflecting integration coverage.

### DIFF
--- a/provider-ci/internal/pkg/templates/base/.github/workflows/build_provider.yml
+++ b/provider-ci/internal/pkg/templates/base/.github/workflows/build_provider.yml
@@ -9,6 +9,13 @@ on:
         required: true
         type: string
         description: Version of the provider to build
+      cover:
+        required: false
+        type: boolean
+        default: false
+        description: Build the provider with `-cover` so it emits runtime
+          coverage data files at exit. Used by the SDK integration test
+          workflow when `integrationTestCoverage` is enabled.
       matrix:
         required: false
         type: string
@@ -113,6 +120,7 @@ jobs:
       - name: Build provider
         run: make "provider-${{ matrix.platform.os }}-${{ matrix.platform.arch }}"
         env:
+          PROVIDER_BUILD_FLAGS: ${{ inputs.cover && '-cover -covermode=atomic' || '' }}
           AZURE_SIGNING_CLIENT_ID: ${{ steps.esc-secrets.outputs.AZURE_SIGNING_CLIENT_ID }}
           AZURE_SIGNING_CLIENT_SECRET: ${{ steps.esc-secrets.outputs.AZURE_SIGNING_CLIENT_SECRET }}
           AZURE_SIGNING_TENANT_ID: ${{ steps.esc-secrets.outputs.AZURE_SIGNING_TENANT_ID }}
--- a/provider-ci/internal/pkg/templates/base/Makefile
+++ b/provider-ci/internal/pkg/templates/base/Makefile
@@ -341,6 +341,15 @@ bin/$(PROVIDER): .make/schema
 #{{- end }}#
 .PHONY: provider provider_no_deps
 
+# Build a coverage-instrumented provider binary suitable for collecting runtime
+# coverage from the SDK integration tests. The binary is written to the same
+# path as `make provider`. Set `GOCOVERDIR` when running the binary so that the
+# Go runtime writes coverage data files to that directory; convert them to a
+# coverage profile with `go tool covdata textfmt -i=$$GOCOVERDIR -o cover.out`.
+provider_cover:
+	cd provider && GOOS=$(shell go env GOOS) GOARCH=$(shell go env GOARCH) CGO_ENABLED=0 go build -cover -covermode=atomic -o $(WORKING_DIR)/bin/$(PROVIDER) -ldflags "$(LDFLAGS)" $(PROJECT)/$(PROVIDER_PATH)/cmd/$(PROVIDER)
+.PHONY: provider_cover
+
 test: export PATH := $(WORKING_DIR)/bin:$(PATH)
 test:
 	cd #{{ .Config.TestFolder }}# && $(GO_TEST_EXEC) -v -tags=$(TESTTAGS) -parallel $(TESTPARALLELISM) -timeout 2h $(value GOTESTARGS)
--- a/provider-ci/internal/pkg/templates/base/Makefile
+++ b/provider-ci/internal/pkg/templates/base/Makefile
@@ -320,7 +320,7 @@ build_provider_cmd = #{{ if .Config.BuildProviderPre -}}#
 build_provider_cmd = #{{ if .Config.BuildProviderPre -}}#
                          #{{ .Config.BuildProviderPre }}#;
                      #{{- end -}}#
-                     cd provider && GOOS=$(1) GOARCH=$(2) CGO_ENABLED=0 go build $(PULUMI_PROVIDER_BUILD_PARALLELISM) -o "$(3)" -ldflags "$(LDFLAGS)" $(PROJECT)/$(PROVIDER_PATH)/cmd/$(PROVIDER)
+                     cd provider && GOOS=$(1) GOARCH=$(2) CGO_ENABLED=0 go build $(PULUMI_PROVIDER_BUILD_PARALLELISM) $(PROVIDER_BUILD_FLAGS) -o "$(3)" -ldflags "$(LDFLAGS)" $(PROJECT)/$(PROVIDER_PATH)/cmd/$(PROVIDER)
 #{{- end }}#
 
 provider: bin/$(PROVIDER)
--- a/provider-ci/internal/pkg/templates/native/.github/workflows/build.yml
+++ b/provider-ci/internal/pkg/templates/native/.github/workflows/build.yml
@@ -431,6 +431,18 @@ jobs:
         2h -parallel 4 ./...
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    #{{- else if .Config.IntegrationTestCoverage }}#
+    # `go test -cover` would override the test process's GOCOVERDIR with a
+    # private tmpdir, which then gets inherited (and discarded) by the
+    # provider subprocess. Omit -cover so the env var we set below sticks.
+    - name: Run tests
+      run: >-
+        set -euo pipefail
+
+        cd examples && go test -count=1 -timeout 2h -tags=${{ matrix.language }} -parallel 4 .
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GOCOVERDIR: ${{ github.workspace }}/coverage
     #{{- else }}#
     - name: Run tests
       run: >-
@@ -439,9 +451,6 @@ jobs:
         cd examples && go test -count=1 -cover -timeout 2h -tags=${{ matrix.language }} -parallel 4 .
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        #{{- if .Config.IntegrationTestCoverage }}#
-        GOCOVERDIR: ${{ github.workspace }}/coverage
-        #{{- end }}#
     #{{- end }}#
     #{{- if .Config.IntegrationTestCoverage }}#
     - name: Convert coverage data
--- a/provider-ci/internal/pkg/templates/native/.github/workflows/build.yml
+++ b/provider-ci/internal/pkg/templates/native/.github/workflows/build.yml
@@ -121,7 +121,7 @@ jobs:
     #{{- end }}#
     #{{- if ne .Config.Provider "kubernetes" }}#
     - name: Build Provider
-      run: make provider
+      run: #{{ if .Config.IntegrationTestCoverage }}#make provider_cover#{{ else }}#make provider#{{ end }}#
     #{{- end }}#
     - name: Check worktree clean
       id: worktreeClean
@@ -421,6 +421,10 @@ jobs:
         node_image: kindest/node:v1.29.2
         config: kind.config.yml
     #{{- end }}#
+    #{{- if .Config.IntegrationTestCoverage }}#
+    - name: Prepare coverage directory
+      run: mkdir -p ${{ github.workspace }}/coverage
+    #{{- end }}#
     #{{- if eq .Config.Provider "kubernetes" }}#
     - name: Run tests
       run: cd tests/sdk/${{ matrix.language }} && go test -v -count=1 -cover -timeout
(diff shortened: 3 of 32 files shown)


pose:
(@iwahbe I'll have a look, sorry I have been on holidays)

provider-ci/internal/pkg/templates/base/Makefile
```diff
@@ -341,6 +341,15 @@ bin/$(PROVIDER): .make/schema
 #{{- end }}#
 .PHONY: provider provider_no_deps
 
+# Build a coverage-instrumented provider binary suitable for collecting runtime
+# coverage from the SDK integration tests. The binary is written to the same
+# path as `make provider`. Set `GOCOVERDIR` when running the binary so that the
+# Go runtime writes coverage data files to that directory; convert them to a
+# coverage profile with `go tool covdata textfmt -i=$$GOCOVERDIR -o cover.out`.
+provider_cover:
```
pose, line 349: nit: Could we move `provider_cover` into its own file under `scripts/`, e.g. `scripts/provider_cover.mk`, the same way `crossbuild.mk` is factored out today? The base Makefile is already long and this is feature-gated scaffolding rather than core build behavior. Keeps the main file tidy and makes it easier to evolve or rip out later if priorities shift.

provider-ci/internal/pkg/templates/base/.github/workflows/build_provider.yml
```diff
@@ -113,6 +120,7 @@ jobs:
       - name: Build provider
         run: make "provider-${{ matrix.platform.os }}-${{ matrix.platform.arch }}"
         env:
+          PROVIDER_BUILD_FLAGS: ${{ inputs.cover && '-cover -covermode=atomic' || '' }}
```
pose, line 123: The PR introduces two different ways to build with coverage: this file uses a `PROVIDER_BUILD_FLAGS: -cover -covermode=atomic` env var injected into the cross-platform `make provider-${os}-${arch}` target, while `prerequisites.yml` and the four native workflow files all call `make provider_cover` directly.

I'm guessing the split is because `provider_cover` as defined only builds for host OS/arch, so it can't replace the matrix build here. If that's right, my preference would be to consolidate so the flag list lives in the Makefile rather than the YAML. Either extend `provider_cover` to take os/arch parameters and call it from both sites, or push `-cover -covermode=atomic` into a Makefile variable that `make provider-${os}-${arch}` reads when coverage is requested. The broader push is to keep build logic in `make` and have the YAML stay thin, since we're trying to move less of this into ci-mgmt over time, not more.

provider-ci/internal/pkg/templates/native/.github/workflows/build.yml
```diff
@@ -436,6 +452,20 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     #{{- end }}#
+    #{{- if .Config.IntegrationTestCoverage }}#
+    - name: Convert coverage data
+      if: always()
+      run: go tool covdata textfmt -i=${{ github.workspace }}/coverage -o
+        ${{ github.workspace }}/coverage-${{ matrix.language }}.out
+    - name: Upload integration test coverage to Codecov
```
pose, line 460: The same coverage prep / `GOCOVERDIR` env var / `go tool covdata textfmt` / Codecov upload block is duplicated across four files: this one, `prerelease.yml`, `release.yml`, and `run-acceptance-tests.yml`. Each is ~14 lines that have to stay in sync.

Can this be factored out, the way the base template factors `build_provider.yml` into a reusable workflow? Even a composite action under `.github/actions/` that the four call sites pull in would shrink the diff substantially and remove the drift surface I called out in #5 of the top-level comment. If there are differences across the four files that block consolidation I'm missing, that's fine, but they look interchangeable to me.

provider-ci/internal/pkg/templates/native/.github/workflows/build.yml
```diff
@@ -436,6 +452,20 @@ jobs:
...
+      if: always()
+      uses: codecov/codecov-action@57e3a136b779b570ffcdbf80b3bdc90e7fab3de2 # v6.0.0
```
pose, line 462: The Codecov action version is hardcoded here as `codecov/codecov-action@57e3a136...` (same pattern in the new upload steps added by this PR in `prerelease.yml`, `release.yml`, and `run-acceptance-tests.yml`). Worth flagging that the precedent isn't introduced by this PR: the pre-existing "Upload coverage reports to Codecov" step at line 170 of this file already hardcodes the version, and each of the four native templates has the same pre-existing pattern. So the PR is multiplying the hardcoded callsites from 4 to 8 rather than breaking new ground.

The four new ones could pick up `#{{ .Config.ActionVersions.Codecov }}#` the way `base/test.yml` does, which at least stops the drift surface from doubling. Aligning the pre-existing four is fair to leave for a follow-up.

provider-ci/internal/pkg/templates/native/.github/workflows/release.yml
```diff
@@ -383,12 +383,28 @@ jobs:
...
+    # `go test -cover` would override the test process's GOCOVERDIR with a
+    # private tmpdir, which then gets inherited (and discarded) by the
+    # provider subprocess. Omit -cover so the env var we set below sticks.
+    - name: Run tests
+      run: >-
+        set -euo pipefail
+
+        cd examples && go test -count=1 -timeout 2h -tags=${{ matrix.language }} -parallel 4 .
+      env:
+        GTIHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
pose, line 406: nit: pre-existing typo (`GTIHUB_TOKEN` rather than `GITHUB_TOKEN`). The PR adds one more instance on this line, but there are also two pre-existing ones at lines 395 (kubernetes branch) and 415 (else branch). Trivially fixable in all three while we're touching the file.

provider-ci/internal/pkg/templates/native/.github/workflows/build.yml
```diff
@@ -421,12 +421,28 @@ jobs:
         node_image: kindest/node:v1.29.2
         config: kind.config.yml
     #{{- end }}#
+    #{{- if .Config.IntegrationTestCoverage }}#
+    - name: Prepare coverage directory
+      run: mkdir -p ${{ github.workspace }}/coverage
+    #{{- end }}#
     #{{- if eq .Config.Provider "kubernetes" }}#
```
pose, line 428: If a kubernetes provider opts in to `integrationTestCoverage`, this branch wins and the coverage path is silently skipped: the kubernetes test step doesn't receive `GOCOVERDIR` and the post-test upload doesn't run. Probably fine in practice since no kubernetes provider needs this today, but the failure mode of "I set the flag and nothing happened" is unfortunate. Either calling the incompatibility out on the `IntegrationTestCoverage` godoc, or having the generator warn when both are set, would be more defensive. Same conditional structure exists in `prerelease.yml`, `release.yml`, and `run-acceptance-tests.yml`.

pose (changes requested):
Thanks for putting this together, and apologies for the slow turnaround. One bit of context up front so we're working from the same picture: the team direction has been to simplify ci-mgmt and reduce the amount of behavior it owns, not grow it. This PR expands surface area instead, which raises the bar for me on the "why now" and "what does this unlock" questions below.

The items I'd want resolved before merge are inline on the diff (Makefile organization, the build.yml asymmetry, the duplication across the four native workflow files, and the Codecov action version pinning), plus a `GTIHUB_TOKEN` typo nit and one question about the kubernetes interaction. Everything below this line is "help me understand" rather than blocking.

**1. Runtime overhead.** Did you measure how much `-cover` adds to integration test wall time? Go's instrumented binaries usually run a few percent slower but can be worse depending on the workload, and the SDK matrix is already a long pole. A rough before/after on `pulumi-command` would help me size this.

**2. Who consumes the data.** I get that having this data is desirable in the abstract; what I'm less sure about is how actionable it is right now. Could you walk me through how *you* personally use the integration coverage signal? Is it the Codecov comment on the PR that you scroll to, a dashboard you check, the patch coverage status on the merge button, something else? I want a concrete sense of the loop, both to evaluate the cost/benefit and because if there's a workflow I'm not seeing I'd like to learn it.

**3. YAML test direction.** Worth flagging that we're still exploring how integration tests should be structured. One of the active threads (see pulumi/home#4687) is moving language-agnostic integration tests out of the per-language SDK runs and into YAML programs, which would mean fewer language-specific test jobs over time. Not blocking on this PR, but I'd find it useful to hear how you'd expect the `integration-${language}` flag setup to work in that world. Does YAML coverage become its own flag? Do the language flags collapse? Mostly want to make sure we're not optimizing for a shape we may end up moving away from.

**4. Why now.** Genuinely trying to understand the prioritization. The Providers team hasn't flagged missing integration coverage as a pain point that I've heard, so I'd like to know what surfaced this and how it stacks against the other work in flight. Specific incident? Strategic call from somewhere? "I was here so figured I'd fix it"? Any of those is a fine answer, I just want to be calibrated on what we collectively think is highest value right now.

**5. Rollout.** Once this lands, what's the plan? Are you driving the per-provider opt-ins yourself, or is each provider taking it on as they need it? And do we have a way to notice if it stops working in some downstream repo (e.g. the Makefile target drifts)? I'll add that for the native providers in particular the rollout monitoring matters more, because between the duplicated YAML blocks and the hand-maintained `provider_cover` target the drift surface is wider than for base.

iwahbe:
> Thanks for putting this together, and apologies for the slow turnaround. One bit of context up front so we're working from the same picture: the team direction has been to simplify ci-mgmt and reduce the amount of behavior it owns, not grow it. This PR expands surface area instead, which raises the bar for me on the "why now" and "what does this unlock" questions below.

I'd be happy to do this outside ci-mgmt, but I'm not aware of an extension point that would facilitate this. If you can think of a way to do this without changes here, I'm all ears. 

> **1. Runtime overhead.** Did you measure how much `-cover` adds to integration test wall time? Go's instrumented binaries usually run a few percent slower but can be worse depending on the workload, and the SDK matrix is already a long pole. A rough before/after on `pulumi-command` would help me size this.

I did not, but I know of no providers that are compute bound. On pulumi-command, it adds about a minute of build time, but that's within a 2 minute random variation. Without running it dozens of times, its not statistically significant. 

> **2. Who consumes the data.** I get that having this data is desirable in the abstract; what I'm less sure about is how actionable it is right now. Could you walk me through how _you_ personally use the integration coverage signal? Is it the Codecov comment on the PR that you scroll to, a dashboard you check, the patch coverage status on the merge button, something else? I want a concrete sense of the loop, both to evaluate the cost/benefit and because if there's a workflow I'm not seeing I'd like to learn it.

I look at the codecov data within PRs so I can tell if a change is covered. I'm not aware of any meaningful aggregation there. 

> **3. YAML test direction.** Worth flagging that we're still exploring how integration tests should be structured. One of the active threads (see [pulumi/home#4687](https://github.com/pulumi/home/issues/4687)) is moving language-agnostic integration tests out of the per-language SDK runs and into YAML programs, which would mean fewer language-specific test jobs over time. Not blocking on this PR, but I'd find it useful to hear how you'd expect the `integration-${language}` flag setup to work in that world. Does YAML coverage become its own flag? Do the language flags collapse? Mostly want to make sure we're not optimizing for a shape we may end up moving away from.

Long term: we shouldn't have per-language tests in the provider and should move all integration testing to PCL (rather then YAML, it exposes more capabilities). 

In general: Any provider with `integrationTestCoverage` set, when the provider binary for tests, should build with cover so we can upload code coverage. 

> **4. Why now.** Genuinely trying to understand the prioritization. The Providers team hasn't flagged missing integration coverage as a pain point that I've heard, so I'd like to know what surfaced this and how it stacks against the other work in flight. Specific incident? Strategic call from somewhere? "I was here so figured I'd fix it"? Any of those is a fine answer, I just want to be calibrated on what we collectively think is highest value right now.

Because I was doing work on pulumi-pulumiservice & pulumi-command and both have test coverage, which doesn't include integration tests.

> **5. Rollout.** Once this lands, what's the plan? Are you driving the per-provider opt-ins yourself, or is each provider taking it on as they need it? And do we have a way to notice if it stops working in some downstream repo (e.g. the Makefile target drifts)? I'll add that for the native providers in particular the rollout monitoring matters more, because between the duplicated YAML blocks and the hand-maintained `provider_cover` target the drift surface is wider than for base.

Provider authors that want coverage can access it, but we don't need to drive roll-out. This is only as useful as maintainers find it. It's mostly only useful to native providers anyway.

Like I said, if you have a suggestion for how to do this without ci-mgmt, I'd honestly prefer that. I know it's a hot take, but IMO we shouldn't have ever used ci-mgmt for non-bridged providers. It just makes the provider maintainer's experience worse for dubious gain.

provider-ci/internal/pkg/templates/native/.github/workflows/build.yml
iwahbe: Are there multiple kubernetes providers, or just the one?

iwahbe:
I can go through your individual points if/when we decide we want to take this change.

PR #2198, opened 2026-05-04
Support allow-major provider upgrade config

corymhall:
## Summary
- add `disableMajorProviderUpgrades` to `.ci-mgmt.yaml` config, defaulting to `false`
- render `allow-major: true` in generated `.upgrade-config.yml` unless `disableMajorProviderUpgrades: true` is set
- remove the embedded provider-defaults approach and keep the behavior on the existing global defaults path
- cover default-enabled and repo-local-disable behavior in unit tests

## Context
This wires ci-mgmt generation to the new `allow-major` upgrade-provider config from pulumi/upgrade-provider#372. Major provider upgrades are now allowed by default for generated internal bridged provider upgrade configs; individual provider repos can opt out with `disableMajorProviderUpgrades: true` in `.ci-mgmt.yaml`.

## Testing
- `cd provider-ci && make all`

## Follow up

Need to follow up this with setting `disableMajorProviderUpgrades: true` on the following providers:

- `aws`
- `gcp`
- `azure`
- `azuread`
- `cloudflare`
- `random`
- `tls`

closes https://github.com/pulumi/home/issues/4562

### DIFF
--- a/provider-ci/internal/pkg/config.go
+++ b/provider-ci/internal/pkg/config.go
@@ -48,6 +48,10 @@ type Config struct {
 	// https://github.com/search?q=org%3Apulumi+path%3A.ci-mgmt.yaml+%22upstream-provider-repo%22&type=code
 	UpstreamProviderRepo string `yaml:"upstream-provider-repo"`
 
+	// DisableMajorProviderUpgrades is used in the bridge upgrade config to prevent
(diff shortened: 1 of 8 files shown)


guineveresaenger (approved):
Which providers should we _not_ allow to upgrade across major boundaries?

pose (approved):

PR #2232, opened 2026-05-13
Fix always-false condition on cluster teardown step

guineveresaenger:
We should always run the cluster teardown step. The old code was incorrectly gating it on a pull request ref, which _never_ exists on a `main` or `publish` workflow. There's the possibility for extra noise if the cluster _setup_ failed but it would be fairly obvious in context, so "always" is the defensible condition.

This will address the part of https://github.com/pulumi/pulumi-kubernetes/issues/4329 where we kept leaking clusters into our GCP project and relied on the internal cleanup sweep.

- **bugfix: Remove always-false condition on cluster cleanup step and just always run**
- **Regenerate test providers**

### DIFF
--- a/provider-ci/internal/pkg/templates/native/.github/workflows/build.yml
+++ b/provider-ci/internal/pkg/templates/native/.github/workflows/build.yml
@@ -678,7 +678,7 @@ jobs:
     needs:
     - build-test-cluster
     - test
-    if: ${{ always() }}
+    if: ${{ always() && github.repository_owner == 'pulumi' }}
     permissions:
       contents: read
       id-token: write # For Pulumi access token OIDC.
--- a/provider-ci/internal/pkg/templates/native/.github/workflows/build.yml
+++ b/provider-ci/internal/pkg/templates/native/.github/workflows/build.yml
@@ -678,7 +678,7 @@ jobs:
     needs:
     - build-test-cluster
     - test
-    if: github.event.pull_request.head.repo.full_name == github.repository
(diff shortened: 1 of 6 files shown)


pose (changes requested):
This was designed to gate against running on forks. What this condition should mean is "run only if the repo is called `pulumi/pulumi-kubernetes`" (for instance). I don't know the implications of allowing `always` here, but I think we should find a way to express that condition instead of using an `always` here. What do you think?

pose:
@claude have a look at my previous comment and suggest a path forward

pose:
Thanks @claude . You died, but I appreciate the 👀 reaction.

guineveresaenger:
@pose - these workflows are gated on push to only certain branches:

```
    branches:
    - master
    - main
    - feature-**
 ```
 
 and tags in the case of the release builds.
 
I could see a few situations (actually just one: the `feature-**` prefix) where this could possibly trigger across a fork, but we do have branch protections and maintainer protections, and this job won't trigger until _both_ `build-test-cluster` and `test` have run. (`always` is a little bit misleading here; it doesn't mean "literally always run", it means "run regardless of the outcome of the previous tests")
 
In addition, I think a _teardown_ job is the least possible risk to expose to a third party, and it was the easiest fix since it's so difficult to have insight into an Actions environment and its variable scope.

guineveresaenger:
I've gated the check on the code owner to be extra safe.

pose (approved):

pose (approved):

PR #2286, opened 2026-06-02
native: inject local env into integration test steps

iwahbe:
## Summary

The `Run tests` step in the native `build`, `prerelease`, `release`, and `run-acceptance-tests` workflows hardcoded its `env` to only `GITHUB_TOKEN`. Unlike the `Test Provider Library` step (which uses `renderLocalEnv`), it never received the provider-local secret environment variables configured via `envOverride` and consumed from ESC.

For providers whose integration tests depend on those secrets, the `examples/` tests silently skipped (or, once made strict, failed) in CI. This is the root cause behind:

- pulumi/pulumi-terraform#1021 — `TFE_TOKEN` / `TFE_ORGANIZATION` missing from the test step since the ESC migration.
- pulumi/pulumi-terraform#1045 — the resulting `build.yml` failures on `master`.

Those repos are currently patched with manual edits to the generated workflows (pulumi/pulumi-terraform#1044, pulumi/pulumi-terraform#1046); this PR is the durable upstream fix so the next regeneration keeps them.

## Change

Apply `renderLocalEnv` to the `Run tests` steps so they get the same secret env as the provider build step:

```diff
       env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+#{{ .Config | renderLocalEnv | indent 8 }}#
```

This also fixes a pre-existing typo in `release.yml` where `GITHUB_TOKEN` was misspelled as `GTIHUB_TOKEN`.

## Regenerated test-providers

- `docker-build` (native, has ESC secrets) now receives `ARM_CLIENT_SECRET`, `DIGITALOCEAN_TOKEN`, `DOCKER_HUB_PASSWORD` in its `Run tests` step across all four workflows — demonstrating the fix.
- The `GTIHUB_TOKEN`→`GITHUB_TOKEN` typo fix propagates to `release.yml` for all native providers.
- Providers without secret env are unchanged (`renderLocalEnv` emits the same `GITHUB_TOKEN` line).

`make gen` is clean, actionlint passes, and `go test ./internal/...` passes.

## Draft notes / open questions

- Scope is the **native** template only (the template pulumi-terraform uses). The `bridged` and other templates likely have the same hardcoded `Run tests` env and may warrant the same treatment — intentionally left out of this PR; happy to extend.
- Applied to **both** the kubernetes (`tests/sdk`) and default (`examples`) `Run tests` variants for consistency; for providers without secret env this is a no-op.


pose (approved):

pose:
Looks good. Thanks for raising this PR!

PR #2299, opened 2026-06-04
Bump pinned pulumi CLI to 3.242.0

pose:
## Summary

The mise pin for the pulumi CLI lagged well behind current releases, so any tool that shells out to `pulumi` ran against a stale CLI. In particular the `gen-sdk` codegen path exercised by the new `compare_sdk_<lang>` target (#2290) needs a current CLI to be a meaningful comparison. This bumps the pin from 3.183.0 to 3.242.0 so local and CI runs use a current pulumi.

Verified locally that `pulumi package gen-sdk` runs cleanly on 3.242.0 and the compare-sdk flow passes end to end against it. The newest published release is 3.245.0; this lands on 3.242.0 and can be moved forward separately if desired.

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -2,7 +2,7 @@
 go = '1.24'
 golangci-lint = '2.2.2'
 node = '24.1.0'
-"github:pulumi/pulumi" = '3.183.0'
+"github:pulumi/pulumi" = '3.242.0'
 yarn = '1.22.22'
 dotnet = '8'
 shellcheck = '0.10.0'


guineveresaenger:
Should this be floating, not pinned?

guineveresaenger (approved):

pose:
> Should this be floating, not pinned?

Yes! I think I should wire it up with Renovate, but I haven't got to that yet and I want latest `pulumi` for some tests I am doing.

Issue #2291, opened 2026-06-02
Remove gen-sdk vs legacy comparison tooling after full migration


pose:
## Context

Part of the gen-sdk migration epic, pulumi/home#4787. The shadow-gen diff tooling (#2282) and the generated `make compare_sdk_<lang>` target exist only to prove, per provider+language, that `pulumi package gen-sdk` output matches the legacy per-provider codegen during the migration. They are transitional scaffolding, not a permanent part of the build.

## Scope

Once every provider+language has been flipped to `pulumi package gen-sdk` and verified, remove the comparison mechanism entirely:

- the `compare_sdks` / `compare_sdk_<lang>` targets from the base Makefile template (the section is marked temporary in-place),
- the `provider-ci compare-sdk` subcommand and the `internal/pkg/comparesdk` package,
- the comparison CI job/workflow,
- the aggregate parity reporting (#2287).

If the legacy codegen path is fully retired at the same time, the now-unused `useXPackageGenSdk` opt-in flags can be removed too (or tracked as a follow-up).

## Acceptance criteria

- [ ] Comparison tooling, make targets, CI job, and aggregate reporting removed from ci-mgmt.
- [ ] Regenerated provider repos no longer carry the temporary Makefile section.
- [ ] Gated on all providers being verified as migrated (tracked via the epic's per-provider status).

Part of pulumi/home#4787.

PR #2300, opened 2026-06-04
Add shadow-gen diff as a central reusable CI workflow

pose:
## Summary

We're migrating SDK generation from the legacy per-provider codegen binaries to the generic `pulumi package gen-sdk`. To trust that switch, we need ongoing proof that both paths produce the same SDK, per provider and language. This PR adds that verification to provider CI: it runs `make compare_sdk_<lang>` (legacy vs gen-sdk) as an informational, non-blocking, parallel check. TEMPORARY, removed once the migration is verified (pulumi/ci-mgmt#2291).

**Experiment:** instead of baking the workflow into all ~86 provider repos, this hosts it once here as a reusable workflow (`.github/workflows/compare_sdk.yml`) and has providers reference it remotely (`uses: pulumi/ci-mgmt/.github/workflows/compare_sdk.yml@master`). Template variables become workflow inputs, so each provider carries only a thin calling job and the logic is updated in one place. This is the first cross-repo reusable-workflow call from a provider into ci-mgmt, and it follows the "make workflows dumb" direction.

## What it does

- New `workflow_call` workflow with inputs (`version`, `languages`, `runs-on`, `mise-version`, `checkout-submodules`); `languages` expands via `fromJSON` into a per-language matrix.
- Thin `compare_sdk` calling job generated into bridged `main.yml` / `run-acceptance-tests.yml` (gated `not NoSchema`, fork-guarded on PRs).
- Non-blocking (`continue-on-error`) and parallel; report goes to `$GITHUB_STEP_SUMMARY`. Reuses the `prerequisites-bin` build (no rebuild) and needs only the auto `GITHUB_TOKEN` (no ESC, no cloud creds).

## Verified

`make all` green (deterministic generation, tests, 0 lint issues); `actionlint` passes on the new workflow and the regenerated providers. The `make compare_sdk_<lang>` path was smoke-tested end to end locally.

## End-to-end validation

Validated against a real provider via throwaway PRs [pulumi/pulumi-xyz#2430](https://github.com/pulumi/pulumi-xyz/pull/2430) and [#2452](https://github.com/pulumi/pulumi-xyz/pull/2452) (the latter after the version fix), pinning the refs to this branch (neither the workflow nor the `compare-sdk` command is on `master` yet). Attempts:

- **Attempt 1** ([run](https://github.com/pulumi/pulumi-xyz/actions/runs/26962301718)): `compare_sdk` jobs went green but the comparison never ran (`make compare_sdk_<lang>` printed "Nothing to be done"). Cause: the generated `.PHONY` line marked the per-language pattern targets phony, which stops Make applying the pattern rule's recipe. Fixed in #2290 so only the `compare_sdks` aggregate is phony.
- **Attempt 2** ([run](https://github.com/pulumi/pulumi-xyz/actions/runs/26963225007)): the recipe ran, but `go run .../provider-ci@<branch>` failed because Go rejects a branch ref containing slashes ("disallowed version string"). Test-harness artifact only, the real recipe uses `@master`; worked around by pinning the test recipe to the commit SHA.
- **Attempt 3** ([run](https://github.com/pulumi/pulumi-xyz/actions/runs/26963625336)): the comparison ran end to end and reported real differences for xyz (java: 88.2% parity, 15/17 files identical, `Pulumi.yaml` only in legacy). This confirms the centralized path works: reuse the prerequisites build, run legacy codegen + `pulumi package gen-sdk`, write the report to the step summary. The jobs showed red because the command exits non-zero on a real diff, so the comparison step now tolerates that (`continue-on-error`) to keep the check green while informational; the diff stays in the summary.
- **Attempt 4** ([run](https://github.com/pulumi/pulumi-xyz/actions/runs/26963625336/attempts/2)): re-ran with the step-level `continue-on-error` fix. All five `compare_sdk` checks are now green while still running the comparison and reporting diffs (java's summary still shows 88.2% parity). End-to-end behavior confirmed.
- **Attempt 5** ([run](https://github.com/pulumi/pulumi-xyz/actions/runs/27013373133)): most of those diffs were version noise. The legacy binary bakes its build version via ldflags while gen-sdk used a synthetic default, so version-bearing files (`build.gradle`, `.csproj`, `version.txt`, `pulumiUtilities.go`) differed on the version string alone. Fixed in #2288: the command now defaults `--version` to `PROVIDER_VERSION` so both sides embed the same version. Parity jumped (java 94.1%, go 91.7%, dotnet 93.8%, nodejs 94.7%, python 94.1%), leaving one genuine diff per language: `Pulumi.yaml (only in legacy)` (legacy tfgen emits an SDK-root `Pulumi.yaml` that gen-sdk does not).

## Notes

- Cannot execute until it lands on ci-mgmt `master` (providers reference `@master`); `actionlint` does not resolve remote reusable workflows, so generation stays green.
- Requires the org Actions policy to allow calling ci-mgmt reusable workflows from provider repos.

Stacked on #2290 (base `pose/chores/compare-sdk-make-target-2282`), which stacks on #2288. Merge order: #2288, #2290, this.

Part of pulumi/home#4787.

### DIFF
--- a/.github/workflows/compare_sdk.yml
+++ b/.github/workflows/compare_sdk.yml
@@ -1,0 +1,112 @@
+# Reusable workflow: shadow-gen diff. Generates a provider's SDK both ways - via
+# the legacy per-provider codegen binary and via `pulumi package gen-sdk` - from
+# the same schema.json and reports any un-allowlisted difference.
+#
+# It is hosted centrally here in pulumi/ci-mgmt rather than baked into every
+# provider repo. Provider repos call it from their generated main.yml /
+# run-acceptance-tests.yml:
+#
+#   compare_sdk:
+#     needs: prerequisites
+#     uses: pulumi/ci-mgmt/.github/workflows/compare_sdk.yml@master
+#     with:
+#       version: ${{ needs.prerequisites.outputs.version }}
+#       languages: '["nodejs","python","dotnet","go","java"]'
+#       runs-on: ubuntu-latest
+#       mise-version: <mise version>
+#       checkout-submodules: "false"
+#
+# TEMPORARY: this workflow exists only to track the migration to
+# `pulumi package gen-sdk`. It is informational and non-blocking. Once every
+# provider+language has been migrated and verified, it is removed along with the
+# rest of the comparison tooling. See pulumi/ci-mgmt#2291.
+name: compare_sdk
+
+on:
+  workflow_call:
+    inputs:
+      version:
+        description: Provider version. Only needed to satisfy the build; the comparison normalizes the version away.
+        required: true
+        type: string
+      languages:
+        description: JSON array of SDK languages to compare, e.g. '["nodejs","go"]'.
+        required: true
+        type: string
+      runs-on:
+        description: Runner label to execute on.
+        required: false
+        type: string
+        default: ubuntu-latest
+      mise-version:
+        description: Version of mise to install.
+        required: true
+        type: string
+      checkout-submodules:
+        description: Value passed to actions/checkout's submodules input.
+        required: false
+        type: string
+        default: "false"
+
+jobs:
+  compare_sdk:
+    name: compare_sdk (${{ matrix.language }})
+    # Job-level: keep the calling workflow non-blocking even if an infra step
+    # (checkout, mise, artifact restore) fails. The expected "SDKs differ" case is
+    # handled at the Compare step below so the check itself stays green.
+    continue-on-error: true
+    runs-on: ${{ inputs.runs-on }}
+    permissions:
+      contents: read
+    env:
+      PROVIDER_VERSION: ${{ inputs.version }}
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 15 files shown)


guineveresaenger:
I didn't know it was possible (I forgot?) to reference Workflows across repos. 🤯 that's pretty cool! Am I correct in understanding that this would mean that _rather than wait for fanout_, a Workflow is immediately available in repos as soon as merged here? (well, after the fanout of the reference at least.)

pose:
> I didn't know it was possible (I forgot?) to reference Workflows across repos. 🤯 that's pretty cool! Am I correct in understanding that this would mean that _rather than wait for fanout_, a Workflow is immediately available in repos as soon as merged here? (well, after the fanout of the reference at least.)

Exactly! That's the direction I would like to explore since it would mean we can move to a decentralized/federated ci-mgmt model with smaller templates and allowing a "pick-and-choose" model.

corymhall (approved):

PR #2365, opened 2026-06-30
lint-actions: ignore gh-aw maintenance empty-option; pin actionlint via mise

pose:
gh-aw 0.80.9 generates `.github/workflows/agentics-maintenance.yml` with an empty `workflow_dispatch` choice option (`- ''`) so the optional `operation` can default to no-op. That is valid GitHub Actions, but actionlint's `syntax-check` rejects the empty string and fails `lint-actions`.

Add a scoped `ignore` for that file in `.github/actionlint.yaml`, mirroring the existing `concurrency.queue: max` gh-aw false-positive ignore. The empty option is rejected by every actionlint release tested (1.6.27 through 1.7.12), so suppressing the false positive is the fix rather than a version pin.

Also pin `actionlint` in `mise.toml` and run it from the mise toolchain in `lint-actions.yml`, replacing `bash <(curl .../download-actionlint.bash)`. This removes the pipe-to-shell install and the always-latest moving target.

Verified by regenerating the maintenance file with gh-aw 0.80.9 and running `actionlint` over the repo (exit 0).

Part of #2327

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -17,6 +17,13 @@ paths:
       # recognize the `queue` key, so it false-positives this valid syntax.
       - 'unexpected key "queue" for "concurrency" section'
 
+  ".github/workflows/agentics-maintenance.yml":
+    ignore:
+      # gh-aw emits an empty choice option ('') so the optional maintenance
+      # operation can default to "no operation". This is valid GitHub Actions,
+      # but actionlint's syntax-check flags the empty string.
+      - 'string should not be empty'
+
   ".github/workflows/build_provider.yml":
     ignore:
       - 'property ".*" is not defined in object type {}'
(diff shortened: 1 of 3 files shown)


pose:
Note: the validate (Validate Agentic Workflows) check is expected to fail here and does not block this PR (it is not a required check).

It fails for a pre-existing reason on master, unrelated to this change: the committed *.lock.yml files reference github/gh-aw-actions/setup@v0.80.9 (bumped by the renovate digest PRs #2362/#2363/#2364), while mise.toml still pins the gh-aw CLI at 0.76.1. The job runs gh-aw compile (0.76.1) and then checks git diff/template sync, so those refs get rewritten back to v0.76.1 and the lockfiles/templates report as out of sync. This PR only triggers that workflow because it edits mise.toml (a path in the workflow's filter); it touches no gh-aw or lock files.

The skew is resolved by #2280, which bumps the gh-aw CLI to 0.80.9 and regenerates the lock files. All required checks are green, including Run actionlint and shellcheck, which this PR fixes.

guineveresaenger (approved):

Issue #2336, opened 2026-06-17
test_provider fanout job (#2218) breaks providers whose unit tests load the built provider binary


pose:
## Summary

The acceptance workflow refactor in #2218 moved provider unit tests into a standalone `test_provider` job that restores only the `prerequisites-bin` artifact and never builds the provider. This breaks any bridged provider whose `make test_provider` suite includes pulumitest-style program tests that load the built provider plugin from `./bin`, because `bin/pulumi-resource-<name>` is no longer present when the tests run.

## Symptom

The unit test runs `pulumi preview`/`up` against a program that references `pulumi:providers:<name>` with a local `path: ./bin`. With the binary absent, the Pulumi plugin loader falls back to looking for `PulumiPlugin.yaml` and fails:

```
Error loading schema from plugin: failed to load plugin .../bin/pulumi-resource-<name>: loading PulumiPlugin.yaml: open .../bin/PulumiPlugin.yaml: no such file or directory
```

Confirmed on:
- pulumi/pulumi-databricks#1168 (`TestNoDebugLogsOnStartup`)
- pulumi/pulumi-auth0#1249 (`TestOrganizationConnection` and other `provider_program_test.go` tests)

A 12-provider spot check showed only these two failing; aws, gcp, azuread, cloudflare, snowflake, datadog, github, gitlab, newrelic, and mongodbatlas all pass. The differentiator is whether `make test_provider` includes tests that load the built provider binary. There may be a few more affected providers in the full set.

## Root cause

In the old layout, `make test_provider` ran inside the `prerequisites` job right after `make provider`, so `bin/pulumi-resource-<name>` existed. In the fanout layout (`acceptance_fanout`), the lightweight `prerequisites` job skips the provider build (the `prerequisites-bin` artifact contains only the tfgen binary), and the standalone `test_provider` job runs `make test_provider` without building the provider.

## Options

1. Build the provider in the `test_provider` job before running unit tests. `schema-embed.json` and the tfgen binary are already restored into the job, so a lightweight `go build` (embedded schema, no upstream/codegen) is enough. Fixes all affected providers centrally; adds a provider build to every `test_provider` job, but the job is off the critical path so wall-clock is unaffected.
2. Leave ci-mgmt as-is and have each affected provider add the dependency locally (e.g. a `.mk/*.mk` adding `test_provider: bin/$(PROVIDER)`). Targeted, but future providers with such tests will silently regress.

Happy to implement whichever direction is preferred.

[...]

PR #2367, opened 2026-06-30
Build provider binary in fanout test_provider job when opted in

pose:
Adds an opt-in `testProviderNeedsProviderBinary` flag. When set, the fanout `test_provider` job runs `make provider_no_deps` before the unit tests, so providers whose unit tests load the built provider plugin from `./bin` work. Default off keeps the job lightweight for every other provider. All 15 test providers regenerate deterministically; only the opted-in `xyz` golden gains the new step.

Part of #2336

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/provider-ci/internal/pkg/config.go
+++ b/provider-ci/internal/pkg/config.go
@@ -303,6 +303,17 @@ type Config struct {
 	// This function is called without arguments to run unit tests for the provider binary.
 	TestProviderCmd string `yaml:"testProviderCmd"`
 
+	// TestProviderNeedsProviderBinary builds the provider binary in the
+	// standalone acceptance test_provider job before running unit tests. Set
+	// this for providers whose `make test_provider` suite includes tests that
+	// load the built provider plugin from ./bin (e.g. pulumitest program
+	// tests). Defaults to false so the fanout test_provider job stays
+	// lightweight for providers that do not need the binary. The build uses
+	// `make provider_no_deps`, which relies on the schema-embed.json restored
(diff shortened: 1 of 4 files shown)


provider-ci/internal/pkg/config.go
```diff
@@ -303,6 +303,17 @@ type Config struct {
...
 
+	// TestProviderNeedsProviderBinary builds the provider binary in the
+	// standalone acceptance test_provider job before running unit tests. Set
+	// this for providers whose `make test_provider` suite includes tests that
+	// load the built provider plugin from ./bin (e.g. pulumitest program
+	// tests). Defaults to false so the fanout test_provider job stays
+	// lightweight for providers that do not need the binary. The build uses
+	// `make provider_no_deps`, which relies on the schema-embed.json restored
+	// from the prerequisites artifact and so does not regenerate the schema.
+	// See pulumi/ci-mgmt#2336.
```
guineveresaenger, line 314: Should this link to a playbook doc instead?

guineveresaenger (approved):

guineveresaenger (approved):

provider-ci/internal/pkg/config.go
pose: @guineveresaenger I am blinded by [the curse of knowledge](https://en.wikipedia.org/wiki/Curse_of_knowledge) in this particular case so I appreciate the feedback. What would you expect that playbook to contain? Thanks!

PR #2412, opened 2026-07-21
Centralize Renovate updates for generated Mise configuration

corymhall:
## Motivation

Renovate now has a built-in Mise manager. It finds `.config/mise*.toml` files, including both ci-mgmt's canonical templates and the generated copies committed in provider repositories.

That gives the generated files two competing owners. Renovate can bump a generated copy in a provider, but the next ci-mgmt update regenerates the file from the canonical template and changes the version back. The same problem occurs in this repository's committed test-provider snapshots, which Renovate can update directly even though they are generated by `make build`.

This PR establishes ci-mgmt as the single owner of generated Mise defaults. It avoids adding another version file because the existing canonical templates are valid TOML and Renovate can update them directly.

## How this works in practice

- Renovate continues to update the canonical templates at:
  - `provider-ci/internal/pkg/templates/all/.config/mise.toml`
  - `provider-ci/internal/pkg/templates/all/.config/mise.test.toml`
- ci-mgmt's Renovate configuration ignores `provider-ci/test-providers/**` for the Mise manager, so snapshots are not treated as independent dependency sources.
- The existing Renovate post-upgrade command runs `make build`, which regenerates all committed test-provider snapshots after a canonical version changes.
- Generated files now say that dependency updates are managed centrally in `pulumi/ci-mgmt`.
- Provider-owned root `mise.toml` files are unaffected. Providers can continue to override or augment the generated defaults there, and Renovate can continue to update those root files.

The Mise templates live in the `all` template, so this behavior applies to both bridged and native providers without changing their template composition.

## Downstream rollout dependency

This PR handles the ci-mgmt side, but it does not by itself stop Renovate from updating generated `.config/mise*.toml` files in provider repositories.

That exclusion should live in a provider-scoped preset in `pulumi/renovate-config`. The rule should disable the `mise` manager only for root-relative `.config/mise*.toml` files while leaving root `mise.toml` enabled. The README documents the proposed rule.

This PR intentionally does not generate or replace provider `renovate.json5` files. Providers have their own Renovate rules, and replacing those files would risk losing repository-specific configuration. The complete ownership change therefore requires the provider preset to be added and rolled out separately while preserving existing provider configuration.

## Review notes

- The test-provider changes are generated and only add the ownership comment to both Mise files for each fixture.
- No duplicate Mise version source was introduced.
- Root-level provider overrides remain provider-owned and Renovate-manageable.

## Validation

- `cd provider-ci && make clean && make all`
- `npx --yes --package=renovate@43.274.0 renovate-config-validator renovate.json5`
- Parsed both canonical Mise templates with Python `tomllib`
- `git diff --check`


pose:
Thanks again @corymhall for raising this PR.

I agree with this solution as a temporary stop gap. In the long term, I would love to eject `.config/mise.toml` entirely and make each provider update their dependencies at their own pace through Renovate. The less files we have under `ci-mgmt` the better we would be.

### Downstream rollout dependency

I had a look at the README and the following proposed actions:

> This PR handles the ci-mgmt side, but it does not by itself stop Renovate from updating generated .config/mise*.toml files in provider repositories.
> [...]
> That exclusion should live in a provider-scoped preset in pulumi/renovate-config. The rule should disable the mise manager only for root-relative .config/mise*.toml files while leaving root mise.toml enabled. The README documents the proposed rule.

But, unfortunately, I wasn't able to find an easy or simple way to scope the "do not update `.config/mise.toml` files" to only certain provider repos. I was wondering if you had any thoughts about that. Thanks!

PR #2431, opened 2026-08-04
Add a monthly security patch tracker for LTS maintenance branches

guineveresaenger:
This is the first item in a series of work to enable ops alerts and automation for tracking maintenance branch maintenance tasks.

This pull request allows for a provider to set a MaintenanceBranch and an optional MaintenanceReleaseDay field in their ci-mgmt config to receive a Workflow that will set up monthly tracker tickets for security patches to the maintenance branch.

Currently this workflow will:
- Open a new ticket on the provided date (default to first of the month)
- Inform the ops person to run security checks on the maintenance branch and perform necessary updates within the SLO of one month.

Having a tracking ticket will allow us to hook these up to our ops dashboard.
Automated vulnerability detection will be added in a separate Workflow in a future PR.

EDIT: Example issue created: https://github.com/pulumi/pulumi-azure/issues/3615 - note the comment is done by the separate Scan Workflow.

Addresses https://github.com/pulumi/home/issues/4881.


pose (approved):
Same as the other PR. I think this is a good start, but we should consider moving the `bash` code into a GitHub Action we could consume or in a way we could write tests for. Thanks!

PR #2433, opened 2026-08-05
Workflow to scan maintenance branches with osv-scanner

guineveresaenger:
Stacked on https://github.com/pulumi/ci-mgmt/pull/2431.

This pull request adds a daily scanner for security vulnerabilities of severity 7 or higher. It scans `provider/` and `sdks/` folders, and it tracks code reachability as well, to minimize maintenance noise.

It then posts outstanding CVEs to the monthly maintenance patch release issue, where a maintainer can then address the necessary updates.

We could effect something similar with Renovate; however, Renovate's security updates are not granular enough to filter by severity, nor is it possible to use code tracking to see if a vulnerability is even reachable by existing code.

Here's an example issue with table comment: https://github.com/pulumi/pulumi-azure/issues/3615

Addresses https://github.com/pulumi/home/issues/4882.


provider-ci/internal/pkg/templates/all/.github/workflows/maintenance-scan.yml
```diff
@@ -0,0 +1,133 @@
...
+            ./scripts/upstream.sh init
+          else
+            echo "no scripts/upstream.sh; nothing to prepare"
+          fi
+
+      - name: Install osv-scanner
+        run: |
+          set -euo pipefail
+          curl -sSfL -o /usr/local/bin/osv-scanner \
+            https://github.com/google/osv-scanner/releases/download/v2.4.0/osv-scanner_linux_amd64
```
pose, line 48: (minor) any chance we can model this through `mise`?

pose (approved):
I think this is a good start, but we should consider moving the `bash` code into a GitHub Action we could consume or in a way we could write tests for. Thanks!

PR #2463, opened 2026-09-01
Add a renovate_artifacts job to build_sdk

pose:
A dependency bump can invalidate generated files that the per-language SDK build doesn't cover, principally the component lock files. Today that gap is filled by a Renovate `postUpgradeTasks` hook running `make renovate` on the Renovate worker. This adds a CI job that does the same regeneration and commits the result back to the PR, using the same machinery as the SDK commit step.

Doing it in CI keeps repository code off the Renovate workers, and turns a failure into a job you can see rather than an `artifactErrors` warning that silently opens a PR with the dependency bumped and the generated files stale.

The job is generated only where `.ci-mgmt.yaml` sets `renovateCmd`, so of the test providers only `eks` changes. It is not a matrix job and deliberately does not declare `needs: build_sdk`: the language jobs commit under `sdk/`, this one commits everything else via `git add -A -- ':!sdk' ':!provider/cmd/pulumi-resource-<provider>/schema.json'`, so the paths are disjoint and the existing rebase-retry loop reconciles the pushes. It exits early without an empty commit when regeneration is a no-op.

No behaviour change yet: `postUpgradeTasks` still runs. Removing it per repo is the follow-up.

Part of pulumi/home#4904

### DIFF
--- a/provider-ci/internal/pkg/templates/base/.github/workflows/build_sdk.yml
+++ b/provider-ci/internal/pkg/templates/base/.github/workflows/build_sdk.yml
@@ -173,4 +173,104 @@ jobs:
         uses: ./.github/actions/upload-sdk
         with:
           language: ${{ matrix.language }}
+#{{- if .Config.RenovateCmd }}#
+
+  # Regenerate the files that a dependency bump invalidates but that the
+  # per-language SDK build above does not cover, principally lock files. This
+  # replaces a Renovate `postUpgradeTasks` hook: doing it here keeps repo code
+  # off the Renovate workers, and turns a failure into a visible job rather
+  # than a warning that silently ships a half-updated PR.
+  #
+  # Not a matrix job, and no `needs:` on build_sdk. It races the language jobs
+  # deliberately; they commit under sdk/ and this commits everything else, so
+  # the paths are disjoint and the retry loop below reconciles the pushes.
+  renovate_artifacts:
+    name: renovate_artifacts
+    if: contains(github.actor, 'renovate') && github.event_name == 'pull_request'
+    runs-on: #{{ if .Config.Runner.BuildSDK }}##{{- .Config.Runner.BuildSDK }}##{{ else }}##{{- .Config.Runner.Default }}##{{ end }}#
+    permissions:
+      contents: write # To push the regenerated files back to the PR.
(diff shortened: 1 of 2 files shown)


provider-ci/internal/pkg/templates/base/.github/workflows/build_sdk.yml
```diff
@@ -173,4 +173,104 @@ jobs:
...
+        env:
+          GITHUB_TOKEN: ${{ steps.esc-secrets.outputs.PULUMI_BOT_TOKEN }}
+      - name: Download prerequisites
+        uses: ./.github/actions/download-prerequisites
+      - name: Update path
+        run: echo "${{ github.workspace }}/bin" >> "$GITHUB_PATH"
+      - name: Restore makefile progress
+        run: make --touch provider schema
+      - name: Regenerate Renovate artifacts
+        run: make renovate
```
pose, line 226: This is the crux of this PR

provider-ci/internal/pkg/templates/base/.github/workflows/build_sdk.yml
```diff
@@ -173,4 +173,104 @@ jobs:
         uses: ./.github/actions/upload-sdk
         with:
           language: ${{ matrix.language }}
+#{{- if .Config.RenovateCmd }}#
+
+  # Regenerate the files that a dependency bump invalidates but that the
```
corymhall, line 178: I think we should go in a different direction with these renovate commit jobs. Rather than committing everything and pushing in each fanout job I think we should have each of these jobs upload git patches as artifacts. It would look something like:

1. Change SDK jobs to upload patches
2. Change this `renovate_artifacts` to upload a patch
3. Add one final commit job

```
   build_nodejs ───────> nodejs.patch ───┐
   build_python ───────> python.patch ───┤
   build_dotnet ───────> dotnet.patch ───┼─> commit_renovate_changes
   build_go ───────────> go.patch ───────┤          |
   renovate_artifacts ─> renovate.patch ─┘          └─ one commit, one push
```

Projen has a similar `self-mutation` concept that works pretty well (e.g. https://github.com/corymhall/pulumi-aws-connectors/blob/main/.github/workflows/build.yml)

PR #2465, opened 2026-09-02
Add a disableIntegrationTests toggle

pose:
Adds a `disableIntegrationTests` option to `.ci-mgmt.yaml` for LTS branches whose acceptance tests can no longer run.

When set, the `test` job renders as a no-op instead of being removed, so `release.yml`, `prerelease.yml`, `nightly-test.yml`, `run-acceptance-tests.yml` and `main.yml` all keep resolving `needs: test`. Unit tests are unaffected; they run from `prerequisites.yml`.

Defaults to false. Generating a provider with the flag off produces a byte-identical `test.yml` to the committed golden, the stub passes actionlint, and `go test ./internal/...` is green.

### DIFF
--- a/provider-ci/internal/pkg/templates/base/.github/workflows/test.yml
+++ b/provider-ci/internal/pkg/templates/base/.github/workflows/test.yml
@@ -19,6 +19,12 @@ env:
 
 jobs:
   test:
+#{{- if .Config.DisableIntegrationTests }}#
+    runs-on: ubuntu-latest
+    steps:
+    - name: Integration tests disabled
+      run: echo "Integration tests are disabled for this branch."
+#{{- else }}#
     permissions:
       contents: read
       id-token: write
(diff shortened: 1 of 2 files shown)


provider-ci/internal/pkg/templates/base/.github/workflows/test.yml
```diff
@@ -19,6 +19,12 @@ env:
 
 jobs:
   test:
+#{{- if .Config.DisableIntegrationTests }}#
```
corymhall, line 22: The LTS policy says LTS branches are gated by unit tests only, could we just somehow disable tests for all LTS branches and not need a toggle?

pose: I think that if the tests are still passing, we shouldn't disable them. I would disable them only when they start failing and we cannot make time for it. What do you think?

pose: Discussed offline: Toggle on LTS/backport or backport config to centralize configuration related to the backport. Unify based on the options from the tier 1 major upgrade guide.

